PR #93, opened 2025-04-03
refactor: Remove unused scheduler state

sitaowang1998:
# Description

In early design, scheduler has three states: normal, gc and recovery. However, this design has been ditched a long time ago. Now the scheduler does not have any state so it can serves schedule task request all the time, and gc and recovery are performed in another thread. This pr removes the codes and tests related to the scheduler state.

# Checklist

  breaking change.

# Validation performed

* [ ] Github workflows pass
* [x] Unite tests pass in dev container
* [x] Integration tests pass in dev container

### DIFF
--- a/tools/scripts/storage/init_db.sql
+++ b/tools/scripts/storage/init_db.sql
@@ -9,7 +9,6 @@ CREATE TABLE IF NOT EXISTS `schedulers`
     `id`      BINARY(16)                        NOT NULL,
     `address` VARCHAR(40)                       NOT NULL,
     `port`    INT UNSIGNED                      NOT NULL,
-    `state`   ENUM ('normal', 'recovery', 'gc') NOT NULL,
     CONSTRAINT `scheduler_driver_id` FOREIGN KEY (`id`) REFERENCES `drivers` (`id`) ON UPDATE NO ACTION ON DELETE CASCADE,
     PRIMARY KEY (`id`)
 );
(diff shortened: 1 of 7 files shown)


tools/scripts/storage/init_db.sql
```diff
@@ -107,6 +106,16 @@ CREATE TABLE IF NOT EXISTS `task_instances`
...
+CREATE TABLE IF NOT EXISTS `scheduler_leases`
+(
+    `scheduler_id` BINARY(16) NOT NULL,
+    `task_id`      BINARY(16) NOT NULL,
+    `lease_time`   TIMESTAMP  NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+    CONSTRAINT `lease_scheduler_id` FOREIGN KEY (`scheduler_id`) REFERENCES `schedulers` (`id`) ON UPDATE NO ACTION ON DELETE CASCADE,
+    CONSTRAINT `lease_task_id` FOREIGN KEY (`task_id`) REFERENCES `tasks` (`id`) ON UPDATE NO ACTION ON DELETE CASCADE,
+    INDEX (`scheduler_id`)
+);
+
```
davidlion, line 118: Is this change unrelated to the PR? There is no mention of it in the description and none of the other changes are related to leases or require a new table (as far as I can tell).

Is it possible to move this to another more relevant PR?

davidlion (changes requested):

tools/scripts/storage/init_db.sql
davidlion: After looking at #98 I'm pretty sure this table should be removed from this PR and added in that one.
```suggestion
```

PR #130, opened 2025-05-08
fix: Switch from `abseil` to `std::vector` for task function and name mapping (fixes #131).

### LINKED ISSUE #131: Crash in task executor when loading task library built with different abseil version.
### Bug

As title. Different `abseil` version has different memory layout for `absl::flat_hash_map`, and the task executor cannot correctly interpret the map stored in task library.

### Spider version

0f88fc29d99d767e3db54d7b11f7ab6867d57e59

### Environment

Ubuntu 22.04

### Reproduction steps

Build task library and task executor with different `abseil` versions and run the integration tests.

sitaowang1998:
# Description

`FunctionMap` and `FunctionNameMap` uses `absl::flat_hash_map` to store the mapping between functions and names. However, difference in `abseil` versions between a task executor and a task library means that the task executor's `flat_hash_map` has different memory layout than the map stored by the task library, and the reading of the map fails and crashes.

This pr solves this issue by using a plain `std::vector` for the map. The data are stored as a consecutive array, and layout remains the same across different compiler versions. Fixes #131.

# Checklist

  breaking change.

# Validation performed
* [x] Integration tests pass with task library compiled with gcc-12 and other `Spider` components compiled with gcc-13.

## Summary by CodeRabbit

- **Refactor**
	- Improved internal handling of function and function name registration for better consistency and efficiency.
	- Updated internal storage structures for managing functions and their names.
- **Tests**
	- Adjusted tests to align with updated function pointer handling.

### DIFF
--- a/src/spider/core/TaskGraphImpl.hpp
+++ b/src/spider/core/TaskGraphImpl.hpp
@@ -138,7 +138,7 @@ public:
         // NOLINTBEGIN(cppcoreguidelines-pro-type-reinterpret-cast)
         std::optional<std::string> const function_name
                 = FunctionNameManager::get_instance().get_function_name(
-                        reinterpret_cast<void const*>(task_function)
+                        reinterpret_cast<TaskFunctionPointer const>(task_function)
                 );
         // NOLINTEND(cppcoreguidelines-pro-type-reinterpret-cast)
         if (!function_name.has_value()) {
--- a/src/spider/worker/FunctionManager.cpp
+++ b/src/spider/worker/FunctionManager.cpp
@@ -3,6 +3,7 @@
 #include <cstddef>
 #include <optional>
 #include <string>
+#include <string_view>
 #include <tuple>
 #include <type_traits>
 #include <vector>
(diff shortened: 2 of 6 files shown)


davidlion:
> **Actionable comments posted: 1**
> ♻️ Duplicate comments (1)
> 
> > src/spider/worker/FunctionManager.hpp (1)> `381-386`: **Same race-condition applies to `register_function_invoker`** – see previous comment.
> 
> 🧹 Nitpick comments (2)
> 
> > src/spider/worker/FunctionManager.hpp (2)> `49-50`: **Potential linear-search hot-spot after replacing `flat_hash_map` with `std::vector`**
> > > `FunctionMap` is now a simple vector, so every lookup performed by `get_function()` or by the two `register_*` helpers is O(n).  In typical deployments we might register only a handful of tasks, but if this registry grows (micro-service with hundreds of tasks, plugin system, etc.) start-up time and per-call lookup latency will regress compared with the previous hash look-ups.
> > > If keeping a contiguous layout is mandatory (ABI reasons), consider:
> > > 
> > > 1. Maintaining the vector **sorted by name** and using `std::lower_bound` (O(log n)), or
> > > 2. Keeping a parallel `std::unordered_map<std::string_view, std::size_t>` that stores an index into the vector – still preserves contiguous storage while restoring O(1) average lookup.
> > > 
> > > Either approach keeps the external ABI identical while avoiding the worst-case O(n²) behaviour when all calls go through the registry.
> > > 
> > > `392-394`: **`get(std::string_view)` could return `end()` sentinel unconditionally**
> > > The new helper returns a `const_iterator`; good.  Small nit: returning `cend()` instead of `end()` ties the implementation to `const` use only.  If future non-const overloads are added you’ll need a second helper.  Returning `end()` keeps it generic without sacrificing `const`-correctness.
> > > Not blocking, just something to keep in mind.
> 
> 📜 Review details

@coderabbitai What race-condition are you referring to in this comment:
> src/spider/worker/FunctionManager.hpp (1)
> 381-386: Same race-condition applies to register_function_invoker – see previous comment.

Can you link to the original comment?

davidlion (approved):

PR #137, opened 2025-05-12
feat: Add job cancellation for early stop of the job execution.

sitaowang1998:
# Description
User might want to stop a job from executing if the task fails and has no hope of recovery or because of other reason, so `Spider` need to support job cancellation both from inside inside the task through `TaskContext::abort` and from user through `Job::cancel`. The cancelled job will be in `JobStatus::Cancelled` state.

User also need to get the reason why a job is cancelled. This PR adds `job_errors` table to store the cancellation messages. The messages are set inside `TaskContext::abort` as a user argument and set to `Job cancelled by user.` inside `Job::cancel`. The table also stores the function name of the task calling `TaskContext::abort`, or `user` if the job is cancelled using `Job::cancel`. User could retrieve both information using `Job::get_error`.

# Checklist

  breaking change.

# Validation performed

* [x] GitHub workflows pass.
* [x] Unit tests pass in dev container.
* [x] Integration tests pass in dev container.

## Summary by CodeRabbit

- **New Features**
  - Added job and task cancellation methods with detailed error message retrieval.
  - Introduced thread-safe management of task execution and cancellation control.
  - Extended database schema with a `job_errors` table to store cancellation details.
  - Added new task cancellation and abort test tasks.
- **Bug Fixes**
  - Enhanced task state handling to properly reflect cancellations and error conditions.
- **Tests**
  - Added comprehensive unit and integration tests for cancellation workflows.
  - Included a new cancellation test executable with command-line interface.
- **Refactor**
  - Renamed task state check methods for improved clarity and consistency.
- **Chores**
  - Updated build and test configurations to support new features and tests.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1210104031287392

### DIFF
--- a/src/spider/client/Job.hpp
+++ b/src/spider/client/Job.hpp
@@ -166,7 +166,30 @@ public:
      * @throw spider::ConnectionException
      */
     auto get_error() -> std::pair<std::string, std::string> {
-        throw ConnectionException{"Not implemented"};
+        if (nullptr == m_conn) {
+            std::variant<std::unique_ptr<core::StorageConnection>, core::StorageErr> conn_result
+                    = m_storage_factory->provide_storage_connection();
+            if (std::holds_alternative<core::StorageErr>(conn_result)) {
+                throw ConnectionException(std::get<core::StorageErr>(conn_result).description);
+            }
+            auto conn = std::move(std::get<std::unique_ptr<core::StorageConnection>>(conn_result));
+
+            std::pair<std::string, std::string> res;
+            core::StorageErr const err
+                    = m_metadata_storage->get_job_message(*conn, m_id, &res.first, &res.second);
+            if (false == err.success()) {
+                throw ConnectionException{err.description};
+            }
+            return res;
+        }
+
+        std::pair<std::string, std::string> res;
+        core::StorageErr const err
+                = m_metadata_storage->get_job_message(*m_conn, m_id, &res.first, &res.second);
+        if (false == err.success()) {
+            throw ConnectionException{err.description};
+        }
+        return res;
     }
 
 private:
--- a/src/spider/client/Job.hpp
+++ b/src/spider/client/Job.hpp
@@ -177,7 +177,7 @@ public:
 
             std::pair<std::string, std::string> res;
             core::StorageErr const err
-                    = m_metadata_storage->get_job_message(*conn, m_id, &res.first, &res.second);
+                    = m_metadata_storage->get_error_message(*conn, m_id, &res.first, &res.second);
             if (false == err.success()) {
                 throw ConnectionException{err.description};
             }
@@ -186,7 +186,7 @@ public:
 
         std::pair<std::string, std::string> res;
         core::StorageErr const err
-                = m_metadata_storage->get_job_message(*m_conn, m_id, &res.first, &res.second);
+                = m_metadata_storage->get_error_message(*m_conn, m_id, &res.first, &res.second);
         if (false == err.success()) {
             throw ConnectionException{err.description};
         }
--- a/src/spider/client/Job.hpp
+++ b/src/spider/client/Job.hpp
@@ -90,8 +90,7 @@ public:
         }
         auto conn = std::move(std::get<std::unique_ptr<core::StorageConnection>>(conn_result));
 
-        core::StorageErr const err
-                = m_metadata_storage->cancel_job(*conn, m_id, "Job cancelled by user");
+        core::StorageErr const err = m_metadata_storage->cancel_job(*conn, m_id);
         if (!err.success()) {
             throw ConnectionException(err.description);
         }
--- a/src/spider/client/Job.hpp
+++ b/src/spider/client/Job.hpp
@@ -82,7 +82,20 @@ public:
      *
      * @throw spider::ConnectionException
      */
-    auto cancel();
+    auto cancel() -> void {
+        std::variant<std::unique_ptr<core::StorageConnection>, core::StorageErr> conn_result
+                = m_storage_factory->provide_storage_connection();
+        if (std::holds_alternative<core::StorageErr>(conn_result)) {
+            throw ConnectionException(std::get<core::StorageErr>(conn_result).description);
+        }
+        auto conn = std::move(std::get<std::unique_ptr<core::StorageConnection>>(conn_result));
+
+        core::StorageErr const err
+                = m_metadata_storage->cancel_job(*conn, m_id, "Job cancelled by user");
+        if (!err.success()) {
+            throw ConnectionException(err.description);
+        }
+    }
 
     /**
      * @return Status of the job.
--- a/src/spider/client/Job.hpp
+++ b/src/spider/client/Job.hpp
@@ -92,7 +92,8 @@ public:
         }
         auto conn = std::move(std::get<std::unique_ptr<core::StorageConnection>>(conn_result));
 
-        core::StorageErr const err = m_metadata_storage->cancel_job(*conn, m_id);
+        core::StorageErr const err
+                = m_metadata_storage->cancel_job_by_user(*conn, m_id, "Job cancelled by user.");
         if (!err.success()) {
             throw ConnectionException(err.description);
         }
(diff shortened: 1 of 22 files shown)


davidlion (changes requested):

src/spider/storage/MetadataStorage.hpp
```diff
@@ -72,6 +72,41 @@ class MetadataStorage {
...
     ) -> StorageErr
             = 0;
+    /**
+     * Cancel a job. This will set the job state to CANCEL and set all tasks that have not
+     * finished or started to CANCEL.
+     * @param conn
+     * @param id The job id.
+     * @return The error code.
+     */
+    virtual auto cancel_job(StorageConnection& conn, boost::uuids::uuid id) -> StorageErr = 0;
```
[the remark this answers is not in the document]
sitaowang1998: No. `cancel_job` is used by client to cancel a job without any error messages.

tests/integration/test_cancel.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def start_scheduler_worker(
+    storage_url: str, scheduler_port: int
+) -> Tuple[subprocess.Popen, subprocess.Popen]:
+    # Start the scheduler
+    dir_path = Path(__file__).resolve().parent
+    dir_path = dir_path / ".." / ".." / "src" / "spider"
+    scheduler_cmds = [
+        str(dir_path / "spider_scheduler"),
```
davidlion, line 37: Why are we running from the src dir? Shouldn't it be in the build dir?

tests/integration/test_cancel.py
davidlion, line 1: @coderabbitai Please create an issue to call the integration tests' task (`test:integration`) from the task `test:all`. We can create a separate task to only run all unit tests, but `all` should truly run all tests.

src/spider/worker/worker.cpp
```diff
@@ -123,15 +124,49 @@ auto get_environment_variable() -> absl::flat_hash_map<
...
+    if (false == err.success()) {
+        spdlog::error("Failed to get task state: {}", err.description);
+        return;
+    }
+
+    if (spider::core::TaskState::Canceled != task_state) {
+        return;
+    }
+
+    // Cancel the task.
```
davidlion, line 155: As mentioned many times before, please avoid comments like this. I've linked the guidelines in other posts, but the bottom line is that if code is so confuing it needs a comment we should first try to refactor or rename it. If a single line of code is that confusing then it is almost certainly named incorrectly.

src/spider/worker/worker.cpp
```diff
@@ -123,15 +124,49 @@ auto get_environment_variable() -> absl::flat_hash_map<
...
+    spider::core::TaskState task_state = spider::core::TaskState::Running;
+    spider::core::StorageErr err = metadata_store->get_task_state(*conn, task_id, &task_state);
+    if (false == err.success()) {
+        spdlog::error("Failed to get task state: {}", err.description);
+        return;
+    }
+
+    if (spider::core::TaskState::Canceled != task_state) {
+        return;
+    }
```
davidlion, line 153: This function can fail so we should return an error code.

src/spider/worker/ExecutorHandle.hpp
```diff
@@ -0,0 +1,33 @@
...
+    [[nodiscard]] auto get_executor() -> TaskExecutor*;
+    auto executor_cancel() -> void;
+    auto set(boost::uuids::uuid task_id, TaskExecutor* executor) -> void;
+    auto clear() -> void;
+
+private:
+    boost::uuids::uuid m_task_id;
+    TaskExecutor* m_executor
+            = nullptr;  // Do not use std::shared_ptr to avoid calling destructor twice.
+    std::mutex m_mutex;
```
davidlion, line 29: https://www.notion.so/yscope/WIP-Coding-Guidelines-9a308b847a5343958ba3cb97a850be66?source=copy_link#14a04e4d9e6b801dac69d6222e8885a2
```suggestion
    boost::uuids::uuid m_task_id;

    // Do not use std::shared_ptr to avoid calling destructor twice.
    TaskExecutor* m_executor = nullptr;

    std::mutex m_mutex;
```

src/spider/worker/ExecutorHandle.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include <mutex>
+#include <optional>
+
+#include <boost/uuid/uuid.hpp>
+
+#include "TaskExecutor.hpp"
+
+namespace spider::worker {
+/**
+ * This class acts as a handle for thread-safe access to the task executor and task id.
```
davidlion, line 13: If I understand correctly, a TaskExecutor will only ever own a single task (i.e. I don't think it is possible to re-use a TaskExecutor for a new task or for it to switch tasks). If I'm correct, we can just store the task ID when we create the TaskExecutor to avoid extra working keeping them in sync.

src/spider/worker/ExecutorHandle.hpp
davidlion, line 29: With this design it seems easy to accidentally create multiple handles (and therefore mutexes) for the same TaskExecutor.

Is it possible to design things in a way where a handle has unique ownership of the TaskExecutor? It feels possible since currently it looks like each thread gets a reference to the handle.

src/spider/worker/ExecutorHandle.cpp
```diff
@@ -0,0 +1,42 @@
...
+    if (nullptr != m_executor) {
+        return m_task_id;
+    }
+    return std::nullopt;
+}
+
+auto ExecutorHandle::get_executor() -> TaskExecutor* {
+    std::lock_guard const lock_guard{m_mutex};
+    return m_executor;
+}
```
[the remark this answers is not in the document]
davidlion: This is a valid point, why was it marked resolved? What is the use case for getting the TaskExecutor itself?

davidlion (changes requested):
In this PR and future PRs, we need to try and do a better job of explaining what and why in PR descriptions and titles. This is important as we want other engineers in YScope to reference these PRs and understand them without needing to read all the code.

Currently, both the title and description just list the names of things that were added with little to no explanation on their motivation, usage, or design.

src/spider/client/Job.hpp
```diff
@@ -84,7 +84,19 @@ class Job {
...
      * @throw spider::ConnectionException
      */
-    auto cancel();
+    auto cancel() -> void {
+        std::variant<std::unique_ptr<core::StorageConnection>, core::StorageErr> conn_result
+                = m_storage_factory->provide_storage_connection();
+        if (std::holds_alternative<core::StorageErr>(conn_result)) {
+            throw ConnectionException(std::get<core::StorageErr>(conn_result).description);
+        }
+        auto conn = std::move(std::get<std::unique_ptr<core::StorageConnection>>(conn_result));
```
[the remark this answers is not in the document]
sitaowang1998: @coderabbit Please create an issue to refactor getting storage connection into a helper function.

src/spider/client/Job.hpp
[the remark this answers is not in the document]
sitaowang1998: This is because `Job` is a templated class, so all methods must be in the header file.

src/spider/client/Job.hpp
```diff
@@ -155,7 +167,30 @@ class Job {
      * @throw spider::ConnectionException
      */
     auto get_error() -> std::pair<std::string, std::string> {
```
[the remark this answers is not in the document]
sitaowang1998: Actually it is undefined when calling this if job is not in the cancelled state. In the other state, calling `get_error` will throw a `spider::ConnectionException` because there is not error message available. Ideally we can use `ystdlib::Result` to indicate the absence of an error message.

tests/worker/worker-test.cpp
```diff
@@ -70,6 +72,16 @@ auto join_string_test(
     return input_1 + input_2;
 }
 
+auto sleep_test(spider::TaskContext& /*context*/, int milliseconds) -> int {
```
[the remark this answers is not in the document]
sitaowang1998: @coderabbit Please open an issue to use `[[maybe_unused]]` for the rest of the project where parameters are ignored.

LinZhihao-723:
Closed as we stop maintaining Spider Wolf.

PR #141, opened 2025-05-20
refactor: Update include statements to match our latest style guidelines (resolves #140). 

### LINKED ISSUE #140: Use non-relative include paths for headers outside of individual CMake targets
## Problem

When using relative include paths (e.g., `#include "../core/Context.hpp"`) for headers outside of an individual CMake target, those includes will break if the target is installed. This is because the directory structure changes during installation, and relative paths no longer resolve correctly.

For example, in `src/spider/client/Data.hpp`, we currently use:
```cpp
#include "../core/Context.hpp"
#include "../core/DataCleaner.hpp"
```

If the Spider library is installed, these paths would be invalid since the installed headers wouldn't maintain the same relative directory structure.

## Solution

Replace all relative include paths with paths that start from the project's base include directory:
```cpp
#include "spider/core/Context.hpp"
#include "spider/core/DataCleaner.hpp"
```

This would work both during development and after installation, provided that the project's base directory is properly set in the include path.

## Implementation

1. Update all header files to use non-relative include paths
2. Configure CMake to properly set include directories

## Related Issues

This issue is related to #138 (Make spider as an installable package)

sitaowang1998:
# Description

Use non-relative header include paths for `src/spider` and `tests`.

# Checklist

  breaking change.

# Validation performed

* [x] GitHub workflows pass.

## Summary by CodeRabbit

- **Refactor**
  - Standardized all internal include paths across the codebase and tests to use fully qualified, project-root-relative paths (e.g., spider/core/...), replacing previous relative paths for improved clarity and consistency.
  - Updated build configuration files to ensure correct include directories are set for both source and test targets.
  - Enhanced code formatting rules to better distinguish project relative headers from project library headers.

- **Chores**
  - No changes to functionality or public interfaces; all updates are internal refactoring for maintainability.

### DIFF
--- a/src/spider/.clang-format
+++ b/src/spider/.clang-format
@@ -2,7 +2,6 @@ BasedOnStyle: "InheritParentConfig"
 
 IncludeCategories:
   # NOTE: A header is grouped by first matching regex
-
   # Project library headers
   - Regex: "^<spider"
     Priority: 4
--- a/src/spider/.clang-format
+++ b/src/spider/.clang-format
@@ -2,6 +2,12 @@ BasedOnStyle: "InheritParentConfig"
 
 IncludeCategories:
   # NOTE: A header is grouped by first matching regex
+  # Project relative headers
+  - Regex: "^\"spider"
+    Priority: 5
+  # Project library headers
+  - Regex: "^<spider"
+    Priority: 4
   # Library headers. Update when adding new libraries.
   # NOTE: clang-format retains leading white-space on a line in violation of the YAML spec.
   # Ex:
@@ -15,9 +21,3 @@ IncludeCategories:
   # C++ standard libraries
   - Regex: "^<.+>"
     Priority: 2
-  # Project library headers
-  - Regex: "^<spider"
-    Priority: 4
-  # Project relative headers
-  - Regex: "^\"spider"
-    Priority: 5
--- a/src/spider/.clang-format
+++ b/src/spider/.clang-format
@@ -2,9 +2,6 @@ BasedOnStyle: "InheritParentConfig"
 
 IncludeCategories:
   # NOTE: A header is grouped by first matching regex
-  # Project headers
-  - Regex: "^<spider"
-    Priority: 4
   # Library headers. Update when adding new libraries.
   # NOTE: clang-format retains leading white-space on a line in violation of the YAML spec.
   # Ex:
@@ -18,3 +15,9 @@ IncludeCategories:
   # C++ standard libraries
   - Regex: "^<.+>"
     Priority: 2
+  # Project library headers
+  - Regex: "^<spider"
+    Priority: 4
+  # Project relative headers
+  - Regex: "^\"spider"
+    Priority: 5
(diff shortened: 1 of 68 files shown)


anlowee (dismissed):

src/spider/.clang-format
davidlion, line 1: Since we're re-ordering the priorities lets just re-order them all to be descending. (Lets also move priority 3.)

davidlion (commented):
As we discussed let's change the project library header include statements to use `<>` rather than `""`.

src/spider/.clang-format
sitaowang1998: Actually the current order is wrong but we cannot use descending order, as `clang-format` uses the first match.

davidlion: Ah right, good catch sorry.

src/spider/.clang-format
```diff
@@ -2,7 +2,8 @@ BasedOnStyle: "InheritParentConfig"
 
 IncludeCategories:
   # NOTE: A header is grouped by first matching regex
-  # Project headers
+
```
davidlion, line 5:
```suggestion
```

tests/.clang-format
```diff
@@ -2,19 +2,24 @@ BasedOnStyle: "InheritParentConfig"
 
 IncludeCategories:
   # NOTE: A header is grouped by first matching regex
-  # Project headers
+  # Project library headers
   - Regex: "^<spider"
     Priority: 4
+  - Regex: "^<tests"
+    Priority: 4
```
davidlion, line 9:
```suggestion
  # Project library headers
  - Regex: "^<(spider|tests)"
    Priority: 4
```

davidlion (approved):

src/spider/.clang-format
```diff
@@ -2,7 +2,7 @@ BasedOnStyle: "InheritParentConfig"
 
 IncludeCategories:
   # NOTE: A header is grouped by first matching regex
-  # Project headers
+  # Project library headers
   - Regex: "^<spider"
     Priority: 4
   # Library headers. Update when adding new libraries.
```
davidlion, line 8: Sorry I just realized all of the .clang-format files have a weird issue where the project library headers split up the comment. Unfortunately, I can't make the full suggestion.

It should be:
```
  # NOTE: A header is grouped by first matching regex library headers. Update when adding new
  # libraries.
  # NOTE: clang-format retains leading white-space on a line in violation of the YAML spec.
  # Project library headers
  - Regex: "^<spider"
    Priority: 4
  # External library headers
  - Regex: "^<(absl|boost|catch2|fmt|mariadb|msgpack|spdlog|ystdlib)"
    Priority: 3
  # C system headers
```

davidlion (changes requested):

src/spider/.clang-format
sitaowang1998: I believe the "Update when adding new libraries" is a comment for external library headers only. Thus I move it there.

davidlion: Good catch. Thanks!

PR #142, opened 2025-05-21
feat: Add data persistence.

sitaowang1998:
# Description

Add support for client to set a `Data` as persisted when building the data or through `Data::set_persisted` at runtime. The `Data` marked persisted will not be garbage collected.

# Checklist

  breaking change.

# Validation performed

* [x] GitHub workflows pass.
* [x] Unit tests pass in dev container.
* [x] Integration tests pass in dev container.

## Summary by CodeRabbit

- **New Features**
	- Added the ability to mark data as persisted, preventing it from being cleaned up.
	- Users can now set and update the persisted state for data both during creation and afterwards.
- **Bug Fixes**
	- Improved test coverage to verify correct handling of the persisted state in data storage and retrieval.
- **Tests**
	- Enhanced tests to ensure persisted state is correctly set, updated, and validated during data operations.

### DIFF
--- a/src/spider/client/Data.hpp
+++ b/src/spider/client/Data.hpp
@@ -80,11 +80,11 @@ public:
     }
 
     /**
-     * Sets the data as checkpointed, indicating the data should not be cleaned up.
+     * Sets the data as persisted, indicating the data should not be cleaned up.
      *
      * @throw spider::ConnectionException
      */
-    void set_checkpointed() {
+    void set_persisted() {
         m_impl->set_persisted(true);
         if (nullptr != m_connection) {
             m_data_store->set_data_persisted(*m_connection, *m_impl);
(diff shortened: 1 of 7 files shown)


davidlion (changes requested):
1. Add a real PR description.
2. Is there a reason we use both "checkpointed" and "persisted" as terms for this? Are there plans for different types of "persistent" data? It would be less complex to stick to a single term.

sitaowang1998:
>     2. Is there a reason we use both "checkpointed" and "persisted" as terms for this? Are there plans for different types of "persistent" data? It would be less complex to stick to a single term.
For theoretical part, especially for the paper, all the task inputs/outputs that are plain value are considered persisted in the storage. However, in practice, these two terms are used interchangeably.

davidlion:
> > ```
> > 2. Is there a reason we use both "checkpointed" and "persisted" as terms for this? Are there plans for different types of "persistent" data? It would be less complex to stick to a single term.
> > ```
> 
> For theoretical part, especially for the paper, all the task inputs/outputs that are plain value are considered persisted in the storage. However, in practice, these two terms are used interchangeably.

I see. As discussed offline let's only use one for the code.

LinZhihao-723:
Closed as we stop maintaining Spider Wolf.

PR #186, opened 2025-08-08
feat(test): Add pytest and unit test structure in Python.

sitaowang1998:
# Description

> [!Note]
> This PR depends on #179.

This PR:
* Adds `pytest` as Python dev dependency.
* Adds a sample unit test.
* Adds tasks to run `pytest` and renames C++ test tasks.
* Updates developer doc with change in test tasks.
* Adds python tests in GitHub workflow.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* [x] Running python test tasks successfully executes the sample unit tests.
* [x] GitHub `non-storage-unit-tests` workflow executes python test tasks.
* [x] GitHub workflows pass.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

PR #187, opened 2025-08-09
feat: Add custom Spider types and converting Python types to TDL types.

sitaowang1998:
# Description

> [!Note]
> This PR depends on #186.

This PR
* Adds custom primitive Spider Python types.
* Add TDL types and their string representations.
* Adds converting Python types to TDL types.
* Adds tests for type convertion.

# Checklist

  breaking change.

# Validation performed

* [x] Unit tests for type conversion pass.
* [x] GitHub workflows pass.


python/spider-py/src/spider_py/type/tdl_type.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @override
+    def type_str(self) -> str:
+        return self.name
+
+
+class ListType(TdlType):
+    """TDL List type."""
+
+    def __init__(self, key: TdlType) -> None:
```
LinZhihao-723, line 90: Can we call it `element_type`?
* `element` to match the TDL compiler's AST node.
* `_type` to be a little more verbose, indicating this is the type.

python/spider-py/src/spider_py/type/tdl_type.py
```diff
@@ -0,0 +1,130 @@
...
+
+class ClassType(TdlType):
+    """TDL Custom class type."""
+
+    def __init__(self, name: str) -> None:
+        """
+        Creates a TDL custom class type.
+        :param name: The name of the class.
+        """
+        self.name = name
```
LinZhihao-723, line 80: This is also for any future PR: for "private" member variables, can we add `_` prefix to match the guideline? So `self.name` -> `self._name`.

python/spider-py/src/spider_py/type/tdl_type.py
```diff
@@ -0,0 +1,130 @@
...
+    def __init__(self, key: TdlType, value: TdlType) -> None:
+        """
+        Creates a TDL map type.
+        :raises TypeError: If key is not a supported type.
+        """
+        if not is_map_key(key):
+            msg = f"{key} is not a supported type for map key."
+            raise TypeError(msg)
+        self.key = key
+        self.value = value
```
LinZhihao-723, line 126: Similarly, can we add `_type` postfix?

python/spider-py/src/spider_py/type/type.py
```diff
@@ -0,0 +1,66 @@
+"""Custom type module for Spider."""
+
+from typing import cast
+
+
+class BoundedInt(int):
+    """Bounded integer type."""
+
+    def __new__(cls, value: int, bits: int = 32) -> "BoundedInt":
```
LinZhihao-723, line 9: Can we use [this](https://github.com/y-scope/clp-ffi-py/blob/795ee4c43437de2f3e314847a68d9a2aad85f317/clp_ffi_py/ir/query_builder.py#L1) so that the type can be properly annotated as the class itself?

python/spider-py/src/spider_py/type/type.py
```diff
@@ -0,0 +1,66 @@
...
+        """Creates a bounded integer."""
+        min_val = -(1 << (bits - 1))
+        max_val = (1 << (bits - 1)) - 1
+
+        if not (min_val <= value <= max_val):
```
LinZhihao-723, line 14: I understand we discussed this offline today, but can we write it in a non-Pythonic way to improve readability? lol

python/spider-py/src/spider_py/type/type.py
```diff
@@ -0,0 +1,66 @@
...
+            msg = f"Bounded integer value ({value}) must be between {min_val} and {max_val}"
+            raise ValueError(msg)
```
LinZhihao-723, line 16: We don't need `msg` variable, right?

LinZhihao-723 (changes requested):
First batch of comments. Still need to read sth about Python's type system to review the rest.

python/spider-py/src/spider_py/type/type.py
sitaowang1998: `ruff` complains about directly using f-string or even a string literal inside an exception constructor.

PR #188, opened 2025-08-09
feat: Add TDL type parsing and conversion from TDL types to native Python types.

sitaowang1998:
# Description

> [!Note]
> This PR depends on #187.

This PR:
* Adds lark parser as dependency.
* Adds TDL type parsing using lark.
* Adds conversion from TDL types to Python native types.
* Adds unit tests for parsing and conversion.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* [x] Newly added unit tests for TDL parsing and conversion to Python type pass.
* [x] GitHub workflows pass.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

PR #180, opened 2025-08-06
feat(spider-py): Add `Data`, `Task`, and `TaskGraph` to core.

sitaowang1998:
# Description

> [!Note]
> This PR depends on #179.

This PR adds core Python Data, Task and TaskGraph classes for later use of client and task executor.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* [x] GitHub workflows pass.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

- New Features
  - Added foundational data and task primitives (IDs, data containers, inputs/outputs, states) and a Task model with defaults.
  - Added a task-graph utility to model tasks as a DAG with parent/child wiring and helpers to query connections.
- Refactor
  - Centralized and expanded the core public API for simpler imports.
- Other
  - Additive changes only; no runtime behaviour changes.

PR #191, opened 2025-08-11
feat(spider-py): Add support for client-end task graph grouping and chaining.

sitaowang1998:
# Description

> [!Note]
> This PR depends on #180 and #188.

This PR adds two client task composition functions: `group` and `chain`. This PR also changes the underlying core Task and TaskGraph to better fit the requirement in `group` and `chain`.

# Checklist

  breaking change.

# Validation performed

* [x] Add unit tests for `group` and `chain`.
* [x] GitHub workflows pass.

## Summary by CodeRabbit

- **New Features**
  - High-level TaskGraph API: group() and chain() for composing workflows; package root now exposes chain, group, TaskContext and TaskGraph.
  - Added a minimal Data container.

- **Behavior**
  - TaskGraph switched to an index-based representation; task inputs/outputs now use structured dataclasses and runtime signature validation translates type hints into task I/O.

- **Tests**
  - Added tests for grouping, chaining and signature validation (including linkage checks).

- **Chores**
  - Lint config relaxed for tests and added self-like name ignore.


LinZhihao-723 (changes requested):
Reviewd task graph implementation.

python/spider-py/src/spider_py/client/task.py
```diff
@@ -0,0 +1,101 @@
...
+    params = list(signature.parameters.values())
+    if not params or params[0].annotation is not TaskContext:
+        msg = "First argument is not a TaskContext."
+        raise TypeError(msg)
+    for param in params[1:]:
+        if param.kind in {inspect.Parameter.VAR_POSITIONAL, inspect.Parameter.VAR_KEYWORD}:
+            msg = "Variadic parameters are not supported."
+            raise TypeError(msg)
+        if param.annotation is inspect.Parameter.empty:
+            msg = "Argument must have type annotation"
```
LinZhihao-723, line 50:
```suggestion
            msg = "Parameters must have type annotation."
```
It is "parameter", right?
Also the period is missing.

python/spider-py/src/spider_py/client/task.py
```diff
@@ -0,0 +1,101 @@
...
+
+def _is_tuple(t: type | GenericAlias) -> bool:
+    """
+    :param t:
+    :return: Whether t is a tuple.
+    """
+    return get_origin(t) is tuple
+
+
+def _process_parameters(task: core.Task, signature: inspect.Signature) -> None:
```
LinZhihao-723, line 34: How about `_validate_and_set_params`?

python/spider-py/src/spider_py/client/task.py
```diff
@@ -0,0 +1,101 @@
...
+            msg = "Variadic parameters are not supported."
+            raise TypeError(msg)
+        if param.annotation is inspect.Parameter.empty:
+            msg = "Argument must have type annotation"
+            raise TypeError(msg)
+        tdl_type_str = to_tdl_type_str(param.annotation)
+        task.task_inputs.append(TaskInput(tdl_type_str, None))
+
+
+def _process_return(task: core.Task, signature: inspect.Signature) -> None:
```
LinZhihao-723, line 56: How about `_validate_and_set_return`?

python/spider-py/src/spider_py/client/task.py
```diff
@@ -0,0 +1,101 @@
...
+            if r is Data:
+                task.task_outputs.append(TaskOutput(tdl_type_str, TaskOutputData()))
+            else:
+                task.task_outputs.append(TaskOutput(tdl_type_str, TaskOutputValue()))
+    else:
+        tdl_type_str = to_tdl_type_str(returns)
+        if returns is Data:
+            task.task_outputs.append(TaskOutput(tdl_type_str, TaskOutputData()))
+        else:
+            task.task_outputs.append(TaskOutput(tdl_type_str, TaskOutputValue()))
```
LinZhihao-723, line 83:
```suggestion

    if not _is_tuple(returns):
        tdl_type_str = to_tdl_type_str(returns)
        if returns is Data:
            task.task_outputs.append(TaskOutput(tdl_type_str, TaskOutputData()))
        else:
            task.task_outputs.append(TaskOutput(tdl_type_str, TaskOutputValue()))
        return

    args = get_args(returns)
    if Ellipsis in args:
        msg = "Variable-length tuple return types are not supported."
        raise TypeError(msg)
    for arg in args:
        tdl_type_str = to_tdl_type_str(arg)
        if arg is Data:
            task.task_outputs.append(TaskOutput(tdl_type_str, TaskOutputData()))
        else:
            task.task_outputs.append(TaskOutput(tdl_type_str, TaskOutputValue()))
```
* Let's try to avoid long if-else cluster. Use early return to simply it so that it's more readable.
* `r` -> `arg` as: (1) we usually don't allow single char variables except looping index like `i`; (2) `r` itself as `return` can be confusing.

python/spider-py/src/spider_py/client/task.py
```diff
@@ -0,0 +1,101 @@
...
+def _process_return(task: core.Task, signature: inspect.Signature) -> None:
+    """
+    Checks the return type validity and add them to the task.
+    :param task:
+    :param signature:
+    :raises TypeError: If the return type is invalid.
+    """
+    returns = signature.return_annotation
+    if returns is inspect.Parameter.empty:
+        msg = "Return type must have type annotation"
```
LinZhihao-723, line 65:
```suggestion
        msg = "Return type must have type annotation."
```

python/spider-py/src/spider_py/client/task.py
LinZhihao-723, line 1: I feel like we should move these utility functions into core's task so that the creation of a task is more like RAII. The parameter and return type checks can be inside `__init__`, right?

LinZhihao-723 (changes requested):
Reviewed the updated task graph implementation and the task implementation.

python/spider-py/src/spider_py/core/task_graph.py
```diff
@@ -1,62 +1,150 @@
...
+                graph.task_input_output_refs.append(
+                    (input_task_index, i, output_task_index, output_position)
+                )
+                output_position += 1
+                if output_position >= len(graph.tasks[output_task_index].task_outputs):
+                    output_position = 0
+                    task_output_index += 1
+
+        if task_output_index != len(parent_output_tasks) or output_position != 0:
+            raise TypeError(size_mismatch_msg)
```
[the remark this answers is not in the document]
sitaowang1998: Right now we don't support zero output in Spider.

python/spider-py/src/spider_py/client/task.py
```diff
@@ -0,0 +1,101 @@
...
+
+class TaskContext:
+    """Spider task context."""
+
+    # TODO: Implement task context for use in task executor
+
+
+# Check the TaskFunction signature at runtime.
+# Enforcing static check for first argument requires the use of Protocol. However, functions, which
+# are Callable, are not considered a Protocol without explicit cast.
```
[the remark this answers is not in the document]
sitaowang1998: This comment is not a docstring but to explain why I choose the current approach, or more precisely why other ways don't work.
I have tried to enforce the `TaskFunction` type check, i.e. the first parameter must be a `spider.TaskContext`, using type annotations.
1. `Callable` requires either specifying all argument or just use a variadic. It does not support specifying the first parameter type and then a variadic.
2. `Protocol` supports the check if a function starts with a specific type. However, `Protocol` is not considered a `Callable`, and I cannot pass in a function where the type annotation is a `Protocol`.

python/spider-py/src/spider_py/client/task.py
```diff
@@ -0,0 +1,109 @@
...
```
LinZhihao-723, line 22:
```suggestion
# NOTE: This type alias is for clarification purposes only. It does not enforce static type checks.
# Instead, we rely on the runtime check to ensure the first argument is `TaskContext`. To statically
# enforce the first argument to be `TaskContext`, `Protocol` is required, which is not compatible
# with `Callable` without explicit type casting.
```
As a follow up to https://github.com/y-scope/spider/pull/191/files#r2312686452, I made this revising to clarify things a little bit. Notice that for this type of comments, we should start with `NOTE`.

python/spider-py/src/spider_py/client/task.py
```diff
@@ -0,0 +1,109 @@
...
+    :return: Whether t is a tuple.
+    """
+    return get_origin(t) is tuple
+
+
+def _validate_and_convert_params(signature: inspect.Signature) -> list[TaskInput]:
+    """
+    Checks the parameters validity and convert them into core.TaskInput.
+    :param signature:
+    :return: Converted task parameters.
```
LinZhihao-723, line 38:
```suggestion
    Validates the task parameters and converts them into a list of `core.TaskInput`.
    :param signature:
    :return: The converted task parameters.
```
core.TaskInput should be surrounded by `

python/spider-py/src/spider_py/client/task.py
```diff
@@ -0,0 +1,109 @@
...
+        tdl_type_str = to_tdl_type_str(param.annotation)
+        inputs.append(TaskInput(tdl_type_str, None))
+    return inputs
+
+
+def _validate_and_convert_return(signature: inspect.Signature) -> list[TaskOutput]:
+    """
+    Checks the return type validity and add them to the task.
+    :param signature:
+    :return: Converted task returns.
```
LinZhihao-723, line 62:
```suggestion
    Validates the task returns and converts them into a list of `core.TaskOutput`.
    :param signature:
    :return: The converted task returns.
```
Same.

python/spider-py/src/spider_py/client/task.py
```diff
@@ -0,0 +1,109 @@
...
+        else:
+            outputs.append(TaskOutput(tdl_type_str, TaskOutputValue()))
+    return outputs
+
+
+def create_task(func: TaskFunction) -> core.Task:
+    """
+    Creates a core Task object from the task function.
+    :param func:
+    :return:
```
LinZhihao-723, line 96:
```suggestion
    :return: The created core Task object.
```
As suggested in previous comments, please don't leave the return field empty.

python/spider-py/src/spider_py/client/task.py
```diff
@@ -0,0 +1,109 @@
...
+        msg = "`func` is not a function."
+        raise TypeError(msg)
+    signature = inspect.signature(func)
+    inputs = _validate_and_convert_params(signature)
+    outputs = _validate_and_convert_return(signature)
+    return core.Task(
+        function_name=func.__qualname__,
+        task_inputs=inputs,
+        task_outputs=outputs,
+    )
```
LinZhihao-723, line 109:
```suggestion
    return core.Task(
        function_name=func.__qualname__,
        task_inputs=_validate_and_convert_params(signature),
        task_outputs=_validate_and_convert_return(signature),
    )
```
Nit: Using "rvalue" for initialization in Python is more "safe", as the init essentially would just do a shallow copy. Using the "rvalue" avoids the list itself gets modified after use for initialization. (in this case it's nit because this is a return statement)

LinZhihao-723 (changes requested):
Last few comments.

LinZhihao-723 (approved):
For the PR title, how about:
```
feat(spider-py): Add support for client-end task graph grouping and chaining.
```

PR #193, opened 2025-08-14
feat(tdl): Add `Struct` AST node. (#190)

LinZhihao-723:
# Description

This PR adds `Struct` AST node as an implementation of `Type`. A `Struct` has only one child: the identifier. It also holds a reference to a `StructSpec` as a shared pointer. However, such a spec is not initialized during the construction. This is because the AST nodes are constructed at the parsing stage. The spec should be checked and set in the semantic analysis stage, while a symbol table is built.

# Checklist

  breaking change.

# Validation performed

* [ ] Add unit tests to cover the basic behavior of `Struct`.

### DIFF
--- a/src/spider/tdl/parser/ast/node_impl/type_impl/Struct.cpp
+++ b/src/spider/tdl/parser/ast/node_impl/type_impl/Struct.cpp
@@ -1,0 +1,72 @@
+#include "Struct.hpp"
+
+#include <cstddef>
+#include <memory>
+#include <string>
+#include <utility>
+
+#include <fmt/format.h>
+#include <ystdlib/error_handling/ErrorCode.hpp>
+#include <ystdlib/error_handling/Result.hpp>
+
+#include <spider/tdl/parser/ast/Node.hpp>
+#include <spider/tdl/parser/ast/node_impl/StructSpec.hpp>
+#include <spider/tdl/parser/ast/utils.hpp>
+
+using spider::tdl::parser::ast::node_impl::type_impl::Struct;
+using StructErrorCodeCategory = ystdlib::error_handling::ErrorCategory<Struct::ErrorCodeEnum>;
+
+template <>
+auto StructErrorCodeCategory::name() const noexcept -> char const* {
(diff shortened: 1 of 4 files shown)


src/spider/tdl/parser/ast/node_impl/type_impl/Struct.cpp
```diff
@@ -0,0 +1,72 @@
...
+            "{}[Type[Struct]]:\n{}Name:\n{}",
+            create_indentation(indentation_level),
+            create_indentation(indentation_level + 1),
+            YSTDLIB_ERROR_HANDLING_TRYX(
+                    // The factory function ensures that the first child is of type `Identifier`.
+                    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-static-cast-downcast)
+                    get_child_unsafe(0)->serialize_to_str(indentation_level + 2)
+            )
+    );
+}
```
sitaowang1998, line 58: Should we also add the `StructSpec` in the serialization if not null?

tests/tdl/test-parser-ast.cpp
```diff
@@ -326,35 +328,71 @@ TEST_CASE("test-ast-node", "[tdl][ast][Node]") {
...
+
+            SECTION("Struct serialization") {
+                constexpr std::string_view cExpectedSerializedResult{"[Type[Struct]]:\n"
+                                                                     "  Name:\n"
+                                                                     "    [Identifier]:TestStruct"};
+                auto const serialized_result{struct_node->serialize_to_str(0)};
+                REQUIRE_FALSE(serialized_result.has_error());
+                REQUIRE(serialized_result.value() == cExpectedSerializedResult);
+            }
+
```
sitaowang1998, line 385: If you add the not-null `StructSpec` serialization suggested above, add a test for it.

sitaowang1998 (approved):
LGTM. It would be nice to have the last coderabbit suggestion.

sitaowang1998 (approved):

PR #196, opened 2025-08-14
feat(tdl): Add `Namespace` AST node.

LinZhihao-723:
# Description

This PR adds `Namespace` AST node. In TDL definition, a namespace contains an identifier and a list of functions.

# Checklist

  breaking change.

# Validation performed

* [x] Add unit tests to cover basic behaviors for `Namespace`.


tests/tdl/test-parser-ast.cpp
```diff
@@ -668,6 +694,73 @@ TEST_CASE("test-ast-node", "[tdl][ast][Node]") {
...
+                    Namespace::create(Identifier::create(std::string{cTestNamespaceName}), {})
+            };
+            REQUIRE(namespace_result.has_error());
+            REQUIRE(namespace_result.error()
+                    == Namespace::ErrorCode{Namespace::ErrorCodeEnum::EmptyNamespace});
+        }
+
+        SECTION("Duplicated names") {
+            // The execution model of `SECTION` ensures objects are not moved when this section is
+            // executed, so use and move `functions` is safe.
```
sitaowang1998, line 751: This explanation is confusing and misleading. Please improve it.

sitaowang1998 (changes requested):

tests/tdl/test-parser-ast.cpp
LinZhihao-723: As discussed offline, we will use the following to annotate this suppression:
```
// The `SECTION` execution model ensures that objects are not reused across
// parallel `SECTION`s. Variables in different `SECTION`s are independent.
// Suppress warnings about potential use-after-move, as this is intentional.
```
I will apply this new comment to other occurrences in the unit test.

LinZhihao-723:
> Could we change the title and description to indicate the change of comment?

Hmmm, I usually don't include this in the title. This change is more like a refactoring of some existing pattern when adding new code. And the refactoring is just to make things consistent. This change from a PR's title perspective could be irrelevant.
I updated the PR description to include the changes.

PR #208, opened 2025-08-22
feat!: Add task language column to MariaDB storage for multi-language support.

sitaowang1998:
# Description

> [!Note]
> This PR depends on #207.

This PR
* Adds a new `language` column in the task table so that task executor knows the language of the task and starts the appropriate task executor (This feature will be introduced later in other PR).
* Adds `StorageErrorCode` alongside with existing `StorageErr` and its usage in storage for better error handling.
* Removes an unused and untested function in storage.
* Updates the unit tests to add check for the language get from the storage is correct.
* Updates Python MariaDB storage backend for the language column.

# Checklist

  breaking change.

# Validation performed

* [x] C++ unit tests and integration tests pass in Ubuntu 24.04.
* [x] Python unit tests pass in Ubuntu 24.04.
* [x] GitHub workflows pass.

## Summary by CodeRabbit

- **New Features**
  - Tasks now record execution language (C++ or Python) with getter/setter; language is preserved across task graphs and retrievals.

- **Database**
  - Added non-null language column (ENUM 'cpp','python') to tasks; schema and insert/select paths updated.

- **Improvements**
  - Storage flows propagate language and surface explicit errors for unknown mappings.

- **Refactor**
  - Removed legacy child-task insertion; storage helpers converted to result-oriented, batch-aware APIs.

- **Tests**
  - Task equality comparison now includes language.


src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -80,6 +80,27 @@ auto get_sql_string(sql::SQLString const& str) -> std::string {
 
 // NOLINTEND
 
+auto task_language_to_string(spider::core::TaskLanguage language) -> std::string {
```
LinZhihao-723, line 83: We should add `[[nodiscard]]` (even though I realized the rest of functions don't, lol)

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -80,6 +80,27 @@ auto get_sql_string(sql::SQLString const& str) -> std::string {
...
+        case spider::core::TaskLanguage::Python:
+            return "python";
+        case spider::core::TaskLanguage::Cpp:
+            return "cpp";
+        default:
+            return "";
+    }
+}
+
+auto string_to_task_language(std::string const& language) -> spider::core::TaskLanguage {
```
LinZhihao-723, line 94: Can we use `std::string_view`?

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -80,6 +80,27 @@ auto get_sql_string(sql::SQLString const& str) -> std::string {
...
+    if (language == "python") {
+        return spider::core::TaskLanguage::Python;
+    }
+    if (language == "cpp") {
+        return spider::core::TaskLanguage::Cpp;
+    }
+    return spider::core::TaskLanguage::Cpp;
+}
```
LinZhihao-723, line 102: * Is there a better way to reflect the error on unknown language types? I'm not familiar with the current Spider core error handling. Maybe throw an exception since I don't see it's easy to integrate `Result` into the current APIs.
* The current behavior is a bit confusing: the default behavior for string conversion is to return an empty string, whereas the string-to-type conversion returns Cpp by default.
  * If we can't report the error, how about using "unknown" and thus we know it's an unknown type set intentionally.

sitaowang1998: * There was talking about using the `Results` for a long time, but since it was not available at the time of initial development, it is not used at that time, and we don't have any time for the conversion.
* I am ok with an `unknown`. However, I am not 100% sure the convert to empty string by default is good. The function should immediately report an error instead.

How about we start to adopt `Result` by making these two function to return one to enforce the user of these functions to do explicit error check. We can devise a plan for better error handling for the whole C++ project later?

sitaowang1998: Unfortunately now right now, but I don't think it is hard to convert it.

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -1181,7 +1253,11 @@ auto MySqlMetadataStorage::add_child(
 ) -> StorageErr {
     try {
         sql::bytes const job_id = uuid_get_bytes(child.get_id());
-        add_task(static_cast<MySqlConnection&>(conn), job_id, child, std::nullopt);
+        auto res = add_task(static_cast<MySqlConnection&>(conn), job_id, child, std::nullopt);
+        if (res.has_error()) {
+            static_cast<MySqlConnection&>(conn)->rollback();
+            return StorageErr{res.error().get_error(), "Cannot add task"};
+        }
```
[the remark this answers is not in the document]
sitaowang1998: Remove this function because it is not used and tested.

src/spider/core/Error.hpp
```diff
@@ -31,4 +36,6 @@ struct StorageErr {
 };
 }  // namespace spider::core
 
+YSTDLIB_ERROR_HANDLING_MARK_AS_ERROR_CODE_ENUM(spider::core::StorageErrType);
+
```
LinZhihao-723, line 40: As discussed offline, I don't think it's needed if we don't return `std::error_code` as the error type in a `Result`.

src/spider/core/Error.cpp
LinZhihao-723, line 1: Same. In the current implementation, I don't think this is needed.
We only need to implement the category details if it's a `std::error_code`.

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -80,6 +84,29 @@ auto get_sql_string(sql::SQLString const& str) -> std::string {
 
 // NOLINTEND
 
+[[nodiscard]] auto task_language_to_string(spider::core::TaskLanguage language)
+        -> Result<std::string, StorageErrorCode> {
```
LinZhihao-723, line 88: It seems like u don't need `StorageErrorCode`. We can directly return `StorageErrType` here.

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -430,7 +467,16 @@ auto MySqlMetadataStorage::add_job(
...
+            auto const res = add_task(
+                    static_cast<MySqlConnection&>(conn),
+                    job_id_bytes,
+                    *task,
+                    TaskState::Ready
+            );
+            if (res.has_error()) {
+                static_cast<MySqlConnection&>(conn)->rollback();
+                return StorageErr{res.error().get_error(), "Cannot add task"};
+            }
```
LinZhihao-723, line 479:
```suggestion
            if (auto const result = add_task(
                        static_cast<MySqlConnection&>(conn),
                        job_id_bytes,
                        *task,
                        TaskState::Ready
                );
                result.has_error())
            {
                static_cast<MySqlConnection&>(conn)->rollback();
                return StorageErr{result.error().get_error(), "Cannot add task"};
            }
```
* We'd prefer to use `xxx_result` or `result` in its full name.
* We'd prefer to put the `result` inside the if.

LinZhihao-723 (changes requested):
Some comments in terms of using `Result`.

src/spider/storage/mysql/MySqlStorage.hpp
```diff
@@ -118,20 +118,21 @@ class MySqlMetadataStorage : public MetadataStorage {
...
 
-    static void add_task(
+    [[nodiscard]] static auto add_task(
             MySqlConnection& conn,
             sql::bytes job_id,
             Task const& task,
             std::optional<TaskState> const& state
-    );
-    static void add_task_batch(
+    ) -> boost::outcome_v2::std_checked<void, StorageErrType>;
```
LinZhihao-723, line 126:
```suggestion
    ) -> boost::outcome_v2::std_checked<void, StorageErrType>;

```

src/spider/storage/mysql/MySqlStorage.hpp
```diff
@@ -118,20 +118,21 @@ class MySqlMetadataStorage : public MetadataStorage {
...
+    [[nodiscard]] static auto add_task_batch(
             MySqlJobSubmissionBatch& batch,
             sql::bytes job_id,
             Task const& task,
             std::optional<TaskState> const& state
-    );
-    static auto fetch_full_task(MySqlConnection& conn, std::unique_ptr<sql::ResultSet> const& res)
-            -> Task;
+    ) -> boost::outcome_v2::std_checked<void, StorageErrType>;
```
LinZhihao-723, line 132:
```suggestion
    ) -> boost::outcome_v2::std_checked<void, StorageErrType>;

```

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -80,6 +82,29 @@ auto get_sql_string(sql::SQLString const& str) -> std::string {
...
 
+[[nodiscard]] auto task_language_to_string(spider::core::TaskLanguage language)
+        -> boost::outcome_v2::std_result<std::string, StorageErrType> {
+    switch (language) {
+        case spider::core::TaskLanguage::Python:
+            return "python";
+        case spider::core::TaskLanguage::Cpp:
+            return "cpp";
+        default:
+            return StorageErrType{StorageErrType::TaskLanguageErr};
```
LinZhihao-723, line 93:
```suggestion
            return StorageErrType::TaskLanguageErr;
```

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -80,6 +82,29 @@ auto get_sql_string(sql::SQLString const& str) -> std::string {
...
+
+[[nodiscard]] auto string_to_task_language(std::string_view const language)
+        -> boost::outcome_v2::std_result<TaskLanguage, StorageErrType> {
+    if (language == "python") {
+        return spider::core::TaskLanguage::Python;
+    }
+    if (language == "cpp") {
+        return spider::core::TaskLanguage::Cpp;
+    }
+    return StorageErrType{StorageErrType::TaskLanguageErr};
```
LinZhihao-723, line 105:
```suggestion
    return StorageErrType::TaskLanguageErr;
```

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -256,13 +281,16 @@ void MySqlMetadataStorage::add_task(
     task_statement->setBytes(1, &task_id_bytes);
     task_statement->setBytes(2, &job_id);
     task_statement->setString(3, task.get_function_name());
+    auto const task_language
+            = YSTDLIB_ERROR_HANDLING_TRYX(task_language_to_string(task.get_language()));
+    task_statement->setString(4, task_language);
```
LinZhihao-723, line 286:
```suggestion
    task_statement->setString(
            4,
            YSTDLIB_ERROR_HANDLING_TRYX(task_language_to_string(task.get_language()))
    );
```
Let's put a rvalue directly, just in case later mariadb API changes will introduce more efficient constructors (like move).

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -320,28 +348,33 @@ void MySqlMetadataStorage::add_task(
...
     // Add task
     sql::PreparedStatement& task_statement = batch.get_task_stmt();
     sql::bytes task_id_bytes = uuid_get_bytes(task.get_id());
     // NOLINTBEGIN(cppcoreguidelines-avoid-magic-numbers, readability-magic-numbers)
     task_statement.setBytes(1, &task_id_bytes);
     task_statement.setBytes(2, &job_id);
     task_statement.setString(3, task.get_function_name());
+    auto const task_language
+            = YSTDLIB_ERROR_HANDLING_TRYX(task_language_to_string(task.get_language()));
+    task_statement.setString(4, task_language);
```
LinZhihao-723, line 370:
```suggestion
    task_statement.setString(
            4,
            YSTDLIB_ERROR_HANDLING_TRYX(task_language_to_string(task.get_language()))
    );
```
Same.

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -557,12 +612,18 @@ auto MySqlMetadataStorage::add_job_batch(
...
             }
             Task const* task = task_option.value();
-            add_task_batch(
-                    static_cast<MySqlJobSubmissionBatch&>(batch),
-                    job_id_bytes,
-                    *task,
-                    TaskState::Ready
-            );
+
+            if (auto result = add_task_batch(
```
LinZhihao-723, line 616:
```suggestion
            if (auto const result = add_task_batch(
```

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -656,12 +723,16 @@ auto MySqlMetadataStorage::add_job_batch(
 // NOLINTEND(readability-function-cognitive-complexity)
 
 namespace {
-auto fetch_task(std::unique_ptr<sql::ResultSet> const& res) -> Task {
+auto fetch_task(std::unique_ptr<sql::ResultSet> const& res)
+        -> boost::outcome_v2::std_checked<Task, StorageErrType> {
     boost::uuids::uuid const id = read_id(res->getBinaryStream("id"));
     std::string const function_name = get_sql_string(res->getString("func_name"));
+    TaskLanguage const language = YSTDLIB_ERROR_HANDLING_TRYX(
```
LinZhihao-723, line 730:
```suggestion
    auto const language = YSTDLIB_ERROR_HANDLING_TRYX(
```

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -750,8 +821,8 @@ auto fetch_task_graph_task_output(TaskGraph* task_graph, std::unique_ptr<sql::Re
 auto MySqlMetadataStorage::fetch_full_task(
         MySqlConnection& conn,
         std::unique_ptr<sql::ResultSet> const& res
-) -> Task {
-    Task task = fetch_task(res);
+) -> boost::outcome_v2::std_checked<Task, StorageErrType> {
+    Task task = YSTDLIB_ERROR_HANDLING_TRYX(fetch_task(res));
```
LinZhihao-723, line 825:
```suggestion
    auto task = YSTDLIB_ERROR_HANDLING_TRYX(fetch_task(res));
```

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -1224,7 +1271,12 @@ auto MySqlMetadataStorage::get_task(StorageConnection& conn, boost::uuids::uuid
             };
         }
         res->next();
-        *task = fetch_full_task(static_cast<MySqlConnection&>(conn), res);
+        auto fetch_task_result = fetch_full_task(static_cast<MySqlConnection&>(conn), res);
```
LinZhihao-723, line 1274: Please double-check ur code before calling for the next round of review: the same code sections have some marked with `const`, but others are not.

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -804,7 +876,12 @@ auto MySqlMetadataStorage::get_task_graph(
         }
         while (task_res->next()) {
             // get_task_graph has special optimization to get inputs and outputs in batch
-            task_graph->add_task(fetch_task(task_res));
+            auto const fetch_task_result = fetch_task(task_res);
+            if (fetch_task_result.has_error()) {
+                static_cast<MySqlConnection&>(conn)->rollback();
+                return StorageErr{fetch_task_result.error(), "Cannot fetch task"};
+            }
+            task_graph->add_task(fetch_task_result.assume_value());
```
LinZhihao-723, line 884: What's the reason not to use `value()`?

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -1224,7 +1271,12 @@ auto MySqlMetadataStorage::get_task(StorageConnection& conn, boost::uuids::uuid
             };
         }
         res->next();
-        *task = fetch_full_task(static_cast<MySqlConnection&>(conn), res);
+        auto fetch_task_result = fetch_full_task(static_cast<MySqlConnection&>(conn), res);
+        if (fetch_task_result.has_error()) {
+            static_cast<MySqlConnection&>(conn)->rollback();
+            return StorageErr{fetch_task_result.error(), "Failed to fetch full task"};
+        }
+        *task = fetch_task_result.assume_value();
```
LinZhihao-723, line 1279:
```suggestion
        *task = std::move(fetch_task_result.value());
```
Use the move constructor.

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -1874,7 +1927,16 @@ auto MySqlMetadataStorage::get_child_tasks(
...
+            auto const fetch_task_result
+                    = fetch_full_task(static_cast<MySqlConnection&>(conn), res);
+            if (fetch_task_result.has_error()) {
+                static_cast<MySqlConnection&>(conn)->rollback();
+                return StorageErr{
+                        fetch_task_result.error(),
+                        fmt::format("Failed to fetch task {}", boost::uuids::to_string(id))
+                };
+            }
+            children->emplace_back(fetch_task_result.assume_value());
```
LinZhihao-723, line 1939:
```suggestion
            children->emplace_back(std::move(fetch_task_result.value()));
```
Same, use the move constructor and drop `const` in the result.

src/spider/storage/mysql/MySqlStorage.cpp
```diff
@@ -1901,7 +1964,15 @@ auto MySqlMetadataStorage::get_parent_tasks(
...
-            tasks->emplace_back(fetch_full_task(static_cast<MySqlConnection&>(conn), res));
+            auto fetch_task_result = fetch_full_task(static_cast<MySqlConnection&>(conn), res);
+            if (fetch_task_result.has_error()) {
+                static_cast<MySqlConnection&>(conn)->rollback();
+                return StorageErr{
+                        fetch_task_result.error(),
+                        fmt::format("Failed to fetch task {}", boost::uuids::to_string(id))
+                };
+            }
+            tasks->emplace_back(fetch_task_result.assume_value());
```
LinZhihao-723, line 1975:
```suggestion
            tasks->emplace_back(std::move(fetch_task_result.value()));
```

LinZhihao-723 (changes requested):
* Be careful about the use of `const`.
* Use `auto` whenever possible when receiving a return value from a function/macro.

PR #213, opened 2025-08-27
refactor: Remove the variadic template constructor from `TaskExecutor`.

sitaowang1998:
# Description

`TaskExecutor` has two custom constructors. One uses the variadic template to accept task arguments, the other one accept a vector of `msgpack::sbuffer` for arguments. The first one is only used inside unit tests, while the second one is not used in any unit tests, but actually used by worker. This means that unit tests do not test the code that is actually executed by a worker, and we have to maintain two copies of nearly the same code.

This PR removes the variadic template constructor, and uses the only constructor left in unit tests.

# Checklist

  breaking change.

# Validation performed

* [x] Unit tests and integration tests pass in Ubuntu 24.04.
* [x] GitHub workflows pass.

## Summary by CodeRabbit

- New Features
  - Standardized argument handling for task execution using packed buffers for more consistent behaviour.

- Refactor
  - Simplified TaskExecutor construction and public constructor signature while preserving observable runtime behaviour.

- Tests
  - Added a test helper to pack arguments and updated tests to use packed buffers for clearer, more robust coverage.

### DIFF
--- a/tests/worker/test-TaskExecutor.cpp
+++ b/tests/worker/test-TaskExecutor.cpp
@@ -65,10 +65,10 @@ auto get_libraries() -> std::vector<std::string> {
 }
 
 template <typename... Args>
-auto pack_args(Args&&... args) -> std::vector<msgpack::sbuffer> {
+[[nodiscard]] auto pack_args(Args&&... args) -> std::vector<msgpack::sbuffer> {
     std::vector<msgpack::sbuffer> packed_args;
     packed_args.reserve(sizeof...(args));
-    auto pack_args = [&packed_args](auto&& arg) {
+    auto pack_args = [&](auto&& arg) {
         msgpack::sbuffer buffer;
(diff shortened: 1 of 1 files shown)


tests/worker/test-TaskExecutor.cpp
```diff
@@ -64,6 +64,19 @@ auto get_libraries() -> std::vector<std::string> {
     return {lib_path.string()};
 }
 
+template <typename... Args>
+auto pack_args(Args&&... args) -> std::vector<msgpack::sbuffer> {
```
LinZhihao-723, line 68:
```suggestion
[[nodiscard]] auto pack_args(Args&&... args) -> std::vector<msgpack::sbuffer> {
```

LinZhihao-723 (changes requested):
Small fixes otherwise lgtm.

LinZhihao-723 (approved):
For the PR title, how about:
```
refactor: Remove the variadic template constructor from `TaskExecutor`.
```

PR #219, opened 2025-09-06
feat(tdl): Add `StructSpecDependencyGraph` to represent the dependencies among defined structs as a graph.

LinZhihao-723:
# Description

This PR adds a new class, `StructSpecDependencyGraph`, to represent the dependencies among defined structs as a graph. It is added under the namespace `tdl/pass/analysis` since this graph will be used in analysis passes. In future PRs, we will extend this class to:
- Compute the SCC within the dependency graph so that we can report circular dependencies efficiently.
- Compute the topological sort result of the graph so that the generated code can lay out code in a define-before-use manner.

# Checklist

  breaking change.

# Validation performed

* [x] Ensure all workflows passed.
* [x] Manually checked the generated graph matches the expected result. Didn't add unit tests for the graph generation. Will add unit tests when SCC computation is added.

## Summary by CodeRabbit

- **New Features**
  - Added a public dependency graph for struct specifications to analyse relationships between structures.
  - Graph can be constructed from a translation unit and exposes a method to get the total number of struct specifications.

- **Tests**
  - Added tests that build the dependency graph from parsed input and verify expected struct counts.

### DIFF
--- a/src/spider/tdl/pass/analysis/StructSpecDependencyGraph.cpp
+++ b/src/spider/tdl/pass/analysis/StructSpecDependencyGraph.cpp
@@ -2,12 +2,12 @@
 
 #include <cstddef>
 #include <memory>
-#include <set>
 #include <string>
 #include <tuple>
 #include <vector>
 
 #include <absl/container/flat_hash_map.h>
+#include <absl/container/flat_hash_set.h>
 #include <ystdlib/error_handling/Result.hpp>
 
 #include <spider/tdl/parser/ast/nodes.hpp>
@@ -28,8 +28,8 @@ StructSpecDependencyGraph::StructSpecDependencyGraph(
 
     // Build def-use chains
     m_def_use_chains.reserve(num_struct_specs);
-    for (auto const& [_, struct_spec] : struct_specs) {
-        std::set<size_t> use_ids;
+    for (auto const& def : m_struct_spec_refs) {
+        absl::flat_hash_set<size_t> use_ids;
         auto field_visitor
                 = [&](parser::ast::NamedVar const& field) -> ystdlib::error_handling::Result<void> {
             auto const* type_as_struct{dynamic_cast<parser::ast::Struct const*>(field.get_type())};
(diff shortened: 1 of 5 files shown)


src/spider/tdl/pass/analysis/StructSpecDependencyGraph.cpp
```diff
@@ -0,0 +1,59 @@
...
+            if (struct_specs.cend() == it) {
+                // This is a dangling reference, which will be caught in other analysis pass. In
+                // this dependency graph, we just ignore it.
+                return ystdlib::error_handling::success();
+            }
+
+            auto const use_id{m_struct_spec_ids.at(it->second.get())};
+            use_ids.emplace(use_id);
+            return ystdlib::error_handling::success();
+        };
```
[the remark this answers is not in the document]
LinZhihao-723: Yeah, that's a good point. I will add this fix in the next commit.

src/spider/tdl/pass/analysis/StructSpecDependencyGraph.hpp
```diff
@@ -0,0 +1,52 @@
...
+    // Destructor
+    ~StructSpecDependencyGraph() = default;
+
+    // Methods
+    [[nodiscard]] auto get_num_struct_specs() const -> size_t { return m_struct_spec_refs.size(); }
+
+private:
+    // Variables
+    std::vector<std::shared_ptr<StructSpec const>> m_struct_spec_refs;
+    absl::flat_hash_map<StructSpec const*, size_t> m_struct_spec_ids;
```
sitaowang1998, line 47: I prefer to use `indices` instead of `ids`, but I am ok with `ids`. If you change to use `indices`, please also change the `ids` in the cpp file.

LinZhihao-723: We'd actually use/prefer `id` and `ids` in CLP core. It's more concise (shorter to read).
But I'm personally fine with `index` and `indices`, which is why I didn't point it out in the Python client PRs.

sitaowang1998 (commented):
LGTM. Could you expand the unit tests to include something like using struct as a list element type or map value type?

LinZhihao-723:
> LGTM. Could you expand the unit tests to include something like using struct as a list element type or map value type?

The current unit test doesn't actually check the constructed dependency graph unfortunately...
I will make a comprehensive unit test in the next PR with the SCC detection.

sitaowang1998 (approved):

PR #229, opened 2025-09-15
feat(spider-py): Add python task executor.

sitaowang1998:
# Description

This PR adds task executor in `python` that directly executes `python` tasks. The `python` executor has the same command line interface as the `c++` task executor.

# Checklist

  breaking change.

# Validation performed

* [ ] GitHub workflows pass.


python/spider-py/src/spider_py/task_executor/task_executor.py
```diff
@@ -1,8 +1,171 @@
...
+        for result in results:
+            if isinstance(result, client.Data):
+                response_messages.append(result.id.bytes)
+            else:
+                response_messages.append(to_serializable(result))
+    elif isinstance(results, client.Data):
+        response_messages.append(results.id.bytes)
+    else:
+        response_messages.append(to_serializable(results))
+    return response_messages
```
LinZhihao-723, line 104: I suggest to add the following helper:
```python
def parse_single_output_to_serializable(output: object) -> object:
    """
    Parses a single output from the function execution to a serializable form.
    :param output: Output to parse.
    :return: The parsed output.
    """
    if isinstance(output, client.Data):
        return output.id.bytes
    return to_serializable(output)
```
Then rewriting in this way:
```python
    response_messages: list[object] = [TaskExecutorResponseType.Result]
    if not isinstance(results, tuple):
        response_messages.append(parse_single_output_to_serializable(results))
        return response_messages
    # Parse as a tuple
    for result in results:
        response_messages.append(parse_single_output_to_serializable(result))
    return response_messages
```

python/spider-py/src/spider_py/task_executor/task_executor.py
```diff
@@ -1,8 +1,171 @@
...
+
+    with fdopen(input_pipe_fd, "rb") as input_pipe, fdopen(output_pipe_fd, "wb") as output_pipe:
+        input_message = receive_message(input_pipe)
+        arguments = get_request_body(input_message)
+        logger.debug("Args buffer parsed")
+
+        # Get the function to run
+        function_name = func.split(".")[-1] if "." in func else func
+        function = None
+        for lib in libs:
```
LinZhihao-723, line 136: - Seems like using `mods` is more appropriate?
- Why do we allow users to input a list of modules?
- Seems like the function name is a full name from the library to modules to the attribute. In this case, why we just check the function name but not the full path?

python/spider-py/src/spider_py/task_executor/task_executor.py
```diff
@@ -1,8 +1,171 @@
...
+                break
+        if function is None:
+            msg = f"Function {function_name} not found in provided libraries."
+            raise ValueError(msg)
+
+        signature = inspect.signature(function)
+        if len(signature.parameters) != len(arguments) + 1:
+            msg = (
+                f"Function {function_name} expects {len(signature.parameters) - 1} "
+                f"arguments, but {len(arguments)} were provided."
```
LinZhihao-723, line 149:
```suggestion
                f"Function {function_name} expects {len(signature.parameters) - 1} arguments, but"
                f" {len(arguments)} were provided."
```
- We wrap when the line limit reaches 100.
- We usually add space on the second line.

python/spider-py/src/spider_py/task_executor/task_executor.py
```diff
@@ -1,8 +1,171 @@
...
+    else:
+        response_messages.append(to_serializable(results))
+    return response_messages
+
 
 def main() -> None:
-    """Main function to execute the task."""
+    """
+    Main function to execute the task.
+    :raises StorageError: If a storage operation fails.
```
LinZhihao-723, line 110: - This raise seems to be outdated. Please update to reflect ur latest implementation.
- I don't fully understand what's the error reporting model for our task executor. I guess we just let the Python runtime to print the stack trace on exception?

python/spider-py/src/spider_py/task_executor/task_executor.py
```diff
@@ -1,8 +1,171 @@
...
+            msg = f"Function {function_name} not found in provided libraries."
+            raise ValueError(msg)
+
+        signature = inspect.signature(function)
+        if len(signature.parameters) != len(arguments) + 1:
+            msg = (
+                f"Function {function_name} expects {len(signature.parameters) - 1} "
+                f"arguments, but {len(arguments)} were provided."
+            )
+            raise ValueError(msg)
```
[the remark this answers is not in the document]
sitaowang1998: It is easier to print full error messages outside of the function. I'll update the docstring of `parse_arguments`.

python/spider-py/src/spider_py/task_executor/task_executor.py
sitaowang1998: 1. The command line has to use `libs` to keep the interface same with C++ side. We can use `mods` for variable names.
2. Because we don't check which modules contains the function we need in worker.
3. We requires the function name to be unique inside a Spider deployment, so just checking the function name should be enough.

sitaowang1998: The full name here is the `__qualname__`, which does not contain the module name.

python/spider-py/src/spider_py/task_executor/task_executor.py
```diff
@@ -1,8 +1,184 @@
...
+
+
+def parse_task_arguments(
+    storage: Storage, params: list[inspect.Parameter], arguments: list[object]
+) -> list[object]:
+    """
+    Parses arguments for the function to be executed.
+    :param storage: Storage instance to use to get Data.
+    :param params: List of parameters in the function signature.
+    :param arguments: List of arguments to parse. Must be the same length as params.
```
LinZhihao-723, line 71: This doesn't imply the size check is enforced outside, and it will make people think it may handle errors if the length is different.
The right way to document this is to explicitly add a `NOTE` to say the caller is responsible for the size checking.

python/spider-py/src/spider_py/task_executor/task_executor.py
LinZhihao-723: 1 is not true in Python
We can specify the full path of the function, from the module name to the function name. And Python should be able to load it from the root-level module name. It doesn't need to know the actual filesystem path, as long as the library is locatable by the Python runtime.

sitaowang1998: 1 does not hold in Python venv. However, the worker needs to take care of all languages, so we cannot filter out the exact library in worker.

sitaowang1998: * Python can find the library by just giving the module in a venv.
* In the case of C++ task, we don't have the good thing like venv, so we have to use the full path to locate the library, which could even be different among different nodes. Thus, the worker, when invoking C++ executor, have to provide a list of libraries and let the C++ task executor to figure out which one to use. If we want to support getting the exact library, we need to change the spider's storage to add a table that tracks the libraries and their function names, which is only used by Python tasks but not C++ tasks. The worker also needs to change accordingly to support this only for Python. I don't think this is a good idea, especially when we might add other languages as well. The current solution we have is general enough for all languages.

LinZhihao-723: * Is it a problem that the worker can't create a virtual env for Python executor?
* I don't understand. If you give a full function path including the module, it is the executor's job to load the module, and the executor knows itself is a Python executor. You don't have to change anything for any other language.

python/spider-py/src/spider_py/task_executor/task_executor.py
```diff
@@ -1,8 +1,182 @@
...
+    return body
+
+
+def parse_task_arguments(
+    storage: Storage, params: list[inspect.Parameter], arguments: list[object]
+) -> list[object]:
+    """
+    Parses arguments for the function to be executed.
+    Note that `params` does not include the `TaskContext` parameter, and must be the same length as
+    `arguments`.
```
LinZhihao-723, line 70:
```suggestion

    NOTE: `params` does not include the `TaskContext` parameter, and must be the same length as
    `arguments`. The caller is responsible for the size check.

```
As indicated in the previous comment: https://github.com/y-scope/spider/pull/229#discussion_r2352580969
* `NOTE` is the notation we use for special alerts.
* As "explicitly add a NOTE to say the caller is responsible for the size checking", I don't think this comment makes a difference if it doesn't explicitly point out that the caller is responsible for the size checking.

LinZhihao-723 (changes requested):

PR #242, opened 2025-10-17
feat(spider-py): Add support for deserializing output type names from `bytes`.

sitaowang1998:
# Description

## Background
`TaskOutput.type` sometime could contains `bytes` instead of `str`, if user's metadata storage uses `VARBINARY` or `BLOB` instead of `VARCHAR` for the type column. This value will be feed into lark parsing, which only accepts `str` input, and thus crash.

## Solution
This PR adds support for `bytes` by checking the type of the `TaskOutput.type` and decodes `bytes` using `utf-8`.

# Checklist

  breaking change.

# Validation performed

* [x] Python integration tests pass.
* [x] Using `BLOB` for type column in MariaDB works.
* [x] GitHub workflows pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deserialization to correctly handle type information provided as raw bytes by decoding it to UTF-8 before parsing, preventing type-related errors and ensuring consistent processing of outputs regardless of format.

### DIFF
--- a/python/spider-py/src/spider_py/client/job.py
+++ b/python/spider-py/src/spider_py/client/job.py
@@ -53,10 +53,10 @@ def _deserialize_outputs(outputs: list[core.TaskOutput]) -> tuple[object, ...] |
     results = []
     for output in outputs:
         if isinstance(output.value, core.TaskOutputValue):
-            output_type = output.type
-            if isinstance(output_type, bytes):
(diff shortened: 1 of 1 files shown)


LinZhihao-723 (approved):
For the PR title, how about:
```
feat(spider-py): Add support for deserializing output type names from `bytes`.
```

PR #243, opened 2025-10-17
feat(logging): Add support for configuring loggers via environment variables.

sitaowang1998:
# Description

This PR:
* Adds logging to `${SPIDER_LOG_FILE}` if the environment is set.
* Adds logging to `${SPIDER_LOG_DIR}/worker_<worker_id>.log` if the environment is set.
* Adds logging to `${SPIDER_LOG_DIR}/task_<task_id>.log` if the environment is set.

If the environment is not set, each component still logs to `stdout`.

# Checklist

  breaking change.

# Validation performed

* [x] Log files created and contain logs when the corresponding environment is set for each component.
* [x] Integration tests pass with both environment set and unset.
* [ ] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Environment-driven logging: write to a file or per-instance log directory when configured; otherwise fall back to console.
  * Automatic UUID-based log files for tasks and workers to separate per-run logs.

* **Refactor**
  * Centralized logging utility added and logging initialization moved earlier for consistent diagnostics.

* **Chores**
  * Shared logging utility wired into builds for scheduler, worker and task executor.

### DIFF
--- a/src/spider/utils/logging.cpp
+++ b/src/spider/utils/logging.cpp
@@ -70,7 +70,7 @@ auto setup_directory_logger(
 ) -> void {
     // NOLINTNEXTLINE(concurrency-mt-unsafe)
     char const* const log_file_dir = std::getenv("SPIDER_LOG_DIR");
-    if (nullptr == log_file_dir) {
+    if (nullptr != log_file_dir) {
         auto const log_file_path = std::filesystem::path{log_file_dir}
                                    / fmt::format("{}_{}.log", logger_name, to_string(id));
 


src/spider/worker/task_executor.cpp
```diff
@@ -70,6 +75,46 @@ auto parse_arg(int const argc, char** const& argv) -> boost::program_options::va
...
+    try {
+        auto const file_logger
+                = spdlog::basic_logger_mt("spider_task_executor", log_file_path.string());
+        spdlog::set_default_logger(file_logger);
+        spdlog::flush_on(spdlog::level::info);
+    } catch (spdlog::spdlog_ex& ex) {
+        auto const console_logger = spdlog::stdout_color_mt("spider_task_executor_console");
+        spdlog::set_default_logger(console_logger);
+        spdlog::flush_on(spdlog::level::info);
+    }
```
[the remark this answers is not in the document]
sitaowang1998: PR description has been updated.

src/spider/scheduler/scheduler.cpp
```diff
@@ -145,17 +148,46 @@ auto cleanup_loop(
...
 }
 
-constexpr int cSignalExitBase = 128;
-}  // namespace
+/**
+ * Sets up the logger for the scheduler.
+ * Writes logs to `SPIDER_LOG_FILE` if the environment variable is set.
+ * Writes logs to console otherwise.
+ */
+auto setup_logger() -> void {
```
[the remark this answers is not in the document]
LinZhihao-723: How about `logging`?

src/spider/utils/logging.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include <boost/uuid/uuid.hpp>
+
+namespace spider::logging {
+/**
+ * Sets up the logger to write to `$SPIDER_LOG_FILE` if the environment variable is set.
+ * Writes logs to console otherwise.
+ *
+ * @param logger_name The name of the logger.
+ * @param source_name The source name to be included in the log line.
```
LinZhihao-723: `source` is a bit confusing here. How about using `tag` or `group`?

src/spider/utils/logging.cpp
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Sets up a file logger with a fallback to console logger if file creation fails.
+ * @param logger_name The name of the logger.
+ * @param log_file_path The path of the log file.
+ */
+auto setup_logger(std::string const& logger_name, std::string const& log_file_path) -> void {
+    try {
+        auto const file_logger = spdlog::basic_logger_mt(logger_name, log_file_path);
+        spdlog::set_default_logger(file_logger);
+        spdlog::flush_on(spdlog::level::info);
```
LinZhihao-723: Do we need to flush on every info-level log? Can we set it to warning-level?

src/spider/utils/logging.cpp
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Sets up the log format and level for the spdlog logger.
+ * @param source_name The source name to include in the log format.
+ */
+auto setup_log_format(std::string const& source_name) -> void {
+    // NOLINTNEXTLINE(misc-include-cleaner)
+    spdlog::set_pattern(fmt::format("[%Y-%m-%d %H:%M:%S.%e] [%^%l%$] [{}] %v", source_name));
+#ifndef NDEBUG
+    spdlog::set_level(spdlog::level::trace);
+#endif
```
LinZhihao-723: Why not moving this to `setup_logger`? It's not a part of the format setting, right?

LinZhihao-723 (changes requested):
Let's finalize the logger APIs first.

src/spider/utils/logging.cpp
sitaowang1998: This is because `setup_logger` is not called if the env is not set, but the level needs to be set in all cases.

src/spider/utils/logging.hpp
```diff
@@ -0,0 +1,34 @@
...
+ */
+auto setup_file_logger(std::string const& logger_name, std::string const& source_name) -> void;
```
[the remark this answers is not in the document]
LinZhihao-723: You can explicitly cast it to `std::string{xxx}` when passing to `spdlog::basic_logger_mt`.

src/spider/utils/logging.hpp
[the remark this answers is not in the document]
LinZhihao-723: Then it should be under `utils` namespace.

src/spider/utils/logging.cpp
sitaowang1998: Also just FYI, I am using info because it is the default level.

src/spider/utils/logging.cpp
```diff
@@ -0,0 +1,88 @@
...
+    char const* const log_file_dir = std::getenv("SPIDER_LOG_DIR");
+    if (nullptr == log_file_dir) {
+        set_log_config(tag);
+        return;
+    }
+
+    auto const log_file_path = std::filesystem::path{log_file_dir}
+                               / fmt::format("{}_{}.log", logger_name, to_string(id));
+
+    set_default_logger(logger_name, log_file_path.string());
```
LinZhihao-723: Similarly, can we move `set_default_logger` inside if?

PR #249, opened 2025-10-31
fix(ci):  Lock `task` to v3.44.0 in GH workflows to avoid go-task/task#2440.

sitaowang1998:
# Description

https://github.com/go-task/task/issues/2440 describes an issue where even though a task was failing, versions of `task` installed through `npm` would return with a 0 exit. This issue doesn't show up in `task` 3.44.0 which is also the version of `task` that clp requires.

This PR updates all GH workflows to use task 3.44.0.

# Checklist

  breaking change.

# Validation performed

* [x] GitHub workflow fails if a test fails.
* [x] All GitHub workflows pass.

## Summary by CodeRabbit

* **Chores**
  * Build and test workflows now install a specific CLI version (3.44.0) to ensure consistent, reproducible runs.

* **Documentation**
  * System requirements updated to require CLI 3.44.0; note added that this version is pinned and a reference to the related issue was included.

### DIFF
--- a/python/spider-py/src/spider_py/client/job.py
+++ b/python/spider-py/src/spider_py/client/job.py
@@ -30,7 +30,7 @@ class Job:
             fetch_and_update_job_status(self._storage, self._impl)
         return self._impl.status
 
-    def get_results(self) -> object | None:
+    def get_results(self) -> tuple[object, ...] | None:
         """
         :return: The job results if the job ended successfully.
         :return: None if the job is still running or ended unsuccessfully.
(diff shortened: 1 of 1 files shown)


LinZhihao-723 (changes requested):
Can we update the TDL workflow as well?
https://github.com/y-scope/spider/blob/41d0c677ecdf320ac813a2d98181f4cd92352a2d/.github/workflows/tdl-generated-code-checks.yaml#L33

LinZhihao-723 (approved):
Directly updated PR title to mirror https://github.com/y-scope/clp/pull/1350

PR #258, opened 2025-11-12
build(spider-py)!: Bump minimum requried Python version to 3.10; Bump minimum required `mariadb` version to 1.1.14; Bump minimum required `pytest` version to 9.0.0.

sitaowang1998:
# Description

#231 sets minimum required Python version to 3.9. Python 3.9 reached end of life on October 30, 2025. Thus, this PR bumps the minimum required Python version to 3.10. This PR also introduces following new features added in Python 3.10 into the codebase:
* Uses `X | Y` in type annotation instead of `Union[X, Y]`.
* Adds `strict=True` for all calls to `zip`.
However, the eager size checks before `zip` introduced in #231 are not removed for better error messages.

This PR also lift the lock on `mariadb` library version in #231. Now we use `mariadb>=1.1.14`.

This PR also updates `pytest` version to `>=9.0.0` to adopt the latest change in pytest configuration in `pyproject.toml`. The change is included in this PR because `pytest>=9.0.0` requires minimum Python version of `3.10`. We also turn off strict mode to `pyproject.toml` because `tombi` has a bug that is exposed by latest `partial-pytest.json` schema. The issue is tracked by #259.

# Checklist

  breaking change.

# Validation performed

* [x] Unit tests pass.
* [x] Integration tests pass.
* [x] GitHub workflows pass.

## Summary by CodeRabbit

* **Chores**
  * Minimum Python requirement raised to 3.10 and README updated.
  * Dependency constraints updated (MariaDB connector, pytest) and tooling config adjusted.
  * Modernized type annotations and cleaned up module imports.

* **Bug Fixes / Reliability**
  * Stricter iteration checks added to surface mismatches earlier and improve robustness.

### DIFF
--- a/python/spider-py/src/spider_py/storage/mariadb_storage.py
+++ b/python/spider-py/src/spider_py/storage/mariadb_storage.py
@@ -639,7 +639,7 @@ class MariaDBStorage(Storage):
         if status_str not in _StrToJobStatusMap:
             msg = f"Unknown job status: {status_str}."
             raise StorageError(msg)
-        # Uses fetchall after a fetchone to drain the result set even if it is already empty.
+        # Use fetchall after a fetchone to drain the result set even if it is already empty.
         cursor.fetchall()
         return _StrToJobStatusMap[status_str]
 


python/spider-py/src/spider_py/storage/mariadb_storage.py
```diff
@@ -409,7 +403,7 @@ def _gen_task_insertion_params(
         if len(jobs) != len(task_graphs):
             msg = "The lengths of `jobs` and `task_graphs` must match."
             raise ValueError(msg)
```
LinZhihao-723: Since we're using `strict`, do we still need this check?
There are some other instances. Check diff for https://github.com/y-scope/spider/pull/231/files

sitaowang1998: In the PR description:
However, the eager size checks before zip introduced in #231 are not removed for better error messages.

LinZhihao-723 (changes requested):
Do we still need this?
https://github.com/y-scope/spider/blob/cfbc4b6a3eb6d2b88e848b6671171e892f9bedb6/python/spider-py/src/spider_py/storage/mariadb_storage.py#L648

LinZhihao-723 (dismissed):
For the PR title, how about:
```
build(spider-py)!: Bump minimum requried Python version to 3.10; Bump minimum required `mariadb` version to 1.1.14; Bump minimum required `pytest` version to 9.0.0.
```

PR #285, opened 2026-03-14
feat(spider-storage): Define MariaDB schema for Spider Huntsman:

sitaowang1998:
* Add derives to map Rust enums to MySQL enum types.
* Introduce a DB connector that implements DB storage protocols and initializes tables on creation.
* Update MariaDB testing task wrappers.

# Description

This PR:
- Adds MariaDB table initialization in `rust` code.
- Updated tests to start the database.

This PR also does minor improvement:
- Cleans up python lint task.
- Add a `spider-derive` for a new procedure macro to generate `enum` strings at compile time.

# Checklist

  breaking change.

# Validation performed

- [x] Table initialization using both script and `initialize` works.
- [x] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * MariaDB-backed storage with automatic table creation.
  * Database initialization script for MariaDB.

* **New Features / Developer Utilities**
  * New derive macro producing a quoted, comma-separated enum string.
  * Added a local helper crate to the workspace for the macro.

* **Tests**
  * Unit tests validating the enum-string macro.

* **Chores**
  * Updated Rust dependencies and simplified Python lint workflows.


components/spider-derive/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+use proc_macro::TokenStream;
+use syn::{DeriveInput, parse_macro_input};
+
+#[proc_macro_derive(QuotedEnumStr)]
+pub fn derive_quoted_enum_str(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    enum_str::derive_quoted_enum_str(&input)
+        .unwrap_or_else(syn::Error::into_compile_error)
+        .into()
+}
```
LinZhihao-723: I think we need a docstring with some explanations for it. Let me draft one and also figure out how.
Before that, do you think we should name it to `AsMySqlEnumDecl`, since the purpose of this enum is for MySQL enum delcaration?

components/spider-derive/tests/quoted_enum_str.rs
```diff
@@ -0,0 +1,25 @@
...
+#[allow(dead_code)]
+#[derive(QuotedEnumStr)]
+enum Single {
+    Only,
+}
+
+#[test]
+fn single_variant() {
+    assert_eq!(Single::quoted_enum_str(), "'Only'");
+}
```
LinZhihao-723:
```suggestion

#[allow(dead_code)]
#[derive(QuotedEnumStr)]
enum Single {
    Only(i32),
}

#[test]
fn quoted_enum_str() {
    assert_eq!(Color::quoted_enum_str(), "'Red', 'Green', 'Blue'");
    assert_eq!(Single::quoted_enum_str(), "'Only'");
}
```
- We should group all the types
- We should probably try to make sure it still works if the variant derives another type.
- How about grouping all test cases into one for better management, and using the derived name for the test case?

components/spider-derive/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+[lib]
+name = "spider_derive"
+path = "src/lib.rs"
+proc-macro = true
+
+[dependencies]
+proc-macro2 = "1.0"
+quote = "1.0"
+syn = "2.0"
```
LinZhihao-723: I actually missed this in the previous review: we should lock to the patch version to make sure the patch upgrade doesn't break things for some random reason... (that's what happened with nextest)
We should update `serde`'s version in spider-storage.

LinZhihao-723 (changes requested):
Reviewed `spider-derives`. Overall this is good.
Since I'm asking about renaming, and I found it's probably too early to determine whether we really need a mod to group the derives based on enum or based on mysql, shall we just move everything inside `enum_str` into `lib` for now?

components/spider-derive/src/lib.rs
sitaowang1998: I am ok with the rename since I can't think of any other use of this function for now.

sitaowang1998:
> shall we just move everything inside `enum_str` into `lib` for now?

I prefer to make it in a separate file. `rust` requires all `proc_macro` declaration to live in `lib.rs`, so I prefer to leave the file just for declaration. The real derive should be kept in different files. I am ok if you want to rename the file.

LinZhihao-723:
> > shall we just move everything inside `enum_str` into `lib` for now?
> 
> I prefer to make it in a separate file. `rust` requires all `proc_macro` declaration to live in `lib.rs`, so I prefer to leave the file just for declaration. The real derive should be kept in different files. I am ok if you want to rename the file.

How about `mysql` then?

components/spider-storage/src/db/mariadb.rs
```diff
@@ -0,0 +1,207 @@
...
+}
+
+impl MariaDbStorage {
+    #[must_use]
+    pub const fn new(pool: MySqlPool) -> Self {
+        Self { pool }
+    }
+}
+
+impl MariaDbStorage {
```
LinZhihao-723: Should be in one `impl` group?

LinZhihao-723 (changes requested):
Still have some comments on the DB table schema. Will post my comments in a later session.

LinZhihao-723 (changes requested):
A few comments on the DB schema. In general lgtm.

LinZhihao-723 (changes requested):

components/spider-storage/src/db/mariadb.rs
```diff
@@ -0,0 +1,232 @@
...
+    ///
+    /// # Returns
+    ///
+    /// A new instance of `MariaDbStorageConnector` if connection and initialization succeed.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if
+    ///
+    /// * Forwards a [`sqlx::error::Error`] if database operation fails.
```
```suggestion
    /// Connects to database and initializes tables.
    ///
    /// # Returns
    ///
    /// A newly created [`MariaDbStorageConnector`] instance for connection on success.
    ///
    /// # Errors
    ///
    /// Returns an error if:
    ///
    /// * Forwards [`sqlx::mysql::MySqlPoolOptions::connect`]'s return values on failure.
    /// * Forwards [`Self::initialize`]'s return values on failure.
```
Please check what I've fixed.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -0,0 +1,232 @@
...
+    pub async fn connect_and_initialize(config: &DatabaseConfig) -> Result<Self, DbError> {
```
LinZhihao-723: How about simply `connect`?

components/spider-storage/src/db/mariadb.rs
```diff
@@ -0,0 +1,232 @@
...
+    ///
+    /// `MariaDB` does not support transactions for DDL statements. All DDL statements are
+    /// automatically committed. Thus, this function executes each table creation query separately,
+    /// and does not provide atomicity guarantees.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards a [`sqlx::query::Query::execute`]'s return values on failure.
```
```suggestion
    /// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
```

components/spider-storage/src/db/mariadb.rs
```diff
@@ -0,0 +1,232 @@
...
+        io::{TaskInput, TaskOutput},
+    },
+};
+use sqlx::MySqlPool;
+
+use crate::{
+    config::DatabaseConfig,
+    db::{DbError, ExternalJobOrchestration, InternalJobOrchestration, ResourceGroupManagement},
+};
+
```
```suggestion
/// A cloneable storage connector for `MariaDB` database that implements Spider's DB protocols.
```

components/spider-storage/src/config.rs
```diff
@@ -0,0 +1,13 @@
+use secrecy::SecretString;
+use serde::{Deserialize, Serialize};
+
```
LinZhihao-723, line 3:
```suggestion

/// Configuration parameters for connecting to the database.
```

components/spider-storage/src/db/mariadb.rs
```diff
@@ -0,0 +1,232 @@
...
+            .execute(&self.pool)
+            .await?;
+
+        sqlx::query(jobs_creation_query())
+            .execute(&self.pool)
+            .await?;
+
+        Ok(())
+    }
+}
```
LinZhihao-723, line 85:
```suggestion
}

```

components/spider-derive/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+mod mysql;
+
+use proc_macro::TokenStream;
+use syn::{DeriveInput, parse_macro_input};
+
```
LinZhihao-723, line 5:
```suggestion

/// Derives MySQL-compatible enum support for use with [`sqlx`].
///
/// This derive macro generates the necessary trait implementations and helper methods to store a
/// Rust enum as a MySQL `ENUM` column. It uses each variant's identifier name as the corresponding
/// MySQL enum string value (e.g., variant `Ready` maps to the string `'Ready'`).
///
/// # Generated methods
///
/// The macro adds the following `const` methods to the annotated enum:
///
/// ## `as_mysql_enum_decl() -> &'static str`
///
/// Returns the MySQL `ENUM(...)` type declaration string for use in DDL statements. For example, an
/// enum with variants `Red`, `Green`, `Blue` produces `"ENUM('Red','Green','Blue')"`.
///
/// ## `as_str(&self) -> &'static str`
///
/// Returns the unquoted variant name (e.g., `"Red"`).
///
/// ## `as_quoted_str(&self) -> &'static str`
///
/// Returns the single-quoted variant name for use in SQL literals (e.g., `"'Red'"`).
///
/// # Generated trait implementations
///
/// * [`sqlx::Type<sqlx::MySql>`] — Reports the MySQL type info as `str`, so that `sqlx` treats
///   the enum value as a text-compatible type.
/// * [`sqlx::Encode<sqlx::MySql>`] — Encodes the enum by delegating to `&str` encoding using the
///   variant name.
/// * [`sqlx::Decode<sqlx::MySql>`] — Decodes the enum by matching the database string against
///   known variant names. Returns an error for unknown values.
///
/// # Limitations
///
/// * Can only be derived for enums — applying it to a struct or union produces a compile-time
///   error.
/// * All variants must be unit variants (no tuple or struct fields).
/// * Variant names are used as-is for the MySQL values — there is no support for renaming via
///   attributes (e.g., `#[sql(rename = "...")]`).
/// * The generated [`sqlx::Decode`] implementation returns a runtime error (not a compile-time
///   error) if the database contains a value that does not match any variant name.
///
/// # Example
///
/// ```ignore
/// use spider_derive::MySqlEnum;
///
/// #[derive(MySqlEnum)]
/// enum JobState {
///     Ready,
///     Running,
///     Succeeded,
/// }
///
/// // DDL helper
/// assert_eq!(JobState::as_mysql_enum_decl(), "ENUM('Ready','Running','Succeeded')");
///
/// // String conversion
/// assert_eq!(JobState::Running.as_str(), "Running");
/// assert_eq!(JobState::Ready.as_quoted_str(), "'Ready'");
/// ```
```

components/spider-derive/src/mysql.rs
```diff
@@ -0,0 +1,98 @@
...
+use quote::quote;
+use syn::{Data, DataEnum, DeriveInput};
+
+pub fn derive_mysql_enum(input: &DeriveInput) -> syn::Result<TokenStream> {
+    let enum_type_name = &input.ident;
+
+    let Data::Enum(DataEnum { variants, .. }) = &input.data else {
+        return Err(syn::Error::new_spanned(
+            &input.ident,
+            "`QuotedEnumStr` can only be derived for enums",
```
LinZhihao-723:
```suggestion
            "`MySqlEnum` can only be derived for enums",
```

components/spider-derive/tests/test_mysql_enum.rs
LinZhihao-723, line 1: Rename to `test_mysql_enum.rs`?

tools/scripts/mariadb/huntsman/init_db.py
LinZhihao-723, line 1: We probably don't need this anymore.
Suppose we need to test DB operations, the creation of the connector will do this job for all table init. I assume all of our testing flows will have the DB connector created, right?

components/spider-core/src/job.rs
```diff
@@ -1,10 +1,12 @@
-use strum_macros::Display;
+use spider_derive::MySqlEnum;
 
 /// Represents a job in the Spider scheduling framework.
 pub struct Job {}
 
 /// Enum for all possible states of a job.
-#[derive(Debug, Display, Clone, Copy, PartialEq, Eq)]
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, strum_macros::Display, strum_macros::EnumIter, MySqlEnum,
```
LinZhihao-723: As discussed yesterday offline, do we still need `strum_macros::Display, strum_macros::EnumIter`?

LinZhihao-723 (changes requested):

components/spider-core/src/job.rs
sitaowang1998: Because we need to print out enum in error.

components/spider-storage/src/db/mariadb.rs
sitaowang1998: Why not include the `parameters` section?

tools/scripts/mariadb/huntsman/init_db.py
sitaowang1998: Hard to say. We haven't determined how to write the integration test. Removing this needs more changes in an already long PR. Until it becomes clear that this is not necessary, how about just keep this here for now?

components/spider-core/src/job.rs
LinZhihao-723: For `strum_macros::EnumIter`?

tools/scripts/mariadb/huntsman/init_db.py
LinZhihao-723: Do you have a PR draft already?

sitaowang1998: https://github.com/sitaowang1998/spider/tree/mariadb-test. I have tested that removing the script still works.

components/spider-core/src/job.rs
sitaowang1998: For `strum_macros::Display`.

tools/scripts/mariadb/huntsman/init_db.py
LinZhihao-723: I guess it's better to remove it then? I think it's preferred to have the DB schema defined in a single source. If we really need some sort of cli to create these tables, we should create a Rust bin to do it which reuses the Rust code introduced in this PR.

taskfiles/test.yaml
```diff
@@ -74,11 +74,11 @@ tasks:
           CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
           INCLUDE_PATTERNS: ["{{.OUTPUT_DIR}}"]
 
-  # A generic wrapper that runs the given task with a MariaDB storage backend.
+  # A generic wrapper that runs the given task with a wolf MariaDB storage backend.
```
LinZhihao-723:
```suggestion
  # A generic wrapper that runs the given task with a MariaDB storage backend for Spider Wolf.
```

LinZhihao-723 (changes requested):
Otherwise lgtm. Will come up with a multi-line commit message with the approval later.

LinZhihao-723 (approved):
For the PR title and the commit message, how about:
```
feat(spider-storage): Define MariaDB schema for Spider Huntsman:
* Add derives to map Rust enums to MySQL enum types.
* Introduce a DB connector that implements DB storage protocols and initializes tables on creation.
* Update MariaDB testing task wrappers.
```

PR #292, opened 2026-03-25
feat(spider-storage): Implement MariaDB storage.

sitaowang1998:
# Description

This PR implements `MaraiDB` storage functions. We change the storage protocol to pass in expected state for `commit` and `cancel` function because the db does not explicitly store whether there are `commit` and `cleanup` functions.

# Checklist

  breaking change.

# Validation performed

- [x] Tests added in another branch and will be PR later.
- [x] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * ID type exposes a must-use accessor for raw UUID bytes.
  * Database configuration re-exported for easier access.

* **Bug Fixes**
  * Job state transitions now allow Ready → Cancelled.

* **Chores**
  * MariaDB-backed storage implemented with transactional job persistence and simplified storage APIs.
  * Runtime and serialization dependencies updated.
  * Removed an obsolete database error variant.

* **Tests**
  * Test helpers and scaffolding updated to match API changes.


components/spider-core/src/types/id.rs
```diff
@@ -40,6 +40,11 @@ impl<TypeMarker: Debug + PartialEq + Eq> Id<TypeMarker> {
     pub const fn as_uuid_ref(&self) -> &Uuid {
         &self.0
     }
+
+    #[must_use]
+    pub const fn as_bytes(&self) -> &[u8; 16] {
```
LinZhihao-723: Can we create an alias to `uuid::Bytes`:
```rust
pub type UuidBytes = uuid::Bytes;
```
And use `UuidBytes` to replace any use of `[u8; 16]`?

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +63,491 @@ impl MariaDbStorageConnector {
...
+            "INSERT INTO `{table}` (`resource_group_id`, `serialized_task_graph`, \
+             `serialized_job_inputs`) VALUES (?, ?, ?) RETURNING CAST(`id` AS BINARY(16)) AS `id`;",
+            table = JOBS_TABLE_NAME,
+        );
+
+        let serialized_task_graph = task_graph
+            .to_json()
+            .map_err(|e| DbError::TaskGraphSerializationFailure(Box::new(e)))?;
+        let serialized_job_inputs =
+            serde_json::to_string(&job_inputs).map_err(DbError::value_ser)?;
```
LinZhihao-723: Job IOs are in the binary format, I think it's worth serializing as msgpack for efficiency.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +63,491 @@ impl MariaDbStorageConnector {
...
+            .bind(resource_group_id.as_bytes().as_slice())
+            .bind(serialized_task_graph)
+            .bind(serialized_job_inputs)
+            .fetch_one(&self.pool)
+            .await;
+
+        match result {
+            Ok(row) => {
+                let id_bytes: Vec<u8> = row.get(0);
+                job_id_from_bytes(&id_bytes)
```
LinZhihao-723: This might be too inefficient...
Was trying to bind the UUID type with the help of Claude. Can you check if this patch works? (I don't have any test driver code to try it out, might be worth to put up a PR with the testing code).

[uuid.patch](https://github.com/user-attachments/files/26287825/uuid.patch)

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +63,491 @@ impl MariaDbStorageConnector {
...
 #[async_trait]
 impl ExternalJobOrchestration for MariaDbStorageConnector {
     async fn register(
         &self,
-        _resource_group_id: ResourceGroupId,
-        _task_graph: Arc<TaskGraph>,
-        _job_inputs: Vec<TaskInput>,
+        resource_group_id: ResourceGroupId,
+        task_graph: Arc<TaskGraph>,
+        job_inputs: Vec<TaskInput>,
```
LinZhihao-723: Just realized this function doesn't need to take ownership of the inputs. Can we pass in a reference instead?
In reality, we may want to pass the inputs as a ref for registration, and then pass the ownership to the cache layer.

LinZhihao-723 (changes requested):
Partially reviewed with a patch for fixing the ID submission and retrieval.

components/spider-storage/src/db/mariadb.rs
sitaowang1998: Should we do a benchmark to see what's the best way to serialize and/or compression?

components/spider-storage/src/db/mariadb.rs
sitaowang1998: The patch works. I don't really want to put up a PR with 1000+ changes. Maybe next time I'll divide it up into different components instead of implementation+test.

LinZhihao-723: Make sense. I think partial implementation + testing would work.

components/spider-storage/src/db/mariadb.rs
LinZhihao-723: My bad for referring to "efficiency" as the main reason. This is beyond efficiency. I'm not sure if it's a good idea to serialize byte array into a string, considering Rust's string enforces utf8 encoding (unlike C++).

components/spider-core/src/job.rs
LinZhihao-723, line 1: I think the naming of these two types is quite confusing. Looking at how this is used across the storage API, I think it might make sense to just take a boolean to indicate whether the cancellation `has_cleanup_task` or the completion `has_commit_task`.
Adding such a mirror of job state subsets isn't making the code more manageable tbh...

components/spider-core/src/types/id.rs
```diff
@@ -40,29 +41,76 @@ impl<TypeMarker: Debug + PartialEq + Eq> Id<TypeMarker> {
...
+
+impl<'r, TypeMarker, DB> sqlx::Decode<'r, DB> for Id<TypeMarker>
+where
+    TypeMarker: Debug + PartialEq + Eq,
+    DB: Database,
+    Uuid: sqlx::Decode<'r, DB>,
+{
+    fn decode(value: <DB as Database>::ValueRef<'r>) -> Result<Self, sqlx::error::BoxDynError> {
+        Uuid::decode(value).map(|uuid| Self(uuid, PhantomData))
+    }
```
LinZhihao-723: To formalize this part of code, we might want to rename the generic type parameters to match our standard.
I'm ok with `'r` and `'q`, but at least we should use `Db` instead of `DB`.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -232,3 +528,46 @@ CREATE TABLE IF NOT EXISTS `{JOBS_TABLE_NAME}` (
...
+    } else {
+        sqlx::query(UPDATE_JOB_STATE)
+            .bind(new_state)
+            .bind(job_id)
+            .execute(&mut **tx)
+            .await?;
+    }
+
+    Ok(())
+}
```
LinZhihao-723, line 564: Docstring?

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +62,432 @@ impl MariaDbStorageConnector {
...
+        Ok(())
     }
 }
 
+impl DbStorage for MariaDbStorageConnector {}
+
+/// `MySQL` error number for foreign key constraint violation.
+const MYSQL_ER_FK_CONSTRAINT: u16 = 1452;
+/// `MySQL` error number for duplicate entry.
+const MYSQL_ER_DUP_ENTRY: u16 = 1062;
```
LinZhihao-723, line 439: Are these names mirroring MySQL's error code names?

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +62,432 @@ impl MariaDbStorageConnector {
...
+                Ok(job_id)
+            }
+            Err(sqlx::Error::Database(e))
+                if e.try_downcast_ref::<MySqlDatabaseError>()
+                    .is_some_and(|mysql_err| mysql_err.number() == MYSQL_ER_FK_CONSTRAINT) =>
+            {
+                Err(DbError::ResourceGroupNotFound(resource_group_id))
+            }
+            Err(e) => Err(e.into()),
+        }
```
```suggestion
        let job_id: JobId = sqlx::query_scalar(INSERT_QUERY)
            .bind(resource_group_id)
            .bind(serialized_task_graph)
            .bind(serialized_job_inputs)
            .fetch_one(&self.pool)
            .await
            .map_err(|e| match e {
                sqlx::Error::Database(e)
                    if e.try_downcast_ref::<MySqlDatabaseError>()
                        .is_some_and(|mysql_err| mysql_err.number() == MYSQL_ER_FK_CONSTRAINT) =>
                {
                    DbError::ResourceGroupNotFound(resource_group_id)
                }
                e => e.into(),
            })?;
        Ok(job_id)
```

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +62,432 @@ impl MariaDbStorageConnector {
...
     }
 
-    async fn start(
-        &self,
-        _resource_group_id: ResourceGroupId,
-        _job_id: JobId,
-    ) -> Result<(), DbError> {
-        todo!()
+    async fn start(&self, job_id: JobId) -> Result<(), DbError> {
```
LinZhihao-723: Shall we put  `start` and `cancel` to `InternalJobOrchestration`? Also `cancel` seems to be duplicated in both places.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +62,432 @@ impl MariaDbStorageConnector {
...
+            "SELECT `state`, `serialized_job_outputs` FROM `{table}` WHERE `id` = ?;",
+            table = JOBS_TABLE_NAME,
+        );
+
+        let row: Option<(JobState, Option<String>)> = sqlx::query_as(QUERY)
+            .bind(job_id)
+            .fetch_optional(&self.pool)
+            .await?;
+
+        let (state, serialized_outputs) = row.ok_or(DbError::JobNotFound(job_id))?;
```
```suggestion
        let Some((state, serialized_outputs)) = sqlx::query_as::<_, (JobState, Option<String>)>(QUERY)
            .bind(job_id)
            .fetch_optional(&self.pool)
            .await?
        else {
            return Err(DbError::JobNotFound(job_id));
        };
```

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +62,432 @@ impl MariaDbStorageConnector {
...
+        }
+
+        let outputs_str = serialized_outputs.ok_or_else(|| {
+            DbError::CorruptedDbState(format!(
+                "job `{}` succeeded but has no serialized outputs",
+                job_id.as_uuid_ref()
+            ))
+        })?;
+        let outputs: Vec<TaskOutput> =
+            serde_json::from_str(&outputs_str).map_err(DbError::value_de)?;
```
LinZhihao-723: I do believe we should store the binary data in msgpack using bytes then.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +62,432 @@ impl MariaDbStorageConnector {
...
+            "SELECT `state`, `error_message` FROM `{table}` WHERE `id` = ?;",
+            table = JOBS_TABLE_NAME,
+        );
+
+        let row: Option<(JobState, Option<String>)> = sqlx::query_as(QUERY)
+            .bind(job_id)
+            .fetch_optional(&self.pool)
+            .await?;
+
+        let (state, error_message) = row.ok_or(DbError::JobNotFound(job_id))?;
```
LinZhihao-723: Rewrite in the same way?

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +62,432 @@ impl MariaDbStorageConnector {
...
+                .execute(&mut *tx)
+                .await?;
+        } else {
+            sqlx::query(UPDATE_COMMIT_READY_QUERY)
+                .bind(new_state)
+                .bind(&serialized_outputs)
+                .bind(job_id)
+                .execute(&mut *tx)
+                .await?;
+        }
```
```suggestion
        sqlx::query(if new_state.is_terminal() {
            UPDATE_SUCCEEDED_QUERY
        } else {
            UPDATE_COMMIT_READY_QUERY
        })
            .bind(new_state)
            .bind(&serialized_outputs)
            .bind(job_id)
            .execute(&mut *tx)
            .await?;
```

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +62,432 @@ impl MariaDbStorageConnector {
...
+
+        let timeout_secs = expire_after.as_secs();
+        let mut tx = self.pool.begin().await?;
+
+        let job_ids: Vec<(JobId,)> = sqlx::query_as(SELECT_QUERY)
+            .bind(timeout_secs)
+            .fetch_all(&mut *tx)
+            .await?;
+
+        let job_ids: Vec<JobId> = job_ids.into_iter().map(|(id,)| id).collect();
```
LinZhihao-723: Can you check why can't we use `Vec<JobId>` in the previous `query_as`?

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +62,432 @@ impl MariaDbStorageConnector {
...
 
     async fn delete_expired_terminated_jobs(
         &self,
-        _expire_after: Duration,
+        expire_after: Duration,
     ) -> Result<Vec<JobId>, DbError> {
-        todo!()
+        const SELECT_QUERY: &str = formatcp!(
+            "SELECT CAST(`id` AS BINARY(16)) FROM `{table}` WHERE `state` IN \
+             ('Succeeded','Failed','Cancelled') AND `ended_at` < NOW() - INTERVAL ? SECOND FOR \
```
LinZhihao-723: We shouldn't use hardcoded states.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +62,432 @@ impl MariaDbStorageConnector {
...
+                .collect::<Vec<_>>()
+                .join(",");
+            let delete_query =
+                format!("DELETE FROM `{JOBS_TABLE_NAME}` WHERE `id` IN ({placeholders})");
+
+            let mut query = sqlx::query(&delete_query);
+            for job_id in &job_ids {
+                query = query.bind(job_id);
+            }
+            query.execute(&mut *tx).await?;
```
LinZhihao-723: Shall we consider the placeholder limit and do some placeholder batching?

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +62,432 @@ impl MariaDbStorageConnector {
...
+            Err(sqlx::Error::Database(e))
+                if e.try_downcast_ref::<MySqlDatabaseError>()
+                    .is_some_and(|mysql_err| mysql_err.number() == MYSQL_ER_DUP_ENTRY) =>
+            {
+                Err(DbError::ResourceGroupAlreadyExists(
+                    external_resource_group_id,
+                ))
+            }
+            Err(e) => Err(e.into()),
+        }
```
LinZhihao-723: May rewrite as commented in the previous.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +62,432 @@ impl MariaDbStorageConnector {
...
+            None => Err(DbError::ResourceGroupNotFound(resource_group_id)),
+            Some((stored_password,)) => {
+                use subtle::ConstantTimeEq;
+                if stored_password.as_bytes().ct_eq(password.as_bytes()).into() {
+                    Ok(())
+                } else {
+                    Err(DbError::InvalidPassword(resource_group_id))
+                }
+            }
+        }
```
LinZhihao-723: Same, prefer `if Some(expected_password) = sqx::query_as::<_, String>...{} else {}`

components/spider-storage/src/db/protocol.rs
```diff
@@ -55,62 +55,51 @@ pub trait ExternalJobOrchestration {
         &self,
         resource_group_id: ResourceGroupId,
         task_graph: Arc<TaskGraph>,
```
LinZhihao-723: Looks like we can use `&TaskGraph` as well, since we will probably never run this in a detached coroutine.

components/spider-storage/src/db/protocol.rs
```diff
@@ -55,62 +55,51 @@ pub trait ExternalJobOrchestration {
...
 
     /// Cancels a job.
     ///
-    /// The cancelled job will move to:
+    /// The caller specifies the target state via `target`:
     ///
-    /// * [`JobState::CleanupReady`] if the job has a cleanup function.
-    /// * [`JobState::Cancelled`] otherwise.
+    /// * [`CancelTarget::CleanupReady`] if the job has a cleanup task.
+    /// * [`CancelTarget::Cancelled`] otherwise.
```
LinZhihao-723: As suggested in previous, it looks like using a boolean to indicate this should be a cleaner solution before we have the third possible state.

LinZhihao-723 (changes requested):

components/spider-storage/src/db/mariadb.rs
sitaowang1998: Yes and no. `ER_DUP_ENTRY` matches both MySQL error code name. However, the 1452 in MySQL is `ER_NO_REFERENCED_ROW_2`, with description "Cannot add or update a child row: a foreign key constraint fails", which is pretty confusing.

BTW, 1452 is also about foreign key constraint violation. The error code is `ER_ROW_IS_REFERENCED_2`, and the description is "Cannot delete or update a parent row: a foreign key constraint fails".

BTW, MariaDB and MySQL share the error code name.

components/spider-storage/src/db/mariadb.rs
sitaowang1998: For this insert we need to catch error and re-throw manually, can it be rewritten as other query since the error branch also needs pattern matching and is pretty complicated?

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +62,420 @@ impl MariaDbStorageConnector {
...
+        let new_state = if has_cleanup_task {
+            JobState::CleanupReady
+        } else {
+            JobState::Cancelled
+        };
+        let mut tx = self.pool.begin().await?;
+
+        let current_state = fetch_job_state_for_update(&mut tx, job_id).await?;
+        transition_job_state(&mut tx, job_id, current_state, new_state).await?;
+
```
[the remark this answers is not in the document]
sitaowang1998: You can't cancel a job that hasn't started.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,144 +62,420 @@ impl MariaDbStorageConnector {
...
+            failed_state = JobState::Failed.as_str(),
+            cancelled_state = JobState::Cancelled.as_str(),
+        );
+
+        let timeout_secs = expire_after.as_secs();
+        let mut tx = self.pool.begin().await?;
+
+        let job_ids: Vec<JobId> = sqlx::query_scalar(SELECT_QUERY)
+            .bind(timeout_secs)
+            .fetch_all(&mut *tx)
```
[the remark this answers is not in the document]
sitaowang1998: The gc timeout is supposed to be long. It is unlikely to have sub-second timeout. @LinZhihao-723 Do you think we should have a safeguard to either reject too short timeout or round it up?

components/spider-storage/src/db/protocol.rs
```diff
@@ -175,22 +110,32 @@ pub trait ExternalJobOrchestration {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * [`DbError::JobNotFound`] if the `job_id` does not exist.
+    /// * [`DbError::UnexpectedJobState`] if the job is not in [`JobState::Ready`] state.
+    /// * [`DbError::CorruptedDbState`] if the data in the DB is corrupted.
+    /// * Forwards [`sqlx::error::Error`] on DB operation failure.
+    async fn start(&self, job_id: JobId) -> Result<(), DbError>;
```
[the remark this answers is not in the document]
sitaowang1998: `mariadb.rs:202` checks if the job state is `Ready` and returns the `DbError::UnexpectedJobState` first.

components/spider-storage/src/db/mariadb.rs
LinZhihao-723: Ok. I think the names are fair then.

components/spider-storage/src/db/mariadb.rs
LinZhihao-723: What's the difference bewteen this one and the previous commented one [here](https://github.com/y-scope/spider/pull/292#discussion_r3034998030)?

components/spider-storage/src/db/mariadb.rs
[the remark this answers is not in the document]
LinZhihao-723: Seems like we're not using the duration object directly for time control. How about directly receiving `expire_after_sec` as `u64`?

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,152 +62,428 @@ impl MariaDbStorageConnector {
...
+        tx.commit().await?;
+        Ok(())
     }
 }
 
+impl DbStorage for MariaDbStorageConnector {}
+
+/// `MySQL` error number for foreign key constraint violation.
+const MYSQL_ER_FK_CONSTRAINT: u16 = 1452;
+/// `MySQL` error number for duplicate entry.
```
LinZhihao-723, line 438:
```suggestion
const MYSQL_ER_FK_CONSTRAINT: u16 = 1452;

/// `MySQL` error number for duplicate entry.
```
Should always leave an empty line before a docstring block.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,152 +60,429 @@ impl MariaDbStorageConnector {
...
+            "SELECT `state`, `error_message` FROM `{table}` WHERE `id` = ?;",
+            table = JOBS_TABLE_NAME,
+        );
+
+        let row: Option<(JobState, Option<String>)> = sqlx::query_as(QUERY)
+            .bind(job_id)
+            .fetch_optional(&self.pool)
+            .await?;
+
+        let (state, error_message) = row.ok_or(DbError::JobNotFound(job_id))?;
```
```suggestion
        let Some((state, error_message)) =
            sqlx::query_as::<_, (JobState, Option<String>)>(QUERY)
                .bind(job_id)
                .fetch_optional(&self.pool)
                .await?
        else {
            return Err(DbError::JobNotFound(job_id));
        };
```
Please double-check the comment before you mark it as resolved: https://github.com/y-scope/spider/pull/292/changes#r3035007596

components/spider-storage/src/db/mariadb.rs
```diff
@@ -232,3 +515,76 @@ CREATE TABLE IF NOT EXISTS `{JOBS_TABLE_NAME}` (
...
+    })
+    .bind(new_state)
+    .bind(job_id)
+    .execute(&mut **tx)
+    .await?;
+
+    Ok(())
+}
+
+const DELETE_BATCH_SIZE: usize = 1000;
```
LinZhihao-723: 1. This is only used inside one method. We should move it into that method.
2. If you want this to be global, it should be ordered before function methods, right?

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,152 +60,429 @@ impl MariaDbStorageConnector {
...
+
+            let mut query = sqlx::query(&delete_query);
+            for job_id in &batch_job_ids {
+                query = query.bind(job_id);
+            }
+            query.execute(&mut *tx).await?;
+
+            tx.commit().await?;
+            job_ids.extend(batch_job_ids);
+        }
```
```suggestion
        let mut deleted_job_ids: Vec<JobId> = Vec::new();

        loop {
            // We create a new transaction for each batch to limit the locking time.
            let mut tx = self.pool.begin().await?;

            let job_id_batch: Vec<JobId> = sqlx::query_scalar(SELECT_QUERY)
                .bind(expire_after_sec)
                .fetch_all(&mut *tx)
                .await?;

            if job_id_batch.is_empty() {
                break;
            }

            let placeholders = std::iter::repeat_n("?", job_id_batch.len())
                .collect::<Vec<_>>()
                .join(",");
            let delete_query =
                format!("DELETE FROM `{JOBS_TABLE_NAME}` WHERE `id` IN ({placeholders})");

            let mut query = sqlx::query(&delete_query);
            for job_id in &job_id_batch {
                query = query.bind(job_id);
            }
            query.execute(&mut *tx).await?;

            tx.commit().await?;
            deleted_job_ids.extend(job_id_batch);
        }

        Ok(deleted_job_ids)
```
- Rename variables for clarity
- Use `break` and return outside of the loop to make the control flow more friendly to read

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,152 +60,429 @@ impl MariaDbStorageConnector {
...
+            table = JOBS_TABLE_NAME,
+            succeeded_state = JobState::Succeeded.as_str(),
+            failed_state = JobState::Failed.as_str(),
+            cancelled_state = JobState::Cancelled.as_str(),
+        );
+
+        let mut job_ids: Vec<JobId> = Vec::new();
+
+        loop {
+            // We create a new transaction for each batch to limit the locking time.
```
LinZhihao-723: Even though I understand this might not be a problem in the current implementation plan, I'm still a bit hesitant to make this method non-atomic (as a failure call still has side effects).
Since:
- The GC cycle should be long enough to amortize the locking overhead, and
- iiuc, only rows to delete will be locked, which are terminated jobs and should be no longer be updated.

What's the real benefit of splitting them into multiple transactions we care of?

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,152 +60,429 @@ impl MariaDbStorageConnector {
...
+        resource_group_id: ResourceGroupId,
+        password: &[u8],
     ) -> Result<(), DbError> {
-        todo!()
+        const QUERY: &str = formatcp!(
+            "SELECT `password` FROM `{table}` WHERE `id` = ?;",
+            table = RESOURCE_GROUPS_TABLE_NAME,
+        );
+
+        use subtle::ConstantTimeEq;
```
LinZhihao-723, line 411:
```suggestion
        use subtle::ConstantTimeEq;
        
```
For better visual split.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -232,3 +515,76 @@ CREATE TABLE IF NOT EXISTS `{JOBS_TABLE_NAME}` (
         default_state = JobState::Ready.as_quoted_str(),
     )
 }
+
+/// Gets the job state with exclusive lock on the row.
+///
+/// # Parameters
+/// * `tx`: The transaction to execute the query in.
+/// * `job_id`: The ID of the job to fetch the state for.
+///
```
```suggestion
```
No need for param section unless it's a trait declaration.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -54,152 +60,429 @@ impl MariaDbStorageConnector {
...
+        if stored_password.ct_eq(password).into() {
+            Ok(())
+        } else {
+            Err(DbError::InvalidPassword(resource_group_id))
+        }
     }
 
-    async fn delete(&self, _resource_group_id: ResourceGroupId) -> Result<(), DbError> {
-        todo!()
+    async fn delete(&self, resource_group_id: ResourceGroupId) -> Result<(), DbError> {
```
LinZhihao-723: This API will have a problem when integrating with the overall storage system. The most obvious problem is that it deletes the jobs without notifying the cache layer, or there's no way to communicate with the cache layer. That means if a deletion happens while a request attempts to complete a job that belongs to the deleted resource group, the completion request will fail because it can access the cache, but no longer the DB layer. We need to make sure the cache's state is properly reflected before we proceed with such a DB layer deletion.
To properly enforce cache invalidation, this API might need to be redesigned to let the cache layer initiate the resource group deletion. That said, this method might not be as trivial as it looks, and we might want to defer implementing it until we have a full picture of how the overall cache service is implemented. As it's not a blocking feature, shall we just mark it as `todo!("not implemented")` for now?

components/spider-storage/src/db/mariadb.rs
```diff
@@ -232,3 +515,76 @@ CREATE TABLE IF NOT EXISTS `{JOBS_TABLE_NAME}` (
...
+/// # Returns
+///
+/// The current state of the job on success.
+///
+/// # Errors
+///
+/// Returns an error if:
+///
+/// * [`DbError::JobNotFound`] if the `job_id` does not exist.
+/// * Forwards [`sqlx::error::Error`] on DB operations failure.
```
```suggestion
/// * Forwards [`sqlx::query::Query::fetch_optional`]'s return values on failure.
```
I think I brought this up in the previous PR. Please be careful.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -232,3 +515,76 @@ CREATE TABLE IF NOT EXISTS `{JOBS_TABLE_NAME}` (
...
+async fn fetch_job_state_for_update(
+    tx: &mut sqlx::Transaction<'_, sqlx::MySql>,
+    job_id: JobId,
+) -> Result<JobState, DbError> {
+    let row: Option<(JobState,)> = sqlx::query_as(SELECT_JOB_STATE_FOR_UPDATE)
```
LinZhihao-723: `SELECT_JOB_STATE_FOR_UPDATE` seems to be used only inside this method.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -232,3 +515,76 @@ CREATE TABLE IF NOT EXISTS `{JOBS_TABLE_NAME}` (
...
+        .bind(job_id)
+        .fetch_optional(&mut **tx)
+        .await?;
+
+    let (state,) = row.ok_or(DbError::JobNotFound(job_id))?;
+    Ok(state)
```
```suggestion
    let state = sqlx::query_scalar::<_, JobState>(SELECT_JOB_STATE_FOR_UPDATE)
        .bind(job_id)
        .fetch_optional(&mut **tx)
        .await?
        .ok_or(DbError::JobNotFound(job_id))?;
     Ok(state)
```

components/spider-storage/src/db/mariadb.rs
```diff
@@ -232,3 +515,76 @@ CREATE TABLE IF NOT EXISTS `{JOBS_TABLE_NAME}` (
...
+/// * `job_id`: The ID of the job to update the state for.
+/// * `current_state`: The current state of the job. Used for validating state transition.
+/// * `new_state`: The new state to update to.
+///
+/// # Errors
+///
+/// Returns an error if:
+///
+/// * [`DbError::InvalidJobStateTransition`] if the job state transition is invalid.
+/// * Forwards [`sqlx::error::Error`] on DB operations failure.
```
```suggestion
/// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
```

components/spider-storage/src/db/mariadb.rs
```diff
@@ -232,3 +515,76 @@ CREATE TABLE IF NOT EXISTS `{JOBS_TABLE_NAME}` (
...
+            from: current_state,
+            to: new_state,
+        });
+    }
+
+    sqlx::query(if new_state.is_terminal() {
+        UPDATE_JOB_STATE_AND_END
+    } else {
+        UPDATE_JOB_STATE
+    })
```
LinZhihao-723, line 557: Similarly, `UPDATE_JOB_STATE_AND_END` only has one usage.

LinZhihao-723, line 557: Another question (beyond implementation):
How about we don't set `end_time` explicitly. Instead, we use a column to track the last update time which automatically triggers on a DB write. If a job terminates, it shouldn't be updated anyways, and this last update time will be the ending time?
That can save us some verbosity for doing this end time update explicitly.

components/spider-storage/src/db/protocol.rs
```diff
@@ -286,7 +228,8 @@ pub trait InternalJobOrchestration {
     ///
     /// # Parameters
     ///
-    /// * `expire_after` - The duration after termination which a job is considered expired.
+    /// * `expire_after_sec` - The duration after termination which a job is considered expired, in
+    /// seconds.
```
```suggestion
    ///   seconds.
```

LinZhihao-723 (changes requested):

components/spider-storage/src/db/mariadb.rs
sitaowang1998: I can see two ways of solving this:
- `delete` in db returns the job ids deleted, and let cache layer to delete the jobs afterward. This will create dangling jobs for a short period. However, other than wasted time for execution a job that is getting deleted, I don't see a big problem.
- We redesign the delete resource group workflow similar to a two-way commit process.
  1. Mark resource group as to be deleted. Basically it does not accept new jobs.
  2. Cache layer invalidate all jobs of the user.
  3. Delete every job and the user from the db.
  This way we need to track the state of a user beyond just the existence of the foreign key, affecting multiple functions. I need to dig more into the database doc to see if an index could mitigate the overhead.

components/spider-storage/src/db/mariadb.rs
sitaowang1998: We already have an `updated_at` column. My main concern is that the assumption

> if a job terminated, it shouldn't be updated

could break if we add some more internal states or even just some more tracking. I would prefer to explicitly set the `end_at` timestamp to avoid accidental updated.

LinZhihao-723: Fair. I'm ok to keep it untouched.

components/spider-storage/src/db/mariadb.rs
LinZhihao-723: * The first way doesn't solve the problem I mentioned unless it holds the read lock for all touched jobs.
* The second way seems more promising and might be the correct way to implement this deletion.

Anyways, I think there's a very large chance that this API will be redesigned. Let's defer it by marking it unimplemented for now.

components/spider-storage/src/cache/job.rs
LinZhihao-723, line 1: We should probably remove `InternalError::{UndefinedCommitTask, InternalError::UndefinedCleanupTask}`.

LinZhihao-723 (changes requested):
Last catch otherwise we're good.

components/spider-storage/src/cache/job.rs
sitaowang1998: There are other uses of these errors.

LinZhihao-723: Right, right. mb.

LinZhihao-723 (approved):

PR #293, opened 2026-03-30
feat(spider-storage): Add job control block (JCB) for cache-layer job orchestration.

LinZhihao-723:
# Description

This PR introduces the **Job Control Block (JCB)** for Spider Huntsman's cache storage layer, along with a reusable concurrent testing framework that exercises the full job lifecycle.

## Job Control Block (`SharedJobControlBlock`)

The JCB is the central in-memory structure that orchestrates a job's execution in the cache layer. It coordinates three external components via trait abstractions:

- **`ReadyQueueSender`** -- publishes task-ready, commit-ready, and cleanup-ready events to the scheduling queue.
- **`InternalJobOrchestration`** -- persists job state transitions and outputs to the DB layer.
- **`TaskInstancePoolConnector`** -- allocates task instance IDs and registers running instances.

All mutable state (task graph, connectors, queue sender, incomplete-task counter) lives inside `JobExecutionState`, protected by a `tokio::sync::RwLock` through `JobExecutionStateHandle`. The handle provides state-validated accessors (`read_running`, `write_commit_ready`, `write_cancellable`, etc.) that enforce job state invariants before granting access.

### Supported operations

| Method | Description |
|---|---|
| `create` | Factory: builds the JCB from a submitted task graph, job inputs, and connectors. |
| `start` | Transitions from `Ready` to `Running`; enqueues all initially-ready tasks. |
| `create_task_instance` | Registers a new task instance (normal, commit, or cleanup) in the pool. |
| `succeed_task_instance` | Marks a task instance as succeeded; propagates dataflow outputs to children; commits job outputs when all tasks complete. |
| `succeed_commit_task_instance` | Completes the commit task; transitions job to `Succeeded`. |
| `succeed_cleanup_task_instance` | Completes the cleanup task; transitions job to `Cancelled`. |
| `fail_task_instance` | Marks a task instance as failed; retries if eligible, otherwise fails the job. |
| `cancel` | Cancels the job; enqueues the cleanup task if one exists. |

### Job state machine

```
Ready -> Running -> CommitReady -> Succeeded
                 -> Succeeded (no commit task)
                 -> Failed
      -> CleanupReady -> Cancelled
      -> Cancelled (no cleanup task)
```

### Error model

Three error categories in `CacheError`:
- **`InternalError`** -- invariant violations (task graph corruption, undefined termination tasks). Indicates unrecoverable cache state.
- **`StaleStateError`** -- expected concurrent-race rejections (job already terminated, task already succeeded). Safe to ignore.
- **`DbError`** -- forwarded from the DB connector.

## Cache-layer task graph (`cache::task`)

The `TaskGraph` struct holds `SharedTaskControlBlock`s for normal tasks and `SharedTerminationTaskControlBlock`s for commit/cleanup tasks. Each TCB manages:
- Task state (`Pending` -> `Ready` -> `Running` -> `Succeeded`/`Failed`/`Cancelled`)
- Instance pool with configurable concurrency limits (`max_num_instances`)
- Retry counter (`max_num_retry`)
- Dataflow I/O via lock-free `SharedRw<ValuePayload>` buffers
- Parent-dependency tracking for DAG scheduling (decrement-on-complete)

## Integration testing framework

A reusable mock scheduling framework in `tests/test_jcb.rs` that drives concurrent job execution with 64 async worker tasks.

### Mock system architecture

```mermaid
graph LR
    Harness["Test Harness"] --> JCB["JCB"]
    JCB -- "ready tasks" --> Queue["MPMC Queue"]
    Queue --> Workers["64 Workers"]
    Workers -- "register / succeed / fail" --> JCB
    JCB -. "new ready tasks" .-> Queue
    Workers -- "terminal state" --> Harness
    Workers -. "latency samples" .-> Instrument["Instrumentation"]
```

**Data flow**: The harness creates a JCB and calls `start()`, which enqueues initially-ready tasks into the MPMC channel. 64 workers concurrently dequeue one task at a time, call JCB methods (register → succeed/fail), which may enqueue newly-ready child tasks back into the channel. When a worker observes a terminal `JobState`, it broadcasts via a `watch` channel. The harness receives the signal and sends a done notification, causing all workers to exit their `select!` loop.

**Cancellation**: For immediate cancel, the harness calls `jcb.cancel()` synchronously after `start()` but before workers process any messages. Since the MPMC channel is FIFO, workers will dequeue task-ready messages but `create_task_instance` returns `StaleState` (job no longer running). The cleanup-ready message (if any) arrives after all task messages and is processed normally. For concurrent cancel, a separate spawned task cancels after a 1ms delay, racing against worker execution.

### Mock components
- **`MockReadyQueueSender`** -- backed by `async_channel` (true MPMC). Task-ready batches from the JCB are flattened to one message per task index for maximum worker concurrency.
- **`NoopDbConnector`** -- stateless stub returning appropriate state transitions based on commit/cleanup task presence. Generic over `DbConnectorType` so a real connector can be swapped in.
- **`MockTaskInstancePool`** -- atomic counter for ID allocation, no-op registration.

### Worker architecture
- Each worker owns a cloned `WorkerContext` (no shared `Arc` indirection -- all fields are cheaply cloneable via `Arc` or built-in `Clone`).
- Workers `tokio::select!` between the ready-queue receiver and a done signal (`watch<bool>`).
- **Failure injection**: 50% of first-seen tasks spawn 2 concurrent `tokio::spawn` coroutines that each independently register and fail a task instance. This exercises both the retry counter (`max_num_retry = 3`) and the concurrent instance limit (`max_num_instances = 2`).

### Instrumentation

The `InstrumentedJcb` wrapper optionally records per-operation latency for three JCB methods:

- `create_task_instance` (Registration)
- `succeed_task_instance` / `succeed_commit_task_instance` / `succeed_cleanup_task_instance` (Success)
- `fail_task_instance` (Failure)

Each sample records the elapsed `Duration` and whether the call returned `Ok` or `Err`, producing separate `(ok)` and `(err)` rows in the output. When `instrument_sender` is `None`, calls are forwarded directly with zero overhead -- no `Instant::now()` is called.

Instrumentation is enabled by setting the `SPIDER_TEST_INSTRUMENT_OUTPUT_DIR` environment variable. When set, the flat-success and neural-net-success tests create an `mpsc::unbounded_channel`, pass the sender into `run_workload`, and after completion collect all samples into a table formatted by the `tabled` crate. The table is appended to a file named `test_jcb` under the output directory. The table title is derived automatically from the test function name via a `function_name!()` macro.

Example output:
```
test_jcb::test_flat_success_without_db
+--------------------+-------+----------+----------+----------+----------+
| Operation          | Count | Avg (ms) | P50 (ms) | P95 (ms) | P99 (ms) |
+--------------------+-------+----------+----------+----------+----------+
| Registration (ok)  | 20764 | 0.008    | 0.004    | 0.007    | 0.022    |
+--------------------+-------+----------+----------+----------+----------+
| Registration (err) | 4291  | 0.053    | 0.002    | 0.010    | 0.929    |
+--------------------+-------+----------+----------+----------+----------+
| Success (ok)       | 10001 | 0.018    | 0.003    | 0.011    | 0.067    |
+--------------------+-------+----------+----------+----------+----------+
| Success (err)      | 727   | 0.426    | 0.040    | 1.539    | 10.701   |
+--------------------+-------+----------+----------+----------+----------+
| Failure (ok)       | 10036 | 0.005    | 0.003    | 0.005    | 0.009    |
+--------------------+-------+----------+----------+----------+----------+
```

The `(err)` rows capture `StaleState` rejections from concurrent races -- these are expected and help quantify contention. The `Registration (ok)` count should equal `Success (ok)` + `Failure (ok)` since every successfully-registered instance is either succeeded or failed.

### Test workloads (in `tests/task_graph_builder.rs`)
- **Flat**: N independent tasks (configurable count and payload size), each with 1 input and 1 output. Optional commit/cleanup termination tasks.
- **Neural-net**: 10 layers x 1,000 tasks (10,000 total). Each task in layers 1-9 takes 25 random inputs from the previous layer (deterministic via seeded RNG). No termination tasks.

### Generic test design
All test logic is generic over `DbConnectorType` via a factory callback `FnOnce(&SubmittedTaskGraph) -> DbConnectorType`, so the same tests can be rerun with a real DB connector later. Current concrete tests use the `_without_db` suffix.

### Test cases and coverage

| Test | Workload | Behavior | Expected state |
|---|---|---|---|
| `test_flat_success_without_db` | 10k flat + commit/cleanup | Run to completion | `Succeeded` (commit=1, cleanup=0) |
| `test_flat_cancel_without_db` | 10k flat + commit/cleanup | Immediate cancel | `Cancelled` (commit=0, cleanup=1) |
| `test_neural_net_success_without_db` | 10k neural-net | Run to completion | `Succeeded` |
| `test_neural_net_cancel_without_db` | 10k neural-net | Immediate cancel | `Cancelled` |
| `test_always_fail_terminates_job_without_db` | 3 tasks, always fail | All retries exhausted | `Failed` |
| `test_concurrent_success_and_cancel_without_db` | 100 flat + commit/cleanup | Concurrent cancel race | `Succeeded` or `Cancelled` |

These tests collectively verify the following JCB behaviors:

**State transitions**: The full job lifecycle is exercised across all three terminal states (`Succeeded`, `Failed`, `Cancelled`). The flat-success test validates the `Running -> CommitReady -> Succeeded` path (with commit task). The neural-net-success test validates the `Running -> Succeeded` path (without commit task). The cancel tests validate both `Running -> CleanupReady -> Cancelled` (flat, with cleanup task) and `Running -> Cancelled` (neural-net, without cleanup task). The always-fail test validates `Running -> Failed`.

**Concurrency safety**: 64 workers concurrently call `create_task_instance`, `succeed_task_instance`, and `fail_task_instance` on the same JCB. The RwLock-guarded `JobExecutionState` must correctly serialize state mutations while allowing concurrent reads. Failure injection spawns 2 coroutines per task that race against each other, stressing the TCB's mutex-guarded instance pool and retry counter.

**DAG dependency resolution**: The neural-net workload (10-layer, 25-input-per-task DAG) verifies that parent-dependency tracking correctly unblocks child tasks only after all parents succeed. This exercises the `num_unfinished_parents` decrement logic and the ready-child propagation in `succeed_task_instance`.

**Retry and instance limits**: Every task is configured with `max_num_retry = 3` and `max_num_instances = 2`. The failure injection creates 2 concurrent failing instances, consuming retries. The task must eventually be re-enqueued and succeed on a subsequent attempt. The always-fail test verifies that retries are eventually exhausted and the job transitions to `Failed`.

**Termination task execution**: The flat workload includes both commit and cleanup termination tasks. The success test asserts `commit_count == 1` (commit task ran) and `cleanup_count == 0` (cleanup did not run). The cancel test asserts the inverse: `commit_count == 0` and `cleanup_count == 1`. This validates that the correct termination task is dispatched based on the terminal path.

**Task completion accounting**: Success tests assert that `task_success_count == graph.get_num_tasks()`, verifying that every task in the graph was successfully completed exactly once despite concurrent failures and retries. The always-fail test asserts `task_success_count == 0`.

**StaleState error handling**: Workers silently ignore `StaleState` errors from all JCB operations. The concurrent-success-and-cancel test races normal execution against cancellation, producing a nondeterministic mix of `StaleState` errors. The test verifies that the system converges to exactly one of `Succeeded` or `Cancelled` -- no deadlock, no double-termination, no panic.

## Task runner (`taskfiles/test.yaml`)

- Adds `rust-unit-tests` task backed by `huntsman-mariadb-storage-task-executor`.
- The executor spins up a MariaDB Docker container, creates an instrument output directory, runs `cargo nextest run --release` with `SPIDER_TEST_INSTRUMENT_OUTPUT_DIR` set, prints instrument results from all output files, then tears down.

# Checklist

# Validation performed

- All 6 JCB integration tests pass under `cargo test -p spider-storage --test test_jcb` (~0.5s).
- All tests pass with instrumentation enabled (`SPIDER_TEST_INSTRUMENT_OUTPUT_DIR`), producing latency tables for flat and neural-net success workloads.
- `task lint:check-rust` passes cleanly (rustfmt + clippy with pedantic warnings).
- Verified deterministic cancellation behavior: immediate cancel after start produces `Cancelled` for both flat (with cleanup task execution) and neural-net (direct transition) workloads.
- Verified failure injection: 10k-task workloads with 50% random failure on first-seen tasks still complete successfully through retry logic.
- Verified commit/cleanup counters: commit task executes exactly once on success; cleanup task executes exactly once on cancel; neither executes in the wrong scenario.
- Verified `task_success_count` matches `graph.get_num_tasks()` for all success test cases.

## Summary by CodeRabbit

* **New Features**
  * Job control block added: job start/cancel lifecycle, commit/cleanup handling, ready-task publishing, and task-instance pool integration; new task identifier variants supported.

* **Behavior**
  * Tasks can be cancelled non‑terminally; ready/termination signaling and job-state transitions refined.
  * Job-state and identifier utilities improved for ergonomics (eg. copyable markers, running-state query).

* **Tests**
  * Large integration harnesses, task-graph builders and end-to-end tests with optional instrumentation output.

* **Chores**
  * Development/test tooling and crate exports updated to support instrumentation and new connectors.


components/spider-storage/src/cache/job.rs
```diff
@@ -0,0 +1,919 @@
...
+                    .task_graph
+                    .get_task_control_block(task_index)
+                    .ok_or(InternalError::TaskIndexOutOfBound)?;
+                let task_instance_id = job
+                    .task_instance_pool_connector
+                    .get_next_available_task_instance_id();
+                let execution_context = tcb.register_task_instance(task_instance_id).await?;
+                job.task_instance_pool_connector
+                    .register_task_instance(task_instance_id, tcb)
+                    .await?;
```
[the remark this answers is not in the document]
LinZhihao-723: This is considered an internal error. We don't recover the state from an internal error and the entire cache layer needs to restart.

taskfiles/test.yaml
```diff
@@ -214,10 +214,23 @@ tasks:
...
         . "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
         cargo nextest run --all --all-features --run-ignored all --release
+      - |-
+        for f in ${SPIDER_TEST_INSTRUMENT_OUTPUT_DIR}/*; do
+          if [ -f "$f" ]; then
+            echo "=== Instrument: $(basename "$f") ==="
+            cat "$f"
+            echo ""
+          fi
+        done
```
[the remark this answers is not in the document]
LinZhihao-723: That is exactly what we expect.

components/spider-storage/src/cache/error.rs
```diff
@@ -40,6 +43,30 @@ pub enum InternalError {
 
     #[error("task graph input size mismatch: expected {0}, got {1}")]
     TaskGraphInputsSizeMismatch(usize, usize),
+
+    #[error("job not started")]
+    JobNotStarted,
+
+    #[error("unexpected job state: current {current}, expected {expected}")]
```
sitaowang1998: The error message is different from the message for similar database error. We should probably unify them.

components/spider-storage/src/cache/job.rs
```diff
@@ -0,0 +1,919 @@
...
+        id: JobId,
+        owner_id: ResourceGroupId,
+        submitted_task_graph: &SubmittedTaskGraph,
+        inputs: Vec<TaskInput>,
+        ready_queue_sender: ReadyQueueSenderType,
+        db_connector: DbConnectorType,
+        task_instance_pool_connector: TaskInstancePoolConnectorType,
+    ) -> Result<Self, CacheError> {
+        let num_incomplete_tasks = AtomicUsize::new(submitted_task_graph.get_num_tasks());
+        if 0 == submitted_task_graph.get_num_tasks() {
```
sitaowang1998: This two lines together looks weird. I get that we need an atomic for creating a `JobExecutionState`, but avoiding a potential redundant function call is more important.

components/spider-storage/src/cache/job.rs
```diff
@@ -0,0 +1,919 @@
...
+/// * `ReadyQueueSenderType` - The type of the ready queue sender.
+/// * `DbConnectorType` - The type of the DB-layer connector.
+/// * `TaskInstancePoolConnectorType` - The type of the task instance pool connector.
+struct JobControlBlock<
+    ReadyQueueSenderType: ReadyQueueSender,
+    DbConnectorType: InternalJobOrchestration,
+    TaskInstancePoolConnectorType: TaskInstancePoolConnector,
+> {
+    id: JobId,
+    _owner_id: ResourceGroupId,
```
sitaowang1998, line 564: Why the underscore?

components/spider-storage/src/cache/job.rs
```diff
@@ -0,0 +1,919 @@
...
+                    .await?;
+
+                // The lock is intentionally held until just before return so all TCB accesses
+                // observe a consistent state within the lock's scope.
+                drop(job);
+                Ok(ExecutionContext {
+                    task_instance_id,
+                    tdl_context,
+                    timeout_policy,
+                    inputs: Vec::new(),
```
sitaowang1998, line 208: We should add a todo here to fill in the inputs later.

components/spider-storage/src/cache/job.rs
```diff
@@ -0,0 +1,919 @@
...
```
sitaowang1998, line 234: Same as above.

components/spider-storage/src/cache/job.rs
```diff
@@ -0,0 +1,919 @@
...
+            return Ok(job.state);
+        }
+
+        if num_incomplete_tasks != 0 {
+            return Ok(job.state);
+        }
+
+        // Release the read lock prior to acquiring a write lock for committing job outputs.
+        drop(job);
+        let mut job = jcb.job_execution_state.write_running().await?;
```
sitaowang1998, line 309: This is the second time I see this pattern. While currently dropping read lock and get the write lock again does not introduce any races, could we have a method to promote the lock?

components/spider-storage/src/cache/job.rs
```diff
@@ -0,0 +1,919 @@
...
+        job.task_graph
+            .get_cleanup_task_control_block()
+            .ok_or(InternalError::UndefinedCleanupTask)?
+            .succeed_task_instance(task_instance_id)
+            .await?;
+        job.db_connector
+            .set_state(jcb.id, JobState::Cancelled)
+            .await?;
+        job.state = JobState::Cancelled;
+        drop(job);
```
sitaowang1998, line 403: Some functions have `drop` just before last return, some don't. Let's make it consistent. Either all `drop` or all don't.

sitaowang1998 (changes requested):
Just finish the `job.rs`. Look good overall.

components/spider-storage/src/ready_queue.rs
```diff
@@ -0,0 +1,55 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * [`InternalError`] if the tasks fail to be sent to the ready queue.
+    async fn send_task_ready(
+        &self,
+        job_id: JobId,
+        task_indices: Vec<TaskIndex>,
```
sitaowang1998, line 26:
```suggestion
        task_indices: &[TaskIndex],
```

sitaowang1998: Also it would be easier if the queue could store more information about the ready tasks. Otherwise, the storage server need to fetch them from the cache layer, which does not provide these functionality so far.

sitaowang1998 (changes requested):
Before I dig into the tests, there is a point I miss about the cache layer in previous review. It is fine we don't handle the internal errors. Does it mean if the storage server sees an internal error, it will kill itself and restart, or maybe just restart the faulting job from beginning? We need to be clear about the error handling so that it is safe to ignore some potential race condition, especially for the counters.

components/spider-storage/src/ready_queue.rs
LinZhihao-723: - We can update the API to include any other information needed later. This is just a basic skeleton to make things work.
- No, it should take the ownership of the vector so it's free to send it through a channel without making a copy.

components/spider-storage/src/cache/job.rs
LinZhihao-723: All of these are suggested by clippy, and the rule is:
- If the return statement ever needs to use the job, then it shouldn't need an explicit drop.
- Otherwise, the explicit drop makes the resource to be release before the return statement, which is "technically" shorterning the critical region in clippy's view.

components/spider-storage/src/cache/job.rs
LinZhihao-723: I don't think commit/cleanup tasks need any inputs (as we discussed)?
We could make inputs nullable (using Option), but on the otherhand it would be hard to differentiate inputs being None vs. inputs being an empty vector. So I decided to go with empty vector to simplify the receiver end handling (inside execution manager).

components/spider-storage/src/cache/job.rs
LinZhihao-723: This member isn't used in any of the code yet (because we haven't implemented RG validation yet).
Without underscore, clippy would suggest you to remove it. Underscore in Rust means it's unused but it's there on purpose, and clippy won't complain in that way.

components/spider-storage/src/cache/job.rs
LinZhihao-723: I've thought about this when I coded these lines. But I ended up doing it in the current way because tokio doens't support read-to-write lock promotion. To add such a wrapper API, you will need to have a method that takes the ownership of the reader guard and a reference to the RW lock. That is actually making readability worth because:
- The reader guard is dropped inside this wrapper implictly through the ownership transfer.
- There's no way to guarantee the given reader guard is generated from the given RW lock.

In constrast, I think the current pattern is simplier to understand and maintain, at least the lifetime of the reader guard is much obvious. If we really need sth to encapsulate lock promotion, we should probably design our own wrapper on top of tokio's RW lock, and that's a bit overkill atm.

LinZhihao-723:
> Before I dig into the tests, there is a point I miss about the cache layer in previous review. It is fine we don't handle the internal errors. Does it mean if the storage server sees an internal error, it will kill itself and restart, or maybe just restart the faulting job from beginning? We need to be clear about the error handling so that it is safe to ignore some potential race condition, especially for the counters.

I actaully documented this in the docstring of the two error enums: https://github.com/y-scope/spider/blob/e9013e468b82cac89dfacfa79f6060372b50e627/components/spider-storage/src/cache/error.rs#L13.

- Yes, we need to restart the storage server. These errors should behave like a panic. However, in tokio runtime, panic only kills the current coroutine, so we need to report it as an error with a dedicated type to let the top-level error handler to know that this error should cause the service to go down and restart.
- iiuc, all the "safe to ignore" issues are the error types marked as `StaleStateError` (defined below `InternalError`). The clarification behavior you're asking for is exactly the reason to define `CacheError` as an enum of `StaleStateError`, `InternalError`, and `DbError`.

components/spider-storage/src/cache/job.rs
sitaowang1998: I think it is fine right now. `tokio` does not provide a real lock promotion, and whatever wrapper we write would hide the problem and make it look like we have a proper lock promotion without lock dropped in the middle. This only make things worse.

components/spider-storage/src/ready_queue.rs
sitaowang1998: It is unlikely any queue implementation would just directly move the vector inside a queue, but I'm ok with using `Vec` since using slice could only benefits the caller, which is not an issue here.

sitaowang1998 (commented):
The test code itself looks good. However, it would be better if we can split it up into different files. Especially we might reuse some mock later in other components.

LinZhihao-723:
> The test code itself looks good. However, it would be better if we can split it up into different files. Especially we might reuse some mock later in other components.

Moved to a dedicated mod and add the PR description as the mod-level docstring as discussed offline.

sitaowang1998 (approved):

PR #299, opened 2026-04-07
test(spider-storage): Add tests for MariaDB storage.

sitaowang1998:
# Description

This PR:
- Adds unit tests for db layer targeting the MariaDB implementation.
- Extends the cache layer test infra to support testing with both mock db and real MariaDB.
- Restructures tests to avoid dead code.

# Checklist

  breaking change.

# Validation performed

- [x] New tests pass locally on Ubuntu 22.04.
- [x] Serial tests run after all other tests.
- [x] GitHub workflows pass.

## Release Notes

* **Tests**
  * Added comprehensive MariaDB integration test suite validating job registration, orchestration, state transitions, cancellation, output handling, and resource group management.
  * Enhanced test infrastructure with new database connector factory pattern supporting multiple database backends.
  * Improved test utilities for database configuration and test resource setup.


components/spider-storage/tests/test_mariadb.rs
```diff
@@ -0,0 +1,701 @@
...
+use spider_storage::db::{
+    DbError,
+    ExternalJobOrchestration,
+    InternalJobOrchestration,
+    ResourceGroupManagement,
+};
+
+fn minimal_task_graph() -> TaskGraph {
+    TaskGraph::default()
+}
```
LinZhihao-723: Technically we don't allow task graph with zero tasks in the cache layer. It might be worth to test with a task graph with one task. You can use the task graph builder to build a flat task graph with a single task, and have a const for the inputs?

components/spider-storage/tests/test_mariadb.rs
```diff
@@ -0,0 +1,701 @@
...
+
+// ─── ExternalJobOrchestration ───────────────────────────────────────────────
```
LinZhihao-723:
```suggestion
```
We should avoid visual guides in comments/docstrings.
Please also remove the others.

components/spider-storage/tests/test_mariadb.rs
```diff
@@ -0,0 +1,701 @@
...
+    let rg_id = create_test_resource_group(&storage).await;
+
+    let job_id = storage
+        .register(rg_id, &minimal_task_graph(), vec![].as_slice())
+        .await
+        .expect("register should succeed");
+
+    storage.start(job_id).await.expect("start should succeed");
+
+    InternalJobOrchestration::cancel(&storage, job_id, false)
```
LinZhihao-723:
```suggestion
    storage.cancel(job_id, false)
```
Why not call it directly? May need to apply to other occurrences.

.config/nextest.toml
LinZhihao-723, line 1: I don't think we need to config this manually, as using `test_serial` as the prefix for serialized execution is not very obvious and it's hard to document such a behavior.
In log-ingestor, we were using [`serial_test::serial`](https://github.com/y-scope/clp/blob/ad31b8948c5913b9b82165b2d2d0ae5dcd56b984/components/log-ingestor/tests/test_ingestion_job.rs#L270) to make sure some test cases are executed in a serialized manner. I think it should be enough for our use case.

components/spider-storage/tests/mariadb_test.rs
LinZhihao-723, line 1: Don't have any comments on the test cases themselves. Probably just ask claude to apply the style changes I suggest.

components/spider-storage/tests/mariadb_infra.rs
LinZhihao-723, line 1: All docstrings miss `# Returns` section. Please fix.

components/spider-storage/tests/mariadb_infra.rs
```diff
@@ -0,0 +1,47 @@
...
+    db::{MariaDbStorageConnector, ResourceGroupManagement},
+};
+
+/// Creates a [`MariaDbStorageConnector`] from environment variables.
+///
+/// # Panics
+///
+/// Panics if any required environment variable (`MARIADB_PORT`, `MARIADB_DATABASE`,
+/// `MARIADB_USERNAME`, `MARIADB_PASSWORD`) is missing or if the connection fails.
+pub async fn setup() -> MariaDbStorageConnector {
```
LinZhihao-723: `setup` -> `create_mariadb_connector`?
`setup` is too general, especially called without the mod name. Prefer to be explicit so we know the returned object is a mariadb connector.

components/spider-storage/tests/scheduling_infra.rs
```diff
@@ -421,6 +426,31 @@ const NUM_WORKERS: usize = 64;
...
+///
+/// The returned closure receives the submitted task graph and job inputs from [`run_workload`],
+/// registers the job via [`ExternalJobOrchestration::register`], and returns the connector along
+/// with the resulting [`JobId`] and [`ResourceGroupId`].
+///
+/// # Panics
+///
+/// Panics if job registration fails.
+#[must_use]
+pub fn mariadb_db_connector_factory(
```
LinZhihao-723, line 452: The ordering was messed up. This method should not be placed in the middle of private const and private types.

components/spider-storage/tests/scheduling_infra.rs
LinZhihao-723, line 1: A better way to achieve what you want using `AsyncFnOnce`:
[async-func.patch](https://github.com/user-attachments/files/26557662/async-func.patch)

components/spider-storage/tests/test_jcb.rs
```diff
@@ -273,38 +222,77 @@ async fn test_concurrent_success_and_cancel<
...
 #[tokio::test(flavor = "multi_thread")]
 async fn test_concurrent_success_and_cancel_without_db() {
-    test_concurrent_success_and_cancel(noop_db_connector).await;
+    test_concurrent_success_and_cancel(noop_db_connector()).await;
+}
+
+// ─── MariaDB integration tests ─────────────────────────────────────────────
+//
+// Neural-net MariaDB tests are omitted: the 25,000-input payload (10 layers × 1,000 tasks × 25
+// inputs) exceeds MariaDB's `max_allowed_packet` limit during job registration.
```
LinZhihao-723: - We should only have 25 * 1000 inputs right?
- Shall we try to fix the input graph to make it mariadb compatible? We can change its first layer to an input layer so it only takes 1000 inputs. I'm fine either if you want to split it into a new PR or add it in this one.

components/spider-storage/tests/test_jcb.rs
```diff
@@ -273,38 +222,77 @@ async fn test_concurrent_success_and_cancel<
...
+
+use mariadb_infra::{create_test_resource_group, setup as mariadb_setup};
+
+#[tokio::test(flavor = "multi_thread")]
+#[ignore = "requires MariaDB"]
+async fn test_flat_success_with_mariadb() {
+    let storage = mariadb_setup().await;
+    let rg_id = create_test_resource_group(&storage).await;
+    test_flat_success(mariadb_db_connector_factory(storage, rg_id), None).await;
```
LinZhihao-723: We should probably update the test case to return the job ID, the expected terminal state, and an optional output. And we should check the expected values are properly reflected in DB.

LinZhihao-723 (changes requested):

components/spider-storage/tests/jcb_test.rs
```diff
@@ -29,19 +32,10 @@ macro_rules! function_name {
...
-///
-/// * `DbConnectorType` - The DB-layer connector implementation.
-/// * `DbConnectorFactoryType` - A factory that creates `DbConnectorType` from the submitted task
-///   graph.
-async fn test_flat_success<
-    DbConnectorType: InternalJobOrchestration + 'static,
-    DbConnectorFactoryType: FnOnce() -> DbConnectorType,
->(
-    db_connector_factory: DbConnectorFactoryType,
+async fn test_flat_success<Db: InternalJobOrchestration + 'static>(
```
LinZhihao-723: * We shouldn't remove the type parameter section as we still take one type parameter.
* We should rename `Db` back to `DbConnectorType` for clarity: Db sounds like we're testing different types of DBs.
* We should add a return annotation:
  ```Rust
  /// # Returns
  ///
  /// Forwards `run_workload`'s return values.
  ```
Do it for all test cases.

components/spider-storage/tests/jcb_test.rs
```diff
@@ -267,44 +221,168 @@ async fn test_concurrent_success_and_cancel<
...
+    let result = test_flat_success(mariadb_db_connector_factory(storage, rg_id), None).await;
+
+    let verifier = create_mariadb_connector().await;
+    let db_state = verifier
+        .get_state(result.job_id)
+        .await
+        .expect("get_state should succeed");
+    assert_eq!(
+        db_state,
+        JobState::Succeeded,
```
LinZhihao-723:
```suggestion
        result.terminal_state,
```
Do it for other test cases as well.

components/spider-storage/tests/mariadb_test.rs
```diff
@@ -0,0 +1,727 @@
...
+};
+
+/// Input payload size in bytes for the single-task graph used by DB-layer tests.
+const TEST_INPUT_PAYLOAD_SIZE: usize = 128;
+
+/// Builds a task graph with a single task for DB-layer tests.
+///
+/// # Returns
+///
+/// The submitted task graph and corresponding job inputs.
```
LinZhihao-723:
```suggestion
/// Forwards `build_flat_task_graph`'s return values.
```

LinZhihao-723 (changes requested):
Some more style catch. Otherwise lgtm.

components/spider-storage/tests/scheduling_infra.rs
```diff
@@ -270,11 +294,8 @@ pub fn write_instrument_results(
 ///
 /// # Type Parameters
 ///
-/// * `DbConnectorType` - The DB-layer connector implementation. Must be `'static` so that worker
-///   tasks can be spawned onto the tokio runtime.
-/// * `DbConnectorFactoryType` - A factory that creates the DB connector from the submitted task
-///   graph. This allows the caller to defer connector construction until the task graph is
-///   available, avoiding redundant graph builds.
+/// * `Db` - The DB-layer connector implementation. Must be `'static` so that worker tasks can be
```
LinZhihao-723: `Db` -> `DbConnectorType`

components/spider-storage/tests/scheduling_infra.rs
```diff
@@ -193,14 +196,35 @@ pub struct WorkloadResult {
...
 }
 
-/// Creates a [`NoopDbConnector`] from the given submitted task graph.
+/// An async DB connector factory for [`run_workload`].
 ///
-/// # Returns
+/// # Type Parameters
 ///
-/// The created no-op connector.
+/// * `DbConnector` - The DB-layer connector implementation.
```
LinZhihao-723: `DbConnectorType`

LinZhihao-723 (changes requested):

LinZhihao-723 (approved):
For the PR title, how about:
```
test(spider-storage): Add tests for MariaDB storage.
```
Which matches #292.

PR #303, opened 2026-04-13
feat(spider-storage): Implement ready queue.

sitaowang1998:
# Description

This PR:
- Adds receiver interface for the ready queue.
- Extends sender interface to include entry removal.
- Implements ready queue.

## Workflow

When a task is ready to run, it as added into ready queue using one of the three `send_*_ready`, depends on the task type. The task is stored in one of the three corresponding queue, and is assigned a monotonically increasing queue id.

Scheduler polls from the one of the three queue with the start queue id and the upper limit. The queue returns a list of tasks without removing them from the queue, so that when the scheduler is down, these tasks remains ready to be scheduled. When a new scheduler is spawned, it can ask for ready tasks from the very start of the queue by setting the start queue id as 0.

When a task terminated, it can remove corresponding task from the ready queue. If a job is cancelled/failed, it can remove all tasks of the job from the ready queue.

# Checklist

  breaking change.

# Validation performed

- [x] Adds unit tests for the ready queue implementation.
- [ ] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Added batch dequeue operations for task, commit, and cleanup queues with cursor-based pagination support.
  * Introduced synchronous removal capabilities for task and job queue entries.
  * Enhanced resource group tracking throughout queue operations.

* **Refactor**
  * Extended ID types to support hashing, enabling broader collection usage.
  * Improved job ownership tracking in control flow.


LinZhihao-723 (dismissed):
Thanks for the PR — the ready queue design with `BTreeMap` + secondary index is solid and the test coverage is thorough. I have a few items to address, mostly around the project's `unwrap` policy and docstring conventions.

## Summary

1. **`unwrap()` on mutex locks**: There are 11 occurrences of `.lock().unwrap()` across `ReadyQueueSenderHandle` and `ReadyQueueReceiverHandle`. Per coding style, `unwrap` is not allowed anywhere — use `.expect("descriptive message")` instead (e.g., `.lock().expect("ready queue lock poisoned")`).
2. **Missing docstrings on public structs**: `ReadyQueueSenderHandle` and `ReadyQueueReceiverHandle` are `pub` structs without docstrings.
3. **Inconsistent explicit `drop()` calls**: Some methods call `drop(queue)` explicitly after releasing the lock, others don't. Either add them consistently or remove them all (the guard drops at scope exit anyway).
4. **Test functions should return `anyhow::Result<()>`**: Per testing conventions, test functions should return `anyhow::Result<()>` and use `?` instead of `.expect()` where applicable.
5. **`remove_job_entries` efficiency**: The current implementation does a linear scan of all keys in `job_task_index`. Consider a nested index (`HashMap<JobId, HashMap<TaskId, BTreeSet<u64>>>`) for O(1) job lookups. Not blocking, but worth noting for when the queue grows.

sitaowang1998:
Closed as design changed.

Issue #302, opened 2026-04-13
feat(spider-tdl): Add TDL package framework for user-defined task execution


LinZhihao-723:
## Summary

Implement the TDL (Task Definition Language) package framework that allows users to write custom
task functions in Rust, package them into a shared library (cdylib), and have the task executor
load and invoke these functions at runtime via C-FFI.

## Motivation

The spider system needs a pluggable task execution model where:

- Users define task functions with typed parameters and return values.
- Task packages are compiled as shared libraries and loaded at runtime by the execution process.
- Inputs and outputs cross process boundaries as serialized byte streams (msgpack + wire format).
- Multiple packages can coexist, each with its own namespace of tasks.

### New crates

| Crate | Path | Type | Purpose |
|---|---|---|---|
| `spider-tdl` | `components/spider-tdl` | lib | Shared types, traits, wire serde, registration macro |
| `spider-tdl-derive` | `components/spider-tdl-derive` | proc-macro | `#[task]` attribute macro |
| `spider-task-executor` | `components/spider-task-executor` | lib | Loads cdylib packages, calls C-FFI |

### Type system

The TDL type system supports:

- **Primitives:** `int8`/`i8`, `int16`/`i16`, `int32`/`i32`, `int64`/`i64`, `float`/`f32`, `double`/`f64`, `boolean`/`bool`
- **Bytes:** `Vec<u8>` / `Bytes`
- **Collections:** `List<T>` (`Vec<T>`), `Map<K, V>` (`HashMap<K, V>`) where K is restricted to integer types and `Bytes`
- **User-defined structs** with serde `Serialize + Deserialize`

### `#[task]` proc-macro

Transforms a user-defined function into a `Task` trait implementation:

```rust
#[task(name = "my_namespace::add")]
fn add(ctx: TaskContext, a: int32, b: int32) -> Result<(int32,), TdlError> {
    Ok((a + b,))
}
```

Generates:
- A marker struct with the function name
- A params struct with `#[derive(Deserialize)]`
- An `impl Task` with `NAME`, `Params`, `Return`, and `execute()`
- `serialize_return()` has a default implementation that uses a custom serde `Serializer` to decompose tuples into individually-encoded wire payloads

Single-value returns (e.g., `Result<int32, TdlError>`) are auto-wrapped into `(int32,)`.

### Wire format

A two-layer serialization scheme:

| Layer | Format | Purpose |
|-------|--------|---------|
[...]

PR #304, opened 2026-04-13
feat(spider-tdl): Add `spider-tdl` crate foundation:

LinZhihao-723:
* Add `TdlError` for task execution errors forwarding across the C-FFI boundary.
* Add C-FFI wrapper for communication between the task executor and the TDL package library.
* Add `TaskContext` for runtime metadata passed to every task function.
* Add standard TDL type alias.

# Description

This PR is the first PR in serious #302.

This PR implements all items mentioned in the multiline PR title with the following notice:

* The C-FFI layer is partially learned from log-surgeon C-FFI implementation:
  * C array: https://github.com/y-scope/log-surgeon/blob/3e50583e542ee49da61c0fcc89182ab15bb5e52a/rust/src/c_interface.rs#L24-L33
* The TDL type alias will not be enforced in Rust TDL. In fact, it is also hard to enforce because at compile time, we only have token-level type information in a proc macro -- we don't have global symbol table access. The document suggests users to use standard TDL types for cross-language support.

# Checklist

  breaking change.

# Validation performed

* [x] Add unit tests to cover basic serde behaviors and C-ffi wrappers.

## Summary by CodeRabbit

* **New Features**
  * Added a Task Definition Language component exposing FFI-friendly data types, an error type, runtime task context, and TDL-aligned public type aliases for cross-language use.
* **Chores**
  * Registered the new component in the workspace.
* **Tests**
  * Added unit tests for serialization round-trips and FFI/encoding behaviors.


components/spider-tdl/src/ffi.rs
```diff
@@ -0,0 +1,283 @@
...
+        self.length
+    }
+
+    /// # Returns
+    ///
+    /// Whether the array is empty.
+    #[must_use]
+    pub const fn is_empty(&self) -> bool {
+        self.length == 0
+    }
```
modelconsumer: It may be preferable to replace this with a `std::ops::Deref<Target = [ElementType]>` implementation, so all the other slice methods that may be needed in the future come for free. On the other hand, keeping this type minimal and requiring conversion with `.as_slice()` may be preferable for explicitness

Either way can work/either may be better depending on how its used - maybe worth a doc comment explaining your preference (if applicable - doesn't have to have some meaningful reason)

components/spider-tdl/src/ffi.rs
```diff
@@ -0,0 +1,283 @@
...
+impl<'borrow_lifetime, ElementType> CArray<'borrow_lifetime, ElementType> {
+    /// Borrows `slice` as a C-ABI array view.
+    ///
+    /// The returned [`CArray`] is tied to the lifetime of `slice`; the pointer remains valid as
+    /// long as the original slice is not moved or dropped.
+    ///
+    /// # Returns
+    ///
+    /// The constructed C array from the given slice.
+    pub const fn from_slice(slice: &'borrow_lifetime [ElementType]) -> Self {
```
modelconsumer, line 55: `#[must_use]`

components/spider-tdl/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+pub mod error;
+pub mod ffi;
+pub mod r#std;
```
modelconsumer, line 3: This works but I wonder if there's another name that could be used.. maybe `types`, or even `aliases`?

components/spider-tdl/src/ffi.rs
```diff
@@ -0,0 +1,283 @@
...
+            is_error,
+            pointer,
+            length,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{CByteArray, CCharArray, TaskExecutionResult};
```
modelconsumer: I think it's fine to use `use super::*;` in test modules

modelconsumer (commented):
Mainly looked at `ffi.rs`

sitaowang1998 (commented):
Other parts look ok to me.

components/spider-tdl/src/ffi.rs
```diff
@@ -0,0 +1,283 @@
...
+    /// # Safety
+    ///
+    /// The caller must guarantee that:
+    ///
+    /// * `pointer` points to a single, contiguous allocation of at least `length` elements of
+    ///   `ElementType`, properly initialized.
+    /// * The memory remains valid and immutable for the returned lifetime.
+    #[must_use]
+    pub const fn as_slice(&self) -> &'borrow_lifetime [ElementType] {
+        unsafe { std::slice::from_raw_parts(self.pointer, self.length) }
```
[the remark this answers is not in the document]
LinZhihao-723: @modelconsumer Do you think this is necessary?

components/spider-tdl/src/lib.rs
LinZhihao-723: We'd like to emphasize that the type aliases in this mod are the "standard" TDL types. Do you think it's a problem to refer them as `spider_tdl::std::int8`?

components/spider-tdl/src/ffi.rs
LinZhihao-723: Surprisingly this was not caught by the linter...

components/spider-tdl/src/ffi.rs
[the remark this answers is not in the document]
modelconsumer: In terms of the warning, I think a `// SAFETY: blah blah blah` comment before the unsafe block will suffice. In terms of how to treat these FFI values, I'm not 100% sure what's best, but I lean slightly towards treating `CArray` (and everything else) as "pure Rust" types that _happen_ to be FFI-callable - i.e.

1. In Rust, you shouldn't/it shouldn't be possible to create an invalid `CArray`; outside this module, it behaves like a regular slice with a lifetime (e.g. no way to edit the private fields), so for auditing purposes, you just need to make sure the implementation is correct in this module. In Rust, you can call these functions (e.g. in tests) and they really are safe w.r.t. the compiler
2. Yes, the FFI boundary is inherently unsafe, but IMO it's not so meaningful to then mark (almost) every FFI function as unsafe. My mental model is "FFI needs to uphold the same requirements in Rust", i.e. you shouldn't pass an invalid value (e.g. dangling pointer) that couldn't have been created in Rust

In any case, you can/should still try adding a `// SAFETY: ` comment, maybe referring to some module level explanation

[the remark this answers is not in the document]
LinZhihao-723: Thanks!

sitaowang1998 (approved):

PR #305, opened 2026-04-14
feat(ci): Add path-filtering to execute linting and testing workflow conditionally based on the files changed; Move wolf test files into `wolf` directory.

LinZhihao-723:
# Description

This PR adds path-based filtering for GitHub workflows so that they're triggered conditionally based on the files changed. In particular:

- For linting workflows:
  - Toml and yaml linting will always be executed.
  - C++ linting will only be triggered if C++/cmake files are changed.
  - Python linting will only be triggered if Python files are changed.
  - Rust linting will only be triggered if files under Rust project are changed.
- For testing:
  - Wolf version testing is triggered only if C++ or Python files are changed.
  - Huntsman version testing is triggered only if Rust project files are changed.

The filtering logic is learned from CLP's workflow from [here](https://github.com/y-scope/clp/blob/main/.github/workflows/clp-artifact-build.yaml), and is made into an action for reuse across different workflows.

# Checklist

  breaking change.

# Validation performed

- PR with only Python changes won't trigger C++/Rust/Huntsman related workflows: https://github.com/LinZhihao-723/spider/pull/3
- PR with only Rust changes won't trigger Python/C++/Wolf related workflows: https://github.com/LinZhihao-723/spider/pull/4
- PR with only C++ changes won't trigger Python/Rust related workflows: https://github.com/LinZhihao-723/spider/pull/5

## Release Notes

* **Chores**
  - Restructured test directory organization for improved maintainability
  - Implemented conditional CI/CD workflows that selectively run linting and tests based on the types of files changed
  - Updated build system configurations and paths to align with the reorganized test structure


tests/wolf/integration/test_client.py
```diff
@@ -25,7 +25,7 @@ def start_scheduler_workers(
     """
     # Start the scheduler
     dir_path = Path(__file__).resolve().parent
-    dir_path = dir_path / ".." / ".." / "src" / "spider"
+    dir_path = dir_path / ".." / ".." / ".." / "src" / "spider"
```
sitaowang1998, line 28: Prefer to use `Path.parents` for better portability. Same for other python files.

.github/workflows/code-linting-checks.yaml
```diff
@@ -17,7 +17,48 @@ concurrency:
...
+        run: "./tools/scripts/lib_install/linux/install-dev.sh"
+
+      - run: "task lint:toml-check"
+
+      - run: "task lint:yml-check"
+
+  lint-cpp:
+    needs: "filter-relevant-changes"
+    if: >-
+      github.event_name == 'schedule'
```
sitaowang1998, line 59: Why use single quote instead of double quote?

sitaowang1998 (changes requested):

.github/workflows/code-linting-checks.yaml
LinZhihao-723: Trying to match CLP side (which also uses single quote). Will double check tmr to see if we can use double quote.

sitaowang1998: That one uses single quote because it is inside a double quote. Here we use multi line statement.

tests/wolf/integration/test_client.py
LinZhihao-723: This is actually making it consistent with other test files, lol
And I found using parents really confusing since the indexing system starts from 0, so parents[1] is actually the second parent.
Last but not least, I don't think these tests have any portability, lol. Considering we will have very little chance to touch it again, I won't spend more time looking into any style trade off and just keep it in a state that it works.

.github/workflows/code-linting-checks.yaml
LinZhihao-723: I guess it's wrong in this place as well then: https://github.com/y-scope/clp/blob/9ae078ee1a90f948fd5c8af334d551cbb099b43d/.github/workflows/clp-artifact-build.yaml#L220.
Will check more on whether it's a fundamental limitation.

LinZhihao-723: Double quote won't work: https://github.com/LinZhihao-723/spider/actions/runs/24410858591

sitaowang1998 (approved):

PR #308, opened 2026-04-15
feat(spider-tdl): Add wire format to serialize/deserialize task inputs and task outputs.

LinZhihao-723:
# Description

- This PR is the first PR in serious #302.
- This PR depends on #304.

This PR implements the wire format for serializing and deserializing task inputs and outputs, with the design goal of minimizing memory copies required.

For task inputs, `TaskInputsSerializer` is designed to serialize task inputs' payloads into the wire format in a streaming flavor. The serialized byte buffer will be transmitted all the way down to the TDL package library without needing to unpack. Inside the TDL package task execution, the byte buffer will be deserialized into a task param struct in place. The memory copies involved:

* Storage layer serialization: Streamingly serialize task inputs from JCB into the buffer
* Network transmission: transmit from the storage service to the execution manager
* Inter-process transmission: transmit from the execution manager to the task executor process
* TDL package deserialization: deserialize the buffer into the task params, which will be directly used by the task execution

For network transmission and inter-process transmission, they're abstracted by the libraries we use so there might be hidden inefficiency, but this is out of our control. For the buffer serialization and deserialization, this is the best we can do to reduce the number of memory copies needed.

For task outputs, `TaskOutputsSerializer` is designed to serialize a tuple of task outputs of a concrete type into the wire format, while each frame is encoded using msgpack (which is the format we store in JCB). The serialized buffer is passed all the way back to the storage layer, and unpacked into raw payloads. The storage layer never interprets these payload bytes. These payloads can be serialized directly using `TaskInputsSerializer` for child tasks. And overall, it has the same efficiency as the task inputs serialization/deserialization, which minimizes the number of memory copies required.

# Checklist

  breaking change.

# Validation performed

* [x] Add unit tests to test serde round-trip and corrupted deserialization.

## Summary by CodeRabbit

* **New Features**
  * Added wire protocol implementation featuring length-prefixed framing for serialization and deserialization of task data
  * Improved library accessibility through public module declarations and re-exported types at the crate root


components/spider-tdl/src/wire.rs
```diff
@@ -0,0 +1,963 @@
...
+    ///   the declared field length extends past the end of the buffer.
+    fn unframe_payloads(data: &[u8]) -> Result<Vec<Vec<u8>>, WireError> {
+        if data.len() < COUNT_HEADER_LEN {
+            return Err(WireError::InvalidFormat(
+                "buffer too small for the payload count header",
+            ));
+        }
+        let count_bytes: [u8; COUNT_HEADER_LEN] = data[..COUNT_HEADER_LEN]
+            .try_into()
+            .expect("slice length checked above");
```
sitaowang1998: This error message is confusing.

components/spider-tdl/src/wire.rs
```diff
@@ -0,0 +1,963 @@
...
+        let mut payloads = Vec::with_capacity(count);
+        for _ in 0..count {
+            if pos + FIELD_LEN_PREFIX_LEN > data.len() {
+                return Err(WireError::InvalidFormat(
+                    "unexpected end of buffer reading payload length",
+                ));
+            }
+            let len_bytes: [u8; FIELD_LEN_PREFIX_LEN] = data[pos..pos + FIELD_LEN_PREFIX_LEN]
+                .try_into()
+                .expect("slice length checked above");
```
sitaowang1998: Same

components/spider-tdl/src/wire.rs
```diff
@@ -0,0 +1,963 @@
...
+    fn next_field_bytes(&mut self) -> Result<&'de [u8], WireError> {
+        if self.pos + FIELD_LEN_PREFIX_LEN > self.data.len() {
+            return Err(WireError::InvalidFormat(
+                "unexpected end of buffer reading payload length",
+            ));
+        }
+        let len_bytes: [u8; FIELD_LEN_PREFIX_LEN] = self.data
+            [self.pos..self.pos + FIELD_LEN_PREFIX_LEN]
+            .try_into()
+            .expect("slice length checked above");
```
sitaowang1998: Same

sitaowang1998 (commented):
Why are we having both `unframe_payloads` and `StreamDeserializer` at the same time, with overlapping functionality?

LinZhihao-723:
> Why are we having both `unframe_payloads` and `StreamDeserializer` at the same time, with overlapping functionality?

They're not overlapping.

- `unframe_payloads` is to simply unframe the payloads into a vector of byte buffers (`Vec<u8>`). This is to deserialize task outputs as payloads in the storage service.
- `StreamDeserializer` is to deserialize the buffer into a struct. Supported by `serde`, the compiler will generate the code for field access, using the field accessor we provide. Each payload will be further deserialized from msgpack into a concrete typed struct field. This is to deserialize task inputs directly into task parameters inside TDL package.

components/spider-tdl/src/wire.rs
LinZhihao-723: This is actually not reachable, so the error message is more like an internal comment (still don't want to use unwrap).
Rewrote using `first_chunk` to avoid this pattern.

sitaowang1998 (dismissed):

sitaowang1998 (dismissed):

sitaowang1998 (approved):

PR #309, opened 2026-04-15
feat(spider-storage): Add mpmc-channel-based ready queue implementation.

sitaowang1998:
# Description

This PR implements ready queue. Ready queue stores the tasks ready to run.

## Architecture
```mermaid
graph LR
  jcb[JCB]

  subgraph rq[Ready Queue]
    subgraph task[Task]
      task_q[Queue]
    end 

    subgraph commit[Commit]
      commit_q[Queue]
    end

    subgraph cleanup[Cleanup]
      cleanup_q[Queue]
    end
  end

  scheduler[Scheduler]
  
  jcb --> task_q
  jcb --> commit_q
  jcb --> cleanup_q

  task_q --> scheduler
  commit_q --> scheduler
  cleanup_q --> scheduler
```

We make the following decision when designing ready queue architecture:
- We have three separate queues: task, commit and cleanup, so that higher priority commit and cleanup task are not blocked by normal task.
- Each queue is a `mscp::Channel`.

## Misc
This PR also consolidates `strum` and `strum_macro` dependency into one using features.

# Checklist

  breaking change.

# Validation performed

- [x] New unit tests added for the ready queue.
- [x] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * In-memory ready-queue with separate task/commit/cleanup lanes, sender/receiver handles and batch receive API.

* **Improvements**
  * Core dependency upgraded to a newer release and macros consolidated.
  * Identifier and task types now support hashing for broader usage.
  * Ready-queue config validation and clearer channel-closed error variants.

* **Tests**
  * Unit tests added/updated for ready-queue batching, timing and channel-closure behaviour.


components/spider-storage/src/ready_queue.rs
LinZhihao-723, line 1: Please first fill all docstrings.

LinZhihao-723 (changes requested):

LinZhihao-723 (changes requested):
In the PR description:

> A bounded queue that periodically polls from the channel and dedup the tasks.

Based on our discussion, I'm not sure if it's efficient to do this periodically:
- I think the ready queue should be really latency sensitive. The drain interval is hard to tune.
- This violates the ideal of using a bounded buffer on both the channel and the queue: we want the bound to block things when the container is full, and automatically unblock things when the consumer end makes progress.

Shall we just implement the ready queue like a manually implemented channel that:
- Blocks the consumer and producer
- Allow resetting the entire queue (remove all elements) from another async coroutine
- Allow read from the queue with a timeout (safe cancellation)

Sth prototyped by the copilot: 

[async_dedup_queue.md](https://github.com/user-attachments/files/26791419/async_dedup_queue.md)

components/spider-storage/src/ready_queue.rs
LinZhihao-723, line 1: Let's add a mod-level doc to explain the arch of the ready queue.

components/spider-storage/src/ready_queue.rs
```diff
@@ -1,7 +1,53 @@
...
+const DEFAULT_TASK_READY_CAPACITY: usize = 65_536;
+
+/// Default resident commit-task queue capacity.
+const DEFAULT_COMMIT_READY_CAPACITY: usize = 1024;
+
+/// Default resident cleanup-task queue capacity.
+const DEFAULT_CLEANUP_READY_CAPACITY: usize = 1024;
+
+/// Default ingress queue capacity.
+const DEFAULT_INGRESS_CAPACITY: usize = DEFAULT_TASK_READY_CAPACITY;
```
LinZhihao-723: Private const members should come after public symbols (or let me know if there's a linter warning).

components/spider-storage/src/ready_queue.rs
```diff
@@ -53,3 +100,1596 @@ pub trait ReadyQueueSender: Clone + Send + Sync {
...
+impl Default for ReadyQueueConfig {
+    fn default() -> Self {
+        Self {
+            ingress_capacity: DEFAULT_INGRESS_CAPACITY,
+            task_ready_capacity: DEFAULT_TASK_READY_CAPACITY,
+            commit_ready_capacity: DEFAULT_COMMIT_READY_CAPACITY,
+            cleanup_ready_capacity: DEFAULT_CLEANUP_READY_CAPACITY,
+        }
+    }
+}
```
LinZhihao-723: The rule is to group the impl with the type definition. Try using Claude to review the style before submitting the PR, or directly feed the coding guidelines for code generation: https://github.com/LinZhihao-723/claude-instruction/tree/main/rust

components/spider-storage/src/ready_queue.rs
LinZhihao-723, line 1: Shouldn't we include resource group IDs when sending ready messages?

components/spider-storage/src/ready_queue.rs
```diff
@@ -26,7 +72,8 @@ pub trait ReadyQueueSender: Clone + Send + Sync {
         task_indices: Vec<TaskIndex>,
     ) -> Result<(), InternalError>;
 
-    /// Enqueues a signal indicating that the commit task of the given job is ready to be scheduled.
```
LinZhihao-723, line 29: Don't forget to check the diff.

components/spider-storage/src/ready_queue.rs
```diff
@@ -53,3 +100,1596 @@ pub trait ReadyQueueSender: Clone + Send + Sync {
...
+    /// # Returns
+    ///
+    /// The total number of resident entries across all queues.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * [`InternalError`] if the ready queue is corrupted or unavailable.
+    async fn total_len(&self) -> Result<usize, InternalError>;
```
LinZhihao-723: `total_len` -> `num_all`?

LinZhihao-723 (changes requested):

components/spider-storage/src/ready_queue.rs
```diff
@@ -12,44 +123,1418 @@ pub trait ReadyQueueSender: Clone + Send + Sync {
...
+const DEFAULT_COMMIT_READY_CAPACITY: usize = 1024;
+
+/// Default resident cleanup-task queue capacity.
+const DEFAULT_CLEANUP_READY_CAPACITY: usize = 1024;
+
+/// Default ingress queue capacity.
+const DEFAULT_INGRESS_CAPACITY: usize = DEFAULT_TASK_READY_CAPACITY;
+
+#[async_trait]
+impl ReadyQueueSender for ReadyQueueSenderHandle {
```
LinZhihao-723, line 155: These `impl` groups are just randomly located... Make sure you check the ordering before publishing for a review.

components/spider-storage/src/ready_queue.rs
```diff
@@ -12,44 +123,1418 @@ pub trait ReadyQueueSender: Clone + Send + Sync {
...
+    /// * `runtime_handle` - The Tokio runtime handle used to spawn the background flusher tasks.
+    ///
+    /// # Returns
+    ///
+    /// The created ready queue.
+    ///
+    /// # Errors
+    ///
+    /// * [`InternalError::ReadyQueueInvalidConfig`] if any configured capacity is zero.
+    pub fn create(
```
LinZhihao-723: Why not making it like channel APIs to let the factory function returns a tuple of (sender, reciever, resetter)?

components/spider-storage/src/ready_queue.rs
LinZhihao-723, line 1: Design issue 1: The design of `ReadyQueueInner` is pretty confusing. It owns the sender copies of all three channels and the three resident queue, as well as the abort handler of the flusher. But:
1. The sender end doesn't need a copy of the queue. The queue sender should simply take three channel senders. Nothing else.
2. The receiver and the resetter end only need to take the shared resident queues; there is no need to take the sender.
3. None of these above should hold the abort handlers. In fact, it doesn't even make sense to have the abort handler and let them be called inside the drop method. The "flush" coroutines spawned take the writer side of the queue, which is an arc of `ReadQueueInner`. That means as long as the coroutine is running, there will be a reference hold by the coroutine, and `drop` should never be triggered. At by the time `drop` is triggered, abort doesn't seem to be useful since the coroutine must already terminate.

LinZhihao-723, line 1: Design issue 2: I don't think we need to spawn 3 coroutines to drain the channel. It can be one coroutine taking three channel receivers and three queue writers to do this job.
Back to the issue 1: The abort handler shouldn't be owned by the queue itself. Instead, the storage service should hold the abort handler to gracefully cancel this coroutine before exit.

LinZhihao-723, line 1: Design issue 3: Naming. We have:
```
`ReadyQueueInner` -> `ReadyQueueSenderHandle`
                  -> `ReadyQueueReceiverHandle`
                  -> `ReadyQueueResetterHandle`
                  -> `ReadyQueue`
```
But as discussed in issue 1 and 2, there's no need to have `ReadyQueue` since the ready queue's ownership should be shared. Therefore, we should drop `ReadyQueue` and rename `ReadyQueueInner` to `ReadyQueue`.
I think generally this `Inner` suffix should be avoided. Like tcb and jcb, we should call the inner `XXX` and call the arc wrapper `SharedXXX`.

LinZhihao-723, line 1: Design issue 4: Naming again.
```rust
#[derive(Clone)]
struct ResidentQueue {
    writer: ResidentQueueWriter,
    resetter: ResidentQueueResetter,
    receiver: ResidentQueueReceiverHandleInner,
}
```
It's pretty obvious that `ResidentQueueReceiverHandleInner` is inconsistent with others when skimming the code.
I also would suggest the following renaming and redesign:
- `ResidentQueueInner` -> `ResidentQueue` (as discussed in issue 3)
- `ResidentQueueState` -> `BoundedDedupQueue` since it's weird to call the actual container "state".

LinZhihao-723, line 1: Design issue 5: There's a better way to design the resetter.

Before talking about the real issue, the style-wise is problematic: no explanation to the type parameter `I`, and the name `I` itself is also sth we should avoid. Assume `I` means something iterable, then:

The current behavior: you require the JCB table level to prepare all ready tasks in a vector or some other iterable container, and pass it to the resetter that locks the resident queue, clean it up, and replace it with the given container.

This can cause a problem since you can't actually stop the "flush" coroutine to read from the channel. Consider the following interleaving:

- Coroutine A:
  1. Iterate the JCB table and prepare the rebuilt ready tasks in a vector.
  2. Call the resetter method.
- Coroutine B:
  1. Complete a task instance and send the ready children tasks to the channel.
- Coroutine C:
  1. The "flush" coroutine that reads from the channel and enqueues the ready tasks to the resident queue.

If things happen in the order of A1, B1, C1, A2, then the ready tasks sent by B will be lost since the resetter will clean up the resident queue and replace it with the vector prepared in A1, which doesn't include the tasks sent by B.

You may solve this problem by blocking coroutine B by making JCB table access exclusive in A, but that would cause unnecessary blocking.

The way I'd approach this problem is:
1. Use the resetter to clean up the resident queue and do nothing else.
2. Walk through the JCB table and ask each JCB to send the ready tasks to the channel again.
    - This happens after 1, which means even if it cleans up the resident queue, all ready tasks before 1 will be resent to the channel and no ready task will be lost.
3. Rely on the "flush" coroutine to rebuild the resident queue based on the tasks sent to the channel.

Similar as the task instance pool, I think we should try to make the model clear here that essentially the coroutine is an actor (https://ryhl.io/blog/actors-with-tokio/), and the actor is the single source of one operation. In our case:
- All enqueue operations are all done by the "flush" coroutine.
  - Like this rebuild, we just ask JCB to resend into the channel, but the actual enqueue operation still happens inside the "flush" coroutine.
- All dequeue operations (include reset, which is essentially to dequeue all the entries in the queue) are done by the scheduler through RPC calls.

I do prefer this programming model since it's easier to reason about the concurrency while it's not creating unnecessary blocking.

LinZhihao-723 (changes requested):
Some non-trivial issues in the current implementation.
Please read my comments carefully and check if you agree. I hope we can avoid these discussions/issues in future PR (so we can really speed things up).
In the meantime, I think it might be better to let me take over the implementation refactoring, and you can focus on #309 improvement.

components/spider-storage/src/ready_queue.rs
```diff
@@ -1,55 +1,474 @@
...
+/// Returns an error if:
+///
+/// * [`InternalError::ReadyQueueChannelClosed`] if the channel is closed. In a healthy storage
+///   service, the channel should only be closed when the service is shutting down.
+async fn recv_batch<TaskKind>(
+    receiver: &Receiver<ReadyQueueEntry<TaskKind>>,
+    max_items: usize,
+    wait: Duration,
+    lane: QueueType,
+) -> Result<Vec<ReadyQueueEntry<TaskKind>>, InternalError> {
```
sitaowang1998, line 353: Since `TaskKind` and `QueueType` should be one-to-one match, is there a way we only specify one of them?

sitaowang1998 (commented):
LGTM overall.

components/spider-storage/src/ready_queue.rs
LinZhihao-723: They are not exactly the same:
- `QueueType` is like a debug marker.
- `TaskKind` needs to deliver actual task-level info: for regular tasks, it needs to store the task index.

LinZhihao-723: looking at this piece of code again. We could probably drop `QueueType`. In practice, we will log all these errors on the caller side, and the caller should have information on which queue lane it's calling for. It doesn't need to be stored in the error enum.

sitaowang1998 (commented):
LGTM. I can't approve my own PR, lol.

LinZhihao-723 (approved):
For the PR title, how about:
```
feat(spider-storage): Add mpmc-channel-based ready queue implementation.
```

PR #310, opened 2026-04-17
feat(spider-storage): Add task instance pool implementation:

sitaowang1998:
* Introduce execution manager liveness tracking traits to support task instance pool garbage collection.
* Rename all occurrences of "worker" to "execution manager" across crates to align with the architecture design.

# Description

This PR:
- Implements in-memory task instance pool for existing interface.
- Adds soft-timeout GC in task instance pool.

# Checklist

  breaking change.

# Validation performed
- [x] New unit tests added.
- [x] Scheduling test uses implemented task instance pool.
- [x] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Dedicated task-instance pool actor with registration, GC, EM-liveness checks, and per-instance metadata.
  * Public is_terminal queries for task control blocks.

* **Bug Fixes**
  * Surface new error case for corrupted task instance pool.

* **Refactor**
  * Renamed runtime concept to “execution manager”; APIs updated to accept execution-manager IDs and task-instance metadata.
  * Marker IDs updated to introduce ExecutionManagerId and derive hashing.

* **Tests**
  * Added tests for dead-manager recovery, soft-timeout behaviour and terminal cleanup.


LinZhihao-723 (changes requested):
High-level design comments before we proceed:
- I don't think we should have `unregister_running_task_instance`, and this was not a plan. It looks like you will need JCB to call this method to remove a task instance from the pool on every task instance completion. But that means on every completion, you need to lock the task instance pool for removal.
- I also don't think we need to maintain two hash sets. At least the worker to task instance hash set is not needed since the benefit is minimal: it only helps you speed up the GC cycle when you need to remove instances in a dead worker (which in the current stage of Spider, each worker only runs one instance). But you are paying both the runtime and the memory overhead for maintaining an extra hash map. I think in our GC use-case, a linear scan would be sufficient (which I will explain in details below).
- So we come back to the question of what exactly should be covered in a GC cycle? I think for now, maintaining a list of task instance records should be enough. And at the beginning of each GC cycle, we first fetch the dead workers, and then do a linear scan on the list:
  - If the corresponded TCB terminates, mark it to-be-removed.
  - If the task instance is running inside a dead worker, mark it to-be-removed and forcely remove the task instance ID from the TCB.
  - If the task instance is still alive, but it triggers the soft timeout, then re-enqueue it through the ready-queue sender.
- What's the concern with the design:
  - The in-pool operation must be serialized. We can use a lock for now. But if it becomes a problem (which is highly possible), we can do similar things as the ready queue to use a channel to collect task instances in batches, and then add them to the pool at once. This also means task instance registration won't be blocked even in the middle of a GC execution. But even without a middle-layer channel, the critical region it short enough: we're not paying the the overhead of hashing twice inside a critical section.
  - Would the linear scan be a problem? Speed-wise, I don't think so, since the GC might not be speed-sensitive anyway. Memory-usage-wise, maybe, but we can always do both timeout-triggered and pool-size-triggered GC (should be easy to implement using async Rust).
  - Memory efficiency of the list deletion? The deletion happens at the end of a GC cycle with all index-to-delete to be collected. The native way to implement this is to use a double-linked-list, but we could even use an array with swap-and-pop deletion applied from the highest index to remove to the lowest, since we don't need a stablized index for the list.

Unless you see a strong benefit of using a hash map, I think this list + linear scan version is easier to maintain and optimize, which should also fit the requirements for the  task instance pool.

LinZhihao-723:
> High-level design comments before we proceed:
> 
> * I don't think we should have `unregister_running_task_instance`, and this was not a plan. It looks like you will need JCB to call this method to remove a task instance from the pool on every task instance completion. But that means on every completion, you need to lock the task instance pool for removal.
> * I also don't think we need to maintain two hash sets. At least the worker to task instance hash set is not needed since the benefit is minimal: it only helps you speed up the GC cycle when you need to remove instances in a dead worker (which in the current stage of Spider, each worker only runs one instance). But you are paying both the runtime and the memory overhead for maintaining an extra hash map. I think in our GC use-case, a linear scan would be sufficient (which I will explain in details below).
> * So we come back to the question of what exactly should be covered in a GC cycle? I think for now, maintaining a list of task instance records should be enough. And at the beginning of each GC cycle, we first fetch the dead workers, and then do a linear scan on the list:
>   
>   * If the corresponded TCB terminates, mark it to-be-removed.
>   * If the task instance is running inside a dead worker, mark it to-be-removed and forcely remove the task instance ID from the TCB.
>   * If the task instance is still alive, but it triggers the soft timeout, then re-enqueue it through the ready-queue sender.
> * What's the concern with the design:
>   
>   * The in-pool operation must be serialized. We can use a lock for now. But if it becomes a problem (which is highly possible), we can do similar things as the ready queue to use a channel to collect task instances in batches, and then add them to the pool at once. This also means task instance registration won't be blocked even in the middle of a GC execution. But even without a middle-layer channel, the critical region it short enough: we're not paying the the overhead of hashing twice inside a critical section.
>   * Would the linear scan be a problem? Speed-wise, I don't think so, since the GC might not be speed-sensitive anyway. Memory-usage-wise, maybe, but we can always do both timeout-triggered and pool-size-triggered GC (should be easy to implement using async Rust).
>   * Memory efficiency of the list deletion? The deletion happens at the end of a GC cycle with all index-to-delete to be collected. The native way to implement this is to use a double-linked-list, but we could even use an array with swap-and-pop deletion applied from the highest index to remove to the lowest, since we don't need a stablized index for the list.
> 
> Unless you see a strong benefit of using a hash map, I think this list + linear scan version is easier to maintain and optimize, which should also fit the requirements for the task instance pool.

Actually, I realize we can indeed use a dedicated coroutine to execute both GC and task instance in-pool operation.

The arch:

JCB-0 --> sender
JCB-1 --> sender  --> channel --> recv --> task-instance-pool
JCB-2 --> sender

The task instance pool is a single coroutine that runs both enqueue and GC:

```rust
            loop {
                select! {
                    _ = tokio::time::sleep(gc_timer) => {
                        // Run GC: linear scan the list
                        // reset timer
                    }

                    Some(entry) = receiver.recv() => {
                        // Add the entry to the list
                    }
                }
            }
```

- No mutex needed on the list, since it is owned only by one coroutine.
- GC doesn't block task instance creation.
- Theoretically, you could add `unregister_task_instance` as a message variant and let the receiver handle it. But I still don't think it's necessary as explained in the previous reply.

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -1,11 +1,82 @@
...
+    db::DbError,
+    ready_queue::ReadyQueueSender,
 };
 
+/// Metadata for one running task instance tracked by the task instance pool.
+///
+/// This record carries the information needed to re-enqueue soft-timed-out work and to remove all
+/// live task instances associated with a worker during recovery.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct TaskInstanceRecord {
```
LinZhihao-723: `TaskInstanceRecord` -> `TaskInstanceMetadata`?

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -1,11 +1,82 @@
...
+    pub registered_at: SystemTime,
+    pub timeout_policy: TimeoutPolicy,
+}
+
+/// Store for tracking worker liveness state.
+///
+/// Implementations persist worker heartbeat state durably and provide an atomic operation to detect
+/// and mark dead workers for recovery.
+#[async_trait]
+pub trait WorkerLivenessStore: Clone + Send + Sync {
```
LinZhihao-723: In terms of naming: I think we should use `ExecutionManager` instead of `Worker` to match our plan. Otherwise, we need to come back and fix all the namings across different crates.

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -1,11 +1,82 @@
...
+    /// Returns the IDs of workers whose last heartbeat is before `stale_before`, after marking them
+    /// dead.
+    ///
+    /// This operation is atomic: once a worker is returned by this method, it will not be returned
+    /// again in subsequent calls.
```
LinZhihao-723: This would work, but it comes with a synchronization problem: assume in one GC cycle, we collect the IDs of all dead workers in this cycle, do some cleaning, and never touch these IDs again. But assume there's another instance still in the channel, or still in the middle of task instance registration issued from a dead worker before the GC cycle starts, this instance gets into the pool and can only be cleaned in the next cycle. In the next cycle, the previously cleaned worker ID won't be fired again, meaning that this late-registered instance will be stuck forever.
I think no matter how we design the flow, we need a way to prevent this situation. Back to the single coroutine task instance pool model I proposed, I think we can do the following:
- The task instance pool maintains a set of living worker IDs.
- In a GC cycle, the task instance pool calls this method (`get_dead_workers`) to get a list of dead worker IDs and remove them from the storage. Exclude all dead worker IDs from the living set.
  - This removal also enforces that if the dead worker comes back online, it can't update its heartbeat using its previous ID. It must restart itself to register a new worker ID to keep serving.
- When en-pooling a task instance, verify if its ID is in the living set. If not, check the storage to make sure it is a living worker ID.
  - If yes, add it to the living set so future tasks don't need to go through this additional ID validation again.
  - Otherwise, reject this task instance and do force termination (remove it from tcb).
  - Optionally, we can expose a channel for the worker registration endpoint to fast-forward messages for new worker ID registration.

Let me know your thoughts on this issue (or maybe u've already have other plans).

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -26,37 +97,906 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * [`InternalError`] if the pool is in an inconsistent state.
+    async fn drain_worker_task_instances(
+        &self,
+        worker_id: WorkerId,
+    ) -> Result<Vec<TaskInstanceRecord>, InternalError>;
```
LinZhihao-723: Why do we need to expose this method to the connector trait? This trait is used by the JCB only, and it doesn't need to execute any removal logic or dead worker detection, right?

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -26,37 +97,906 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+#[derive(Clone)]
+pub struct TaskInstancePool<
+    ReadyQueueSenderType: ReadyQueueSender,
+    WorkerLivenessStoreType: WorkerLivenessStore,
+> {
+    ready_queue_sender: ReadyQueueSenderType,
+    worker_liveness_store: WorkerLivenessStoreType,
+    worker_stale_cutoff: Duration,
+    next_task_instance_id: Arc<AtomicU64>,
+    state: Arc<Mutex<TaskInstancePoolState>>,
```
LinZhihao-723: As suggested in [this](https://github.com/y-scope/spider/pull/310#issuecomment-4271402552) comment, we only need a single coroutine to execute all logic needed by a task instance pool. We don't need a mutex-guarded shared task instance pool state: the pool itself can own all the data structure it needs to track the task instance metadata and worker info.

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -26,37 +97,906 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+    ///
+    /// * Forwards [`Self::run_gc_cycle_at`]'s return values on failure.
+    pub async fn run_gc_cycle(&self) -> Result<(), InternalError> {
+        self.run_gc_cycle_at(SystemTime::now()).await
+    }
+}
+
+/// A type-erased control block that holds either a regular or a termination TCB.
+#[derive(Clone)]
+enum AnySharedControlBlock {
```
LinZhihao-723: How about just `Tcb`? `Any_` is a confusing prefix.

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -26,37 +97,906 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+        let dead_workers = self
+            .worker_liveness_store
+            .get_dead_workers(
+                gc_started_at
+                    .checked_sub(self.worker_stale_cutoff)
+                    .unwrap_or(SystemTime::UNIX_EPOCH),
+            )
+            .await
+            .map_err(|e| InternalError::TaskInstancePoolCorrupted(e.to_string()))?;
+        let dead_worker_set: Vec<WorkerId> = dead_workers;
```
LinZhihao-723: Why `dead_worker_set` <- `dead_worker`?
- If we plan to use a set, the type is wrong.
- If we just need to rename, then we can just rename `dead_worker`?

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -26,37 +97,906 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+                        continue;
+                    };
+                    if soft_timeout_deadline <= gc_started_at {
+                        soft_timeout_entries.push(entry.record.clone());
+                    } else {
+                        live_entries.push((entry.record.clone(), entry.control_block.clone()));
+                    }
+                } else {
+                    live_entries.push((entry.record.clone(), entry.control_block.clone()));
+                }
```
LinZhihao-723: Before we do any of these checks, we should check whether TCB is already in a terminal state: it could be terminated by other workers, or already terminated by the dead worker. In that case, we just add it to the remove list, right?

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -26,37 +97,906 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+            .lock()
+            .expect("task instance pool mutex should not be poisoned");
+        if let Some(entry) = state
+            .running_task_instances
+            .iter_mut()
+            .find(|entry| entry.record.task_instance_id == task_instance_id)
+        {
+            entry.gc_processed = gc_processed;
+        }
+    }
```
LinZhihao-723: Why not setting it in-place when we make the decision to fire the soft timeout?

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -26,37 +97,906 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+        let task_instance_id = record.task_instance_id;
+        if state
+            .running_task_instances
+            .iter()
+            .any(|entry| entry.record.task_instance_id == task_instance_id)
+        {
+            return Err(InternalError::TaskInstancePoolCorrupted(format!(
+                "task instance {task_instance_id} already registered"
+            )));
+        }
```
LinZhihao-723: I don't think this is necessary: you need to traverse the entire list for every registration.
- Duplicates should be well handled by the current implementation, and prevented from the implementation of how JCB calls the registration.
- If we really need this de-duplication, we should fall back to your previous hash-based data structure.

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -26,37 +97,906 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+    }
+
+    /// Re-enqueues the task corresponding to the given running-record metadata.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * [`InternalError`] if sending the corresponding ready-queue event fails.
+    async fn re_enqueue_task(&self, record: TaskInstanceRecord) -> Result<(), InternalError> {
```
LinZhihao-723: Do we need to take the ownership of `record`?

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -26,37 +97,906 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+                    .send_cleanup_ready(record.job_id)
+                    .await
+            }
+        }
+    }
+}
+
+#[async_trait]
+impl<ReadyQueueSenderType: ReadyQueueSender, WorkerLivenessStoreType: WorkerLivenessStore>
+    TaskInstancePoolConnector for TaskInstancePool<ReadyQueueSenderType, WorkerLivenessStoreType>
```
LinZhihao-723: Our rule is to add the `impl` after the type definition. `TaskInstancePoolConnector` is not defined above this `impl`.

components/spider-storage/src/task_instance_pool.rs
LinZhihao-723, line 1: I think it makes more sense to include the coroutine implementation inside this PR, and test based on it.
The coroutine can be an async method of `TaskInstancePool` which takes the ownership of `self` which allows tokio to spawn it statically.

LinZhihao-723 (changes requested):
Only reviewed the implementation. Didn't check any tests or changes in other mods. Please read my comments carefully since they're not just style suggestions.

components/spider-storage/src/task_instance_pool.rs
sitaowang1998: Could we delay al worker renaming to a later PR?

LinZhihao-723: Why?

sitaowang1998: Because there are other mentioning of `Worker` already exists. For example, the `WorkerId` in the `id.rs`.

LinZhihao-723: That ID exists before we have planned for this entire redesign. I'd suggest creating a new ID type for EM and removing that worker ID in a dedicated PR if you want. I don't see a reason to accept this large number of uses of "worker" when we've already decided on the proper name.
Btw, this should be a one-command thing for claude.

components/spider-storage/src/task_instance_pool.rs
sitaowang1998: Does it mean gc stalls all other requests to the task instance pool?

LinZhihao-723: Yes, which is fine since even in the current implementation, the pool access is still mutually exclusive: if GC is in execution, no eu-pool operation can happen. Then it would be easier to manage if we put them all in a single coroutine/thread.

sitaowang1998:
-

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -1,11 +1,107 @@
...
+        task::{SharedTaskControlBlock, SharedTerminationTaskControlBlock},
+    },
+    db::DbError,
+    ready_queue::ReadyQueueSender,
 };
 
+/// Metadata for one running task instance tracked by the task instance pool.
+///
+/// This metadata carries the information needed to re-enqueue soft-timed-out work and to remove all
+/// live task instances associated with an execution manager during recovery.
```
LinZhihao-723: > remove all live task instances associated with an execution manager during recovery.

This comment doesn't make sense to me.

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -1,11 +1,107 @@
...
+    pub task_instance_id: TaskInstanceId,
+    pub execution_manager_id: ExecutionManagerId,
+    pub registered_at: SystemTime,
+    pub timeout_policy: TimeoutPolicy,
+}
+
+/// Store for tracking execution manager liveness state.
+///
+/// Implementations persist execution manager heartbeat state durably and provide an atomic
+/// operation to detect and mark dead execution managers for recovery.
```
LinZhihao-723: To recover an execution manager, it should register for a new ID, right?

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -1,11 +1,107 @@
...
+pub trait ExecutionManagerLivenessStore: Clone + Send + Sync {
+    /// Checks whether the execution manager with the given ID is alive.
+    ///
+    /// # Parameters
+    ///
+    /// * `id` - The execution manager ID to check.
+    ///
+    /// # Returns
+    ///
+    /// `true` if the execution manager is alive, `false` otherwise, on success.
```
LinZhihao-723:
```suggestion
    /// Whether the execution manager is alive on success.
```
Like other language, we should use "whether...".

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -24,39 +120,723 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+
+/// A handle to the task instance pool for creating and registering task instances.
+///
+/// This is a non-generic, `Clone`-able connector. All clones share the same channel and atomic ID
+/// counter, so they talk to the same underlying pool coroutine.
+#[derive(Clone)]
+pub struct TaskInstancePoolHandle {
+    next_task_instance_id: Arc<AtomicU64>,
+    sender: mpsc::Sender<PoolMessage>,
+}
```
LinZhihao-723, line 168: Again, symbol ordering. This is a public symbol and it should be defined before other private symbols.

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -24,39 +120,723 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+    /// Creates a new task instance pool and returns a handle to it.
+    ///
+    /// # Parameters
+    ///
+    /// * `ready_queue_sender` - The sender for re-enqueuing tasks to the ready queue.
+    /// * `execution_manager_liveness_store` - The store for querying execution manager liveness.
+    /// * `execution_manager_stale_cutoff` - The duration after which an execution manager with no
+    ///   heartbeat is considered stale by the pool's GC cycle.
+    /// * `gc_interval` - The interval between automatic GC cycles.
+    ///
```
LinZhihao-723: 1. We don't need parameter documentation, right?
2. The type parameters are missing.

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -24,39 +120,723 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+    /// The pool uses a bounded channel (capacity 128) between the handle and the coroutine. Because
+    /// `register_task_instance` is called while the caller holds the JCB read lock, if the
+    /// coroutine stalls (e.g., during a GC cycle or a slow liveness check), 128 pending
+    /// registrations will cause subsequent `create_task_instance` callers to block under the
+    /// read lock, potentially starving write-lock holders (`succeed_task_instance`, `cancel`).
+    /// If this becomes an issue under load, consider widening the buffer or restructuring to
+    /// avoid holding the lock during registration.
+    #[must_use]
+    pub fn new<ReadyQueue, LivenessStore>(
```
LinZhihao-723: The ready queue documentation doesn't follow our guideline:
- `ReadyQueue` -> `ReadyQueueSenderType`
- `LivenessStore` -> `LivenessStoreType`

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -24,39 +120,723 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+        ready_queue_sender: ReadyQueue,
+        execution_manager_liveness_store: LivenessStore,
+        execution_manager_stale_cutoff: Duration,
+        gc_interval: Duration,
+    ) -> Self
+    where
+        ReadyQueue: ReadyQueueSender + 'static,
+        LivenessStore: ExecutionManagerLivenessStore + 'static, {
```
LinZhihao-723: I think in general it's easier to read to put the type constrains inline:
```rust
    pub fn new<
        ReadyQueueSenderType: ReadyQueueSender + 'static,
        LivenessStoreType: ExecutionManagerLivenessStore + 'static,
    >(
        ready_queue_sender: ReadyQueueSenderType,
        execution_manager_liveness_store: LivenessStoreType,
        execution_manager_stale_cutoff: Duration,
        gc_interval: Duration,
    ) -> Self {
```

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -24,39 +120,723 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+    ///
+    /// # Backpressure
+    ///
+    /// The pool uses a bounded channel (capacity 128) between the handle and the coroutine. Because
+    /// `register_task_instance` is called while the caller holds the JCB read lock, if the
+    /// coroutine stalls (e.g., during a GC cycle or a slow liveness check), 128 pending
+    /// registrations will cause subsequent `create_task_instance` callers to block under the
+    /// read lock, potentially starving write-lock holders (`succeed_task_instance`, `cancel`).
+    /// If this becomes an issue under load, consider widening the buffer or restructuring to
+    /// avoid holding the lock during registration.
```
LinZhihao-723: Why not make 128 a configurable value?

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -24,39 +120,723 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+
+        Self {
+            next_task_instance_id,
+            sender,
+        }
+    }
+}
+
+/// The task instance pool, running as a tokio coroutine.
+///
```
LinZhihao-723, line 304: Again, type parameter doc is missing.

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -24,39 +120,723 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`ExecutionManagerLivenessStore::get_dead_execution_managers`]'s return values on
+    ///   failure.
+    /// * Forwards [`ReadyQueueSender::send_task_ready`]'s,
+    ///   [`ReadyQueueSender::send_commit_ready`]'s, or [`ReadyQueueSender::send_cleanup_ready`]'s
+    ///   return values on failure.
```
LinZhihao-723: Where does this format come from?

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -24,39 +120,723 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+                self.ready_queue_sender
+                    .send_cleanup_ready(record.job_id)
+                    .await
+            }
+        }
+    }
+}
+
+#[async_trait]
+impl TaskInstancePoolConnector for TaskInstancePoolHandle {
```
LinZhihao-723: Why `impl TaskInstancePoolHandle` doesn't follow `TaskInstancePoolHandle`?

LinZhihao-723 (changes requested):
I will push the changes directly to the commit. However, these are very sloppy mistakes that I've caught more than once.
I'd strongly recommend you to go through the coding guideline I shared manually, and go through the diff file by checking the style against our guideline. Many of these are very clearly documented in the doc. All I do is remember these rules and check them through the diff, and I think you should make sure you follow this procedure properly for every review request.

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -24,39 +120,723 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+                        // TCB still owns the instance; a subsequent GC cycle will retry.
+                    } else {
+                        control_block
+                            .force_remove_task_instance(record.task_instance_id)
+                            .await;
+                    }
+                }
+            }
+        }
+        true
```
LinZhihao-723: Involved some var renaming, but please check this rewriting which should be a cleaner version that achieves the same goal:

```suggestion
        match message {
            PoolMessage::Register { tcb, metadata } => {
                let em_id = &metadata.execution_manager_id;
                if !self.execution_manager_pool.contains(em_id) {
                    if !self
                        .execution_manager_liveness_store
                        .is_execution_manager_alive(em_id)
                        .await?
                    {
                        if tcb
                            .force_remove_task_instance(metadata.task_instance_id)
                            .await
                        {
                            self.re_enqueue_task(&metadata).await?;
                        }
                        return Ok(());
                    }
                    self.execution_manager_pool.insert(em_id.clone());
                }
                self.instances.push(PoolEntry {
                    metadata,
                    tcb,
                    soft_timeout_handled: false,
                });
                Ok(())
            }
        }
```

components/spider-storage/tests/scheduling_infra.rs
LinZhihao-723, line 1: Why do we need to update this infra? We shouldn't need the instance pool in this test infra, right?

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -1,11 +1,107 @@
...
+/// This metadata carries the information needed to re-enqueue soft-timed-out work and to remove all
+/// live task instances associated with an execution manager during recovery.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct TaskInstanceMetadata {
+    pub job_id: JobId,
+    pub task_id: TaskId,
+    pub task_instance_id: TaskInstanceId,
+    pub execution_manager_id: ExecutionManagerId,
+    pub registered_at: SystemTime,
+    pub timeout_policy: TimeoutPolicy,
```
LinZhihao-723: Kind of realize we recompute the soft timeout ddl in every GC cycle...
We should probably just store the soft timeout ddl.

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -24,39 +120,723 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+            .into_iter()
+            .map(|(record, _)| record.task_instance_id)
+            .chain(dead_em_ids_to_remove)
+            .chain(terminated_ids)
+            .collect();
+        self.state
+            .running_task_instances
+            .retain(|entry| !ids_to_remove.contains(&entry.record.task_instance_id));
+
+        Ok(())
```
LinZhihao-723:
```suggestion
        let dead_em_ids: Vec<ExecutionManagerId> = self
            .execution_manager_liveness_store
            .get_dead_execution_managers(
                gc_started_at
                    .checked_sub(self.execution_manager_stale_cutoff)
                    .unwrap_or(SystemTime::UNIX_EPOCH),
            )
            .await?;

        for execution_manager_id in &dead_em_ids {
            self.execution_manager_pool.remove(execution_manager_id);
        }

        let mut indices_to_remove: Vec<usize> = Vec::new();
        let mut indices_to_re_enqueue: Vec<usize> = Vec::new();
        for (idx, entry) in self.instances.iter_mut().enumerate() {
            let tcb = &entry.tcb;
            if tcb.is_terminal().await {
                indices_to_remove.push(idx);
                continue;
            }

            if !self
                .execution_manager_pool
                .contains(&entry.metadata.execution_manager_id)
            {
                // EM no longer lives, force-remove from TCB.
                if tcb
                    .force_remove_task_instance(entry.metadata.task_instance_id)
                    .await
                {
                    indices_to_re_enqueue.push(idx);
                }
                indices_to_remove.push(idx);
                continue;
            }

            if !entry.soft_timeout_handled
                && let Some(soft_timeout_ddl) = entry.metadata.soft_timeout_ddl
                && soft_timeout_ddl <= gc_started_at
            {
                entry.soft_timeout_handled = true;
                indices_to_re_enqueue.push(idx);
            }
        }

        for entry_to_re_enqueue in indices_to_re_enqueue
            .into_iter()
            .map(|idx| self.instances.get(idx))
            .filter(Option::is_some)
            .map(Option::unwrap)
        {
            self.re_enqueue_task(&entry_to_re_enqueue.metadata).await?;
        }

        // Remove entries. `indices_to_remove` is in sorted ascending order. Reversely iterate to
        // apply `swap_remove` is the most efficient way to remove entries in the current
        // implementation.
        for idx_to_remove in indices_to_remove.into_iter().rev() {
            self.instances.swap_remove(idx_to_remove);
        }

        Ok(())
```

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -24,39 +120,723 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+        }
+    }
+}
+
+/// Messages sent to the task instance pool coroutine.
+enum PoolMessage {
+    /// Register a task instance.
+    Register {
+        control_block: Tcb,
+        record: TaskInstanceMetadata,
```
LinZhihao-723: Renaming the type to `TaskInstanceMetadata` indicates we should also rename `record` to `metadata`. The code still uses `record` across the entire file.

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -24,39 +120,723 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+        async fn get_dead_execution_managers(
+            &self,
+            _stale_before: SystemTime,
+        ) -> Result<Vec<ExecutionManagerId>, DbError> {
+            Ok(Vec::new())
+        }
+    }
+
+    #[tokio::test]
+    async fn dead_execution_manager_registration_triggers_recovery() {
```
LinZhihao-723, line 752: The test coverage is low. It doesn't test GC cycles. As a stand-alone component, we should make sure the test case fully covers these basic functionalities and behaviors.

LinZhihao-723 (changes requested):
I'm done with polishing `task_instance_pool.rs`. I still need to:
- Add test coverage for task instance pool GC operations, which I need to figure out how to probe the internal pool inside the test case.
- I need to revert some changes in `scheduling_infra` as we shouldn't need the full task instance pool implementation in this mock setup.

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -24,39 +115,651 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+            instances: Vec::new(),
+            execution_manager_pool: HashSet::new(),
+            receiver,
+        };
+        tokio::spawn(async move {
+            match pool.run(gc_interval).await {
+                Ok(()) => {}
+                Err(_e) => todo!("log this error and terminate the storage service"),
+            }
+        });
```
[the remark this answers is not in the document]
LinZhihao-723: This is expected.

components/spider-storage/src/task_instance_pool.rs
```diff
@@ -24,39 +115,651 @@ pub trait TaskInstancePoolConnector: Clone + Send + Sync {
...
+                        .is_execution_manager_alive(em_id)
+                        .await?
+                {
+                    if tcb
+                        .force_remove_task_instance(metadata.task_instance_id)
+                        .await
+                    {
+                        self.re_enqueue_task(&metadata).await?;
+                    }
+                    return Ok(());
```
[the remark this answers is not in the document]
LinZhihao-723: This is expected.

components/spider-storage/src/cache/job.rs
```diff
@@ -154,87 +157,150 @@ impl<
     /// * Forwards [`SharedTaskControlBlock::register_task_instance`]'s return values on failure.
     /// * Forwards [`SharedTerminationTaskControlBlock::register_task_instance`]'s return values on
     ///   failure.
-    /// * Forwards [`TaskInstancePoolConnector::register_task_instance`]'s return values on failure.
-    /// * Forwards [`TaskInstancePoolConnector::register_termination_task_instance`]'s return values
-    ///   on failure.
```
LinZhihao-723, line 159: Obviously outdated. Please check the diff.

components/spider-storage/src/cache/job.rs
```diff
@@ -154,87 +157,150 @@ impl<
...
-                    .ok_or(InternalError::UndefinedCommitTask)?;
-                let task_instance_id = job
-                    .task_instance_pool_connector
-                    .get_next_available_task_instance_id();
-                let (tdl_context, timeout_policy) =
-                    commit_tcb.register_task_instance(task_instance_id).await?;
-                job.task_instance_pool_connector
-                    .register_termination_task_instance(task_instance_id, commit_tcb)
-                    .await?;
+    async fn create_regular_task_instance(
```
LinZhihao-723: No docstring.

LinZhihao-723 (changes requested):

components/spider-storage/src/cache/job.rs
LinZhihao-723: And these methods are private as well, which should be located after public ones.

LinZhihao-723 (commented):
I'm checking off with all my edits. Please go through my changes carefully.
Again, this is not a research prototype. If you look at how much work I've re-done in these polish commits, I hope you really see how unacceptable it is to request a review on its previous shape. Imagine if I didn't code on my own, how much time, effort, and communication overhead would we need to spend more? We should have an offline chat to decide in detail how to avoid this situation, and we should both agree, as part of the official comment to sign off on this PR, that this should be the **LAST** time for this type of conversation. Reply thumb up to confirm you have read this. Thanks.

LinZhihao-723 (approved):
Updated PR title (made it a multi-line commit message).

PR #311, opened 2026-04-18
feat(spider-tdl): Add task trait and type-erased task handler.

LinZhihao-723:
# Description

This PR is the third PR in series #302.

This PR:

* Add `Task` trait — the contract the `#[task]` proc-macro will implement for every user-defined task function (`NAME`, `Params`, `Return`, `execute()`, and a default `serialize_return()`).
* Add `TaskHandler` trait — type-erased dispatch interface used by the registration macro's dispatch table.
* Add `TaskHandlerImpl<TaskType: Task>` — bridges a concrete `Task` impl to `TaskHandler` by deserializing the context and parameters, invoking the task, and serializing the return tuple (or the resulting `TdlError`) back into bytes.
* Add `ExecutionResult` — a `Result<Vec<u8>, Vec<u8>>` type alias where `Ok` carries wire-format-encoded outputs and `Err` carries a msgpack-encoded `TdlError`.

A few design notes worth surfacing for reviewers:

* **`Task::Return` is required to be a tuple, but the constraint is runtime-only.** Rust's `std::marker::Tuple` is nightly-only, so the bound on the trait is just `serde::Serialize`. The tuple check lives inside `TaskOutputsSerializer::from_tuple`, which is driven through a custom serde `Serializer` that only accepts `serialize_tuple` / `serialize_unit`. A hand-written `type Return = SomeStruct` compiles, but `execute_raw` returns `TdlError::SerializationError` at runtime. The future `#[task]` proc-macro will also validate the return shape at parse time, so in practice the runtime check is a second line of defense rather than the primary guardrail. A sealed-trait workaround (`TaskReturn` implemented only for tuple arities) was prototyped but dropped to keep the trait ergonomics simple and to avoid a hard-coded arity ceiling.
* **`TaskHandler: Send + Sync`.** The registration macro in a future PR will store handlers in a `static LazyLock<HashMap<&str, Box<dyn TaskHandler>>>` shared across threads. `TaskHandlerImpl<TaskType>` only holds `PhantomData<TaskType>`, so the bound holds automatically for any `TaskType: Task + Send + Sync`.
* **`ExecutionResult` is a plain `Result` alias rather than a dedicated enum.** Both arms are already `Vec<u8>` and the distinction (wire-encoded outputs vs. msgpack-encoded `TdlError`) maps cleanly onto `Result::Ok` / `Result::Err`. Using an alias keeps the dispatch site small and lets callers use `?` inside `execute_raw`.

# Checklist

  breaking change.

# Validation performed

* [x] Add unit tests covering the full dispatch pipeline:
  * success path with a two-argument `Add`,
  * task-returned error (`TdlError::ExecutionError`) via `AlwaysFail`,
  * malformed `TaskContext` bytes,
  * malformed task-input wire buffer,
  * illegal non-tuple `Return` (struct-typed return) -- confirms the runtime `TdlError::SerializationError` fires.

## Release Notes

* **New Features**
  * Introduced a new task execution framework enabling developers to implement custom, type-safe task handlers
  * Automatic serialization and deserialization of task parameters and execution results
  * Enhanced output serialization with improved support for tuple-based result formats
  * Expanded the library's public API with task-related modules and utility types


components/spider-tdl/src/task.rs
```diff
@@ -0,0 +1,363 @@
...
+
+    /// The return type of the task (must be a tuple, even for single values).
+    ///
+    /// NOTE: Unfortunately, the tuple trait is a nightly-only feature. Otherwise, we should enforce
+    /// the type to implement [`std::marker::Tuple`]. For now, the type error can only be reported
+    /// at runtime.
+    type Return: serde::Serialize;
+
+    /// The name of the task, used as the lookup key in the registration table.
+    const NAME: &'static str;
```
LinZhihao-723, line 42: Ideally we should order constants before types, but the rust formatter seems to order types before consts inside a trait impl. To make the ordering consistent inside the trait, we move the const below type alias.

sitaowang1998 (approved):

PR #312, opened 2026-04-23
feat(spider-storage): Add session management trait and MariaDB implementation.

sitaowang1998:
# Description

This PR adds interface and MariaDB implementation of session management storage.

The session management storage provides a session id, which bumps everytime a new database connection is created. This session id is to invalidate all previous session requests on storage restart.

# Checklist

  breaking change.

# Validation performed

- [x] New unit tests added for the session management.
- [ ] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Storage connectors now expose monotonic session IDs that increment across reconnects, helping detect and reject stale requests.
  * Session management has been added to the core storage connector interface.
  * A new SessionId identifier type is available for callers.

* **Tests**
  * Added integration tests confirming session ID presence after connect and that subsequent reconnects yield higher session IDs.

### DIFF
--- a/components/spider-storage/src/db/mariadb.rs
+++ b/components/spider-storage/src/db/mariadb.rs
@@ -441,13 +441,13 @@ impl ResourceGroupManagement for MariaDbStorageConnector {
     }
 }
 
-impl DbStorage for MariaDbStorageConnector {}
-
 impl SessionManagement for MariaDbStorageConnector {
     fn session_id(&self) -> SessionId {
         self.session_id
     }
 }
+
+impl DbStorage for MariaDbStorageConnector {}
 
 /// `MySQL` error number for foreign key constraint violation.
 const MYSQL_ER_FK_CONSTRAINT: u16 = 1452;
--- a/components/spider-storage/src/db/mariadb.rs
+++ b/components/spider-storage/src/db/mariadb.rs
@@ -19,6 +19,7 @@ use crate::{
         ExternalJobOrchestration,
         InternalJobOrchestration,
         ResourceGroupManagement,
+        SessionManagement,
         error::ExpectedStates,
     },
 };
@@ -27,6 +28,7 @@ use crate::{
 #[derive(Clone)]
 pub struct MariaDbStorageConnector {
     pool: MySqlPool,
+    session_id: u64,
 }
 
 impl MariaDbStorageConnector {
@@ -59,19 +61,20 @@ impl MariaDbStorageConnector {
             .connect_with(mysql_options)
             .await?;
 
-        let connector = Self { pool };
-
         // MariaDB does not support transactions for DDL statements. All DDL statements are
         // automatically committed. Thus, each table creation query is executed separately, and
         // atomicity is not guaranteed.
         sqlx::query(resource_groups_creation_query())
-            .execute(&connector.pool)
+            .execute(&pool)
             .await?;
-        sqlx::query(jobs_creation_query())
-            .execute(&connector.pool)
+        sqlx::query(jobs_creation_query()).execute(&pool).await?;
+        sqlx::query(sessions_creation_query())
+            .execute(&pool)
             .await?;
 
-        Ok(connector)
+        let session_id = bump_session_id(&pool).await?;
+
+        Ok(Self { pool, session_id })
     }
 }
 
@@ -432,6 +435,12 @@ impl ResourceGroupManagement for MariaDbStorageConnector {
 
 impl DbStorage for MariaDbStorageConnector {}
 
+impl SessionManagement for MariaDbStorageConnector {
+    fn session_id(&self) -> u64 {
+        self.session_id
+    }
+}
+
 /// `MySQL` error number for foreign key constraint violation.
 const MYSQL_ER_FK_CONSTRAINT: u16 = 1452;
 
@@ -440,6 +449,7 @@ const MYSQL_ER_DUP_ENTRY: u16 = 1062;
 
 const RESOURCE_GROUPS_TABLE_NAME: &str = "resource_groups";
 const JOBS_TABLE_NAME: &str = "jobs";
+const SESSIONS_TABLE_NAME: &str = "sessions";
 
 const UPDATE_JOB_STATE: &str = formatcp!(
     "UPDATE `{table}` SET `state` = ? WHERE `id` = ?;",
@@ -484,6 +494,17 @@ CREATE TABLE IF NOT EXISTS `{JOBS_TABLE_NAME}` (
 );",
         state_enum = JobState::as_mysql_enum_decl(),
         default_state = JobState::Ready.as_quoted_str(),
+    )
+}
+
+#[must_use]
+const fn sessions_creation_query() -> &'static str {
+    formatcp!(
+        r"
+CREATE TABLE IF NOT EXISTS `{SESSIONS_TABLE_NAME}` (
+  `session_id` BIGINT UNSIGNED NOT NULL AUTO_INCREMENT,
+  PRIMARY KEY (`session_id`)
+);"
     )
 }
 
@@ -561,4 +582,26 @@ async fn transition_job_state(
     .await?;
 
     Ok(())
+}
+
+/// Bumps the session ID by inserting a new row into the [`SESSIONS_TABLE_NAME`] table.
+///
+/// The table uses `AUTO_INCREMENT` on `session_id`, so each `INSERT` generates the next
+/// monotonically increasing session ID.
+///
+/// # Returns
+///
+/// The new session ID on success.
+///
+/// # Errors
+///
+/// Forwards [`sqlx::query_scalar::QueryScalar::fetch_one`]'s return values on failure.
+async fn bump_session_id(pool: &MySqlPool) -> Result<u64, DbError> {
+    const QUERY: &str = formatcp!(
+        "INSERT INTO `{table}` () VALUES () RETURNING `session_id`;",
+        table = SESSIONS_TABLE_NAME,
+    );
+
+    let session_id = sqlx::query_scalar::<_, u64>(QUERY).fetch_one(pool).await?;
+    Ok(session_id)
 }
--- a/components/spider-storage/src/db/mariadb.rs
+++ b/components/spider-storage/src/db/mariadb.rs
@@ -48,7 +48,7 @@ impl MariaDbStorageConnector {
     ///
     /// * Forwards [`sqlx::mysql::MySqlPoolOptions::connect`]'s return values on failure.
     /// * Forwards [`sqlx::query::Query::execute`]'s return values on failure.
-    /// * Forwards [`bump_session_id`]'s return value on failure.
+    /// * Forwards [`bump_session_id`]'s return values on failure.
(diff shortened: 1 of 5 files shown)


components/spider-storage/src/db/mariadb.rs
```diff
@@ -487,6 +497,17 @@ CREATE TABLE IF NOT EXISTS `{JOBS_TABLE_NAME}` (
...
+#[must_use]
+const fn sessions_creation_query() -> &'static str {
+    formatcp!(
+        r"
+CREATE TABLE IF NOT EXISTS `{SESSIONS_TABLE_NAME}` (
+  `session_id` BIGINT UNSIGNED NOT NULL AUTO_INCREMENT,
+  PRIMARY KEY (`session_id`)
+);"
+    )
+}
```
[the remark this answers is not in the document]
sitaowang1998: @LinZhihao-723 What do you think about this?
1. The first one works, but need to take care of row creation if none exists.
2. I don't think we need `create_at` and periodical gc. After creating a new row we can delete the ones immediately.
3. `CREATE SEQUENCE` is supported by MariaDB and PostgreSQL, but not MySQL.

LinZhihao-723 (dismissed):
> Review generated by Claude Code using the Rust PR review skill at
> https://github.com/LinZhihao-723/claude-instruction/blob/main/rust/pr-review.md

Findings below are scoped to docstring/style rules from `coding-style.md` and the skill above.
The architecture discussion around the unbounded `sessions` table / AUTO_INCREMENT persistence
is already underway with @coderabbitai and is out of scope for this review.

### 1. `SessionManagement::session_id` is missing the short description (`db/protocol.rs`)

```rust
pub trait SessionManagement {
    /// # Returns
    ///
    /// The current session ID.
    fn session_id(&self) -> u64;
}
```

The docstring template in `coding-style.md` requires a one-line description before any section.
Either add the description, or — since this is a simple getter — drop `# Returns` and use a
single line:

```suggestion
pub trait SessionManagement {
    /// Returns the current session ID.
    fn session_id(&self) -> u64;
}
```

### 2. `bump_session_id` cites the wrong sqlx item path (`db/mariadb.rs`)

```rust
/// * Forwards [`sqlx::query_scalar::QueryScalar::fetch_one`]'s return values on failure.
```

`query_scalar` is a function, not a module. The type lives at `sqlx::query::QueryScalar`. The
rest of this file already follows the `sqlx::query::*` convention (e.g.,
`sqlx::query::Query::execute`, `sqlx::query::Query::fetch_optional`). Please change to:

```suggestion
/// * Forwards [`sqlx::query::QueryScalar::fetch_one`]'s return values on failure.
```

The skill's "name the EXACT function" rule was the most-recurring issue from #310 — please
verify the path against the actual sqlx item before citing.

### 3. `connect`'s `# Errors` is now stale (`db/mariadb.rs`)

`connect` now also calls `bump_session_id`, which produces `fetch_one` errors that
`Query::execute` does not cover. Per the skill, when a function delegates to a new helper,
the `# Errors` should forward that helper rather than re-list leaf errors. Add:

```
/// * Forwards [`bump_session_id`]'s return values on failure.
```

### 4. `sessions_creation_query` is missing a docstring (`db/mariadb.rs`)

The skill explicitly requires docstrings on non-public functions. The neighbouring
`jobs_creation_query` / `resource_groups_creation_query` are also undocumented on `main`, so
this matches existing convention — but the convention itself is wrong per the rule. Add a
one-liner:

```rust
/// Returns the DDL query to create the sessions table.
#[must_use]
const fn sessions_creation_query() -> &'static str { ... }
```

(Backfilling docstrings on the other two query helpers can be a follow-up; please don't
expand scope here.)

### 5. Test imports should be hoisted to module scope (`tests/mariadb_test.rs`)

```rust
async fn test_session_id_returned_after_connect() {
    use spider_storage::db::SessionManagement;
    ...
}

async fn test_session_id_bumps_on_reconnect() {
    use spider_storage::db::SessionManagement;
    ...
}
```

Per the skill: function-scope `use` is reserved for imports local to a single function. Since
both new tests need it, hoist into the existing `use spider_storage::db::{...};` block at
the top of the file.

---

PR title `feat(spider-storage): Add session management storage.` conforms to the convention.
Once the items above are addressed, this should be ready to land (modulo the architecture
discussion threads).

components/spider-storage/src/db/mariadb.rs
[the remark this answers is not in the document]
LinZhihao-723: I think it's fine to leave it unbounded for now. Compared to the number of job records, the number of sessions should be fairly manageable.

LinZhihao-723 (changes requested):
> **Auto-generated review** by Claude Code using the Rust PR review skill at
> https://github.com/LinZhihao-723/claude-instruction/blob/main/rust/pr-review.md.
>
> **This review supersedes my earlier review ([#4177510678](https://github.com/y-scope/spider/pull/312#pullrequestreview-4177510678), now dismissed).** The earlier item #1 — claiming the `SessionManagement::session_id` docstring was "missing the short description" — was incorrect. For a trivial getter, the one-line description may be omitted while the `# Returns` section is kept; the original docstring is already in the preferred shape per `coding-style.md`. The skill has been updated to reflect this rule clearly.

Findings below are scoped to docstring/style rules from `coding-style.md` and the skill above.
The architecture discussion around the unbounded `sessions` table / AUTO_INCREMENT persistence
is already underway with @coderabbitai and is out of scope for this review.

### 1. `bump_session_id` cites the wrong sqlx item path (`db/mariadb.rs`)

```rust
/// * Forwards [`sqlx::query_scalar::QueryScalar::fetch_one`]'s return values on failure.
```

`query_scalar` is a function, not a module. The type lives at `sqlx::query::QueryScalar`. The
rest of this file already follows the `sqlx::query::*` convention (e.g.,
`sqlx::query::Query::execute`, `sqlx::query::Query::fetch_optional`). Please change to:

```suggestion
/// * Forwards [`sqlx::query::QueryScalar::fetch_one`]'s return values on failure.
```

The skill's "name the EXACT function" rule was the most-recurring issue from #310 — please
verify the path against the actual sqlx item before citing.

### 2. `connect`'s `# Errors` is now stale (`db/mariadb.rs`)

`connect` now also calls `bump_session_id`, which produces `fetch_one` errors that
`Query::execute` does not cover. Per the skill, when a function delegates to a new helper,
the `# Errors` should forward that helper rather than re-list leaf errors. Add:

```
/// * Forwards [`bump_session_id`]'s return values on failure.
```

### 3. `sessions_creation_query` is missing a docstring (`db/mariadb.rs`)

The skill explicitly requires docstrings on non-public functions. The neighbouring
`jobs_creation_query` / `resource_groups_creation_query` are also undocumented on `main`, so
this matches existing convention — but the convention itself is wrong per the rule. Add a
one-liner:

```rust
/// Returns the DDL query to create the sessions table.
#[must_use]
const fn sessions_creation_query() -> &'static str { ... }
```

(Backfilling docstrings on the other two query helpers can be a follow-up; please don't
expand scope here.)

### 4. Test imports should be hoisted to module scope (`tests/mariadb_test.rs`)

```rust
async fn test_session_id_returned_after_connect() {
    use spider_storage::db::SessionManagement;
    ...
}

async fn test_session_id_bumps_on_reconnect() {
    use spider_storage::db::SessionManagement;
    ...
}
```

Per the skill: function-scope `use` is reserved for imports local to a single function. Since
both new tests need it, hoist into the existing `use spider_storage::db::{...};` block at
the top of the file.

---

PR title `feat(spider-storage): Add session management storage.` conforms to the convention.
Once the items above are addressed, this should be ready to land (modulo the architecture
discussion threads).

components/spider-storage/src/db/mariadb.rs
```diff
@@ -27,6 +28,7 @@ use crate::{
 #[derive(Clone)]
 pub struct MariaDbStorageConnector {
     pool: MySqlPool,
+    session_id: u64,
```
LinZhihao-723: I think we should create a type alias for the session ID.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -562,3 +584,25 @@ async fn transition_job_state(
...
+/// Forwards [`sqlx::query::QueryScalar::fetch_one`]'s return values on failure.
+async fn bump_session_id(pool: &MySqlPool) -> Result<u64, DbError> {
+    const QUERY: &str = formatcp!(
+        "INSERT INTO `{table}` () VALUES () RETURNING `session_id`;",
+        table = SESSIONS_TABLE_NAME,
+    );
+
+    let session_id = sqlx::query_scalar::<_, u64>(QUERY).fetch_one(pool).await?;
+    Ok(session_id)
+}
```
LinZhihao-723: 1. I think it would be better to inline this method into the factory, so that it's secured that no other methods in this mod can accidentally call this function to bump a session ID.
2. If you prefer to keep it as a separate method, we should probably rename it to `register_next_session_id`.

components/spider-storage/src/db/protocol.rs
```diff
@@ -312,3 +318,14 @@ pub trait ResourceGroupManagement {
     /// * Forwards [`sqlx::error::Error`] on DB operation failure.
     async fn delete(&self, resource_group_id: ResourceGroupId) -> Result<(), DbError>;
 }
+
+/// Defines the storage interface for session management.
+///
+/// A session ID is a monotonically increasing value that bumps each time the storage layer
+/// reconnects. Callers can use it to detect and reject requests from before a restart.
```
```suggestion
/// reconnects. Callers can use it to detect and reject stale requests from previous sessions.
```

components/spider-storage/src/db/mariadb.rs
```diff
@@ -487,6 +498,17 @@ CREATE TABLE IF NOT EXISTS `{JOBS_TABLE_NAME}` (
     )
 }
 
+#[must_use]
+const fn sessions_creation_query() -> &'static str {
```
LinZhihao-723, line 509: Why do we need this query as a function instead of a constant like other queries? And this is only used inside the connection factory, so it doesn't need to be a global one, right?

LinZhihao-723 (changes requested):

components/spider-storage/src/db/mariadb.rs
LinZhihao-723: nvm, this is a table creation query.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -432,6 +443,12 @@ impl ResourceGroupManagement for MariaDbStorageConnector {
 
 impl DbStorage for MariaDbStorageConnector {}
 
+impl SessionManagement for MariaDbStorageConnector {
+    fn session_id(&self) -> SessionId {
+        self.session_id
+    }
+}
+
```
LinZhihao-723, line 449: We should probably move this above `impl DbStorage` to group all meaningful trait impl.

LinZhihao-723 (changes requested):
One more nit catch otherwise we're good.

LinZhihao-723 (approved):
For the PR title, how about:
```
feat(spider-storage): Add session management trait and MariaDB implementation.
```

PR #314, opened 2026-04-27
feat(spider-storage): Add MariaDB implementation for execution manager liveness tracking.

sitaowang1998:
# Description

This PR adds execution manager liveness in the database storage. The liveness storage supports:
- Registers a new execution manager.
- Updates an existing execution manager heartbeat.
- Checking liveness of an execution manager.
- Getting all execution managers that just died.

This PR also replaces `serial_test::serial` with `serial_test::file_serial`. `nextest` uses different 
processes to run different tests. `serial_test::serial`'s in-memory lock mechanism does not provide
serialization for `nextest`, while `serial_test::file_serial` serializes tests across processes.

# Checklist

  breaking change.

# Validation performed

- [x] Add new unit tests.
- [x] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Execution manager liveness management: register managers, heartbeat updates, alive/dead queries, and automatic detection/marking of stale managers (MariaDB-backed).

* **Chores**
  * Local dependency and test configuration adjustments for file-lock based test runs.
  * Added clearer execution-manager state error variants.

* **Tests**
  * Expanded DB-layer tests for registration, heartbeats, alive checks, stale detection and atomic mark-and-skip behaviour.

### DIFF
--- a/components/spider-storage/src/db/mariadb.rs
+++ b/components/spider-storage/src/db/mariadb.rs
@@ -447,29 +447,18 @@ enum ExecutionManagerState {
 impl ExecutionManagerLivenessManagement for MariaDbStorageConnector {
     async fn register_execution_manager(
         &self,
-        execution_manager_id: ExecutionManagerId,
         ip_address: IpAddr,
-    ) -> Result<(), DbError> {
+    ) -> Result<ExecutionManagerId, DbError> {
         const INSERT_QUERY: &str = formatcp!(
-            "INSERT INTO `{table}` (`id`, `ip_address`) VALUES (?, ?);",
+            "INSERT INTO `{table}` (`ip_address`) VALUES (?) RETURNING CAST(`id` AS BINARY(16));",
             table = EXECUTION_MANAGERS_TABLE_NAME,
         );
 
-        sqlx::query(INSERT_QUERY)
-            .bind(execution_manager_id)
+        sqlx::query_scalar(INSERT_QUERY)
             .bind(ip_address.to_string())
-            .execute(&self.pool)
+            .fetch_one(&self.pool)
             .await
-            .map(|_| ())
-            .map_err(|e| match e {
-                sqlx::Error::Database(e)
-                    if e.try_downcast_ref::<MySqlDatabaseError>()
-                        .is_some_and(|mysql_err| mysql_err.number() == MYSQL_ER_DUP_ENTRY) =>
-                {
-                    DbError::ExecutionManagerAlreadyExists(execution_manager_id)
-                }
-                e => e.into(),
-            })
+            .map_err(Into::into)
     }
 
     async fn update_execution_manager_heartbeat(
@@ -555,9 +544,8 @@ impl ExecutionManagerLivenessManagement for MariaDbStorageConnector {
                 .collect::<Vec<_>>()
                 .join(",");
             let update_query = format!(
-                "UPDATE `{EXECUTION_MANAGERS_TABLE_NAME}` SET \
-                 `state` = '{dead_state}', `death_confirmed_at` = CURRENT_TIMESTAMP WHERE `id` IN \
-                 ({placeholders})",
+                "UPDATE `{EXECUTION_MANAGERS_TABLE_NAME}` SET `state` = '{dead_state}', \
+                 `death_confirmed_at` = CURRENT_TIMESTAMP WHERE `id` IN ({placeholders})",
                 dead_state = ExecutionManagerState::Dead.as_str(),
             );
             let mut query = sqlx::query(&update_query);
@@ -635,7 +623,7 @@ const fn execution_managers_creation_query() -> &'static str {
     formatcp!(
         r"
 CREATE TABLE IF NOT EXISTS `{EXECUTION_MANAGERS_TABLE_NAME}` (
-  `id` UUID NOT NULL,
+  `id` UUID NOT NULL DEFAULT UUID_v7(),
   `ip_address` VARCHAR(45) NOT NULL,
   `state` {state_enum} NOT NULL DEFAULT {default_state},
   `last_heartbeat_at` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
--- a/components/spider-storage/src/db/mariadb.rs
+++ b/components/spider-storage/src/db/mariadb.rs
@@ -476,38 +476,36 @@ impl ExecutionManagerLivenessManagement for MariaDbStorageConnector {
         &self,
         execution_manager_id: ExecutionManagerId,
     ) -> Result<(), DbError> {
+        const SELECT_STATE_FOR_UPDATE_QUERY: &str = formatcp!(
+            "SELECT `state` FROM `{table}` WHERE `id` = ? FOR UPDATE;",
+            table = EXECUTION_MANAGERS_TABLE_NAME,
+        );
         const UPDATE_QUERY: &str = formatcp!(
             "UPDATE `{table}` SET `last_heartbeat_at` = CURRENT_TIMESTAMP WHERE `id` = ? AND \
              `state` = '{alive_state}';",
             table = EXECUTION_MANAGERS_TABLE_NAME,
             alive_state = ExecutionManagerState::Alive.as_str(),
         );
-        const SELECT_STATE_QUERY: &str = formatcp!(
-            "SELECT `state` FROM `{table}` WHERE `id` = ?;",
-            table = EXECUTION_MANAGERS_TABLE_NAME,
-        );
 
-        let rows_affected = sqlx::query(UPDATE_QUERY)
-            .bind(execution_manager_id)
-            .execute(&self.pool)
-            .await?
-            .rows_affected();
-        if rows_affected != 0 {
-            return Ok(());
-        }
+        let mut tx = self.pool.begin().await?;
 
-        let state = sqlx::query_scalar::<_, ExecutionManagerState>(SELECT_STATE_QUERY)
+        let state = sqlx::query_scalar::<_, ExecutionManagerState>(SELECT_STATE_FOR_UPDATE_QUERY)
             .bind(execution_manager_id)
-            .fetch_optional(&self.pool)
+            .fetch_optional(&mut *tx)
             .await?
             .ok_or(DbError::ExecutionManagerNotFound(execution_manager_id))?;
 
-        match state {
-            ExecutionManagerState::Alive => Ok(()),
-            ExecutionManagerState::Dead => {
-                Err(DbError::ExecutionManagerAlreadyDead(execution_manager_id))
-            }
+        if state == ExecutionManagerState::Dead {
+            return Err(DbError::ExecutionManagerAlreadyDead(execution_manager_id));
         }
+
+        sqlx::query(UPDATE_QUERY)
+            .bind(execution_manager_id)
+            .execute(&mut *tx)
+            .await?;
+
+        tx.commit().await?;
+        Ok(())
     }
 
     async fn is_execution_manager_alive(
--- a/components/spider-storage/src/db/mariadb.rs
+++ b/components/spider-storage/src/db/mariadb.rs
@@ -1,3 +1,5 @@
+use std::{net::IpAddr, time::Duration};
+
 use async_trait::async_trait;
 use const_format::formatcp;
 use secrecy::ExposeSecret;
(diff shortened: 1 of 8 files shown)


sitaowang1998:
@coderabbit review

sitaowang1998:
@coderabbit review

sitaowang1998:
@coderabbit review

components/spider-storage/src/db/protocol.rs
```diff
@@ -312,3 +318,88 @@ pub trait ResourceGroupManagement {
...
+    ///
+    /// * [`DbError::ExecutionManagerAlreadyExists`] if the execution manager ID already exists.
+    /// * Forwards [`sqlx::error::Error`] on DB operation failure.
+    async fn register_execution_manager(
+        &self,
+        execution_manager_id: ExecutionManagerId,
+        ip_address: IpAddr,
+    ) -> Result<(), DbError>;
+
+    /// Updates the heartbeat timestamp of an alive execution manager.
```
```suggestion
    /// Updates the heartbeat of an alive execution manager.
```

components/spider-storage/src/db/protocol.rs
```diff
@@ -312,3 +318,88 @@ pub trait ResourceGroupManagement {
...
+    ///
+    /// # Returns
+    ///
+    /// Whether the execution manager is alive on success.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`sqlx::error::Error`] on DB operation failure.
```
LinZhihao-723, line 380: Shall we return `ExecutionManagerNotFound` if not found?

components/spider-storage/src/db/protocol.rs
```diff
@@ -312,3 +318,88 @@ pub trait ResourceGroupManagement {
...
+    async fn is_execution_manager_alive(
+        &self,
+        execution_manager_id: ExecutionManagerId,
+    ) -> Result<bool, DbError>;
+
+    /// Marks stale execution managers dead and returns their IDs.
+    ///
+    /// This operation is atomic: once an execution manager is returned by this method, it will not
+    /// be returned again in subsequent calls.
```
```suggestion
    /// This operation is atomic: once an execution manager is marked dead and returned by a call of
    /// this method, it will not be returned again in subsequent calls.
```

components/spider-storage/src/db/mariadb.rs
```diff
@@ -487,6 +628,25 @@ CREATE TABLE IF NOT EXISTS `{JOBS_TABLE_NAME}` (
...
+const fn execution_managers_creation_query() -> &'static str {
+    formatcp!(
+        r"
+CREATE TABLE IF NOT EXISTS `{EXECUTION_MANAGERS_TABLE_NAME}` (
+  `id` UUID NOT NULL,
+  `ip_address` VARCHAR(45) NOT NULL,
+  `state` {state_enum} NOT NULL DEFAULT {default_state},
+  `last_heartbeat_at` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `created_at` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `updated_at` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```
LinZhihao-723: Do we need this? I think the only diff between this one and `last_heartbeat_at` is this indicates when the EM is marked as dead. But I'm not quite sure if we need to track that...

components/spider-storage/src/db/protocol.rs
```diff
@@ -312,3 +318,88 @@ pub trait ResourceGroupManagement {
...
+    /// * `execution_manager_id` - The ID of the execution manager to register.
+    /// * `ip_address` - The execution manager IP address.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * [`DbError::ExecutionManagerAlreadyExists`] if the execution manager ID already exists.
+    /// * Forwards [`sqlx::error::Error`] on DB operation failure.
+    async fn register_execution_manager(
```
LinZhihao-723, line 342: Does it make more sense to register and return the ID, instead of passing an EM-generated ID as a parameter?

components/spider-storage/src/db/mariadb.rs
```diff
@@ -430,6 +437,139 @@ impl ResourceGroupManagement for MariaDbStorageConnector {
...
+        let state = sqlx::query_scalar::<_, ExecutionManagerState>(SELECT_STATE_FOR_UPDATE_QUERY)
+            .bind(execution_manager_id)
+            .fetch_optional(&mut *tx)
+            .await?
+            .ok_or(DbError::ExecutionManagerNotFound(execution_manager_id))?;
+
+        if state == ExecutionManagerState::Dead {
+            return Err(DbError::ExecutionManagerAlreadyDead(execution_manager_id));
+        }
+
```
LinZhihao-723, line 501: Do we need a tx here? Would a single statement enough or we can't differentiate row-not-found vs. already dead?

components/spider-storage/src/db/mariadb.rs
```diff
@@ -430,6 +437,139 @@ impl ResourceGroupManagement for MariaDbStorageConnector {
...
+            return Ok(false);
+        };
+
+        match state {
+            ExecutionManagerState::Alive => Ok(true),
+            ExecutionManagerState::Dead => Ok(false),
+        }
+    }
+
+    async fn get_dead_execution_managers(
```
LinZhihao-723, line 534: Just to confirm: the select statement will lock the rows for updates, so if an EM attempts to update the heartbeat in the middle of a this tx, it will be blocked.

components/spider-storage/tests/mariadb_test.rs
```diff
@@ -725,3 +730,245 @@ async fn test_delete_expired_terminated_jobs_no_match() {
...
+/// Registers a new execution manager with a random ID and `127.0.0.1` as the
+/// IP address, then returns the assigned ID.
+async fn register_test_em(storage: &MariaDbStorageConnector) -> ExecutionManagerId {
+    let em_id = ExecutionManagerId::new();
+    storage
+        .register_execution_manager(em_id, IpAddr::V4(Ipv4Addr::LOCALHOST))
+        .await
+        .expect("register_execution_manager should succeed");
+    em_id
+}
```
LinZhihao-723: 1. We should group all the helpers together (not in the middle of test cases).
2. We should enforce the docstring guideline for this method as well.

LinZhihao-723 (changes requested):

components/spider-storage/src/db/mariadb.rs
sitaowang1998: For better usage tracking, I think `created_at` is needed. `updated_at` should probably be replaced by `dead_at`.

components/spider-storage/src/db/mariadb.rs
sitaowang1998: Not in a single statement if we also need to update.

components/spider-storage/src/db/mariadb.rs
sitaowang1998: Yes. If a execution manager lost a few heartbeat and is in the process of being marked dead, a new heartbeat will block and eventually fail with `DbError::ExecutionManagerAlreadyDead`.

components/spider-storage/src/db/protocol.rs
sitaowang1998: This is a good question.

Our current implementation never gc dead execution managers. If we do so later, then it is hard to tell the difference between a dead execution manager that are deleted, and an execution manager that never exists.

We need to decide whether we should distinguish this difference and unify the error interfaces for update and verify.

components/spider-storage/src/db/mariadb.rs
LinZhihao-723: Like OTD? lol
How about `death_confirmed_at`, and don't do `ON UPDATE CURRENT_TIMESTAMP`

components/spider-storage/src/db/mariadb.rs
LinZhihao-723: What would be the error? Or we can only check the number of rows affected?

sitaowang1998: SQL does not allow returning anything from a row we don't update. If we must update the row, we cannot return that there is another row with same id but mismatching state.

components/spider-storage/src/db/protocol.rs
LinZhihao-723: If we don't have GC, that means the DB should preserve all dead EMs forever and returning `ExecutionManagerNotFound` might be necessary to differentiate a dead EM vs. an illegal EM ID.
 
Another way to implement the liveness tracking is that: whenever we call `get_dead_execution_managers`, it actually removes the dead EMs from the DB so that:
- Any further updates will fail.
- Any further call of `get_dead_execution_managers` is guaranteed not to return a previously dead EM.
- `is_execution_manager_alive` only needs to check if the ID exists in the current table.

Both have pros and cons. Let me know ur thoughts.

components/spider-storage/tests/mariadb_test.rs
```diff
@@ -725,3 +742,222 @@ async fn test_delete_expired_terminated_jobs_no_match() {
...
+        .get_dead_execution_managers(1)
+        .await
+        .expect("get_dead_execution_managers should succeed");
+
+    let alive = storage
+        .is_execution_manager_alive(em_id)
+        .await
+        .expect("is_execution_manager_alive should succeed");
+    assert!(!alive, "dead EM should not be alive");
+}
```
LinZhihao-723, line 861: The naming for these test cases are a weird. Probably:
- `test_is_execution_manager_alive_em_alive`
- `test_is_execution_manager_alive_em_not_found`
- `test_is_execution_manager_alive_em_dead`

components/spider-storage/tests/mariadb_test.rs
```diff
@@ -725,3 +742,222 @@ async fn test_delete_expired_terminated_jobs_no_match() {
...
+    );
+}
+
+#[tokio::test]
+#[ignore = "requires MariaDB"]
+#[serial_test::file_serial]
+async fn test_get_dead_execution_managers_multiple() {
+    let storage = create_mariadb_connector().await;
+    let mut em_ids = Vec::with_capacity(3);
+    for _ in 0..3 {
```
LinZhihao-723: We should make 3 a constant.

components/spider-storage/tests/mariadb_test.rs
```diff
@@ -725,3 +742,222 @@ async fn test_delete_expired_terminated_jobs_no_match() {
...
+    }
+
+    // Verify all are now dead.
+    for em_id in &em_ids {
+        let alive = storage
+            .is_execution_manager_alive(*em_id)
+            .await
+            .expect("is_execution_manager_alive should succeed");
+        assert!(!alive, "EM {em_id:?} should be dead");
+    }
```
LinZhihao-723: - Should merge into one loop?
- Should check the size of `dead` matches `em_ids`?

LinZhihao-723 (changes requested):
A few more nit catches.

components/spider-storage/tests/mariadb_test.rs
sitaowang1998: The size check would be wrong. There could be other dead execution managers left over by previous tests.

components/spider-storage/tests/mariadb_test.rs
```diff
@@ -727,6 +742,221 @@ async fn test_delete_expired_terminated_jobs_no_match() {
...
+#[ignore = "requires MariaDB"]
+async fn test_update_execution_manager_heartbeat() {
+    let storage = create_mariadb_connector().await;
+    let em_id = register_test_em(&storage).await;
+
+    storage
+        .update_execution_manager_heartbeat(em_id)
+        .await
+        .expect("update_execution_manager_heartbeat should succeed");
+}
```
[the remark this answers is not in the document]
sitaowang1998: This path is covered by other tests.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -457,6 +587,8 @@ const MYSQL_ER_DUP_ENTRY: u16 = 1062;
 
 const RESOURCE_GROUPS_TABLE_NAME: &str = "resource_groups";
 const JOBS_TABLE_NAME: &str = "jobs";
+
```
```suggestion
```

components/spider-storage/src/db/protocol.rs
```diff
@@ -15,15 +17,15 @@ use crate::db::error::DbError;
 /// * [`ExternalJobOrchestration`]
 /// * [`InternalJobOrchestration`]
 /// * [`ResourceGroupManagement`]
+/// * [`ExecutionManagerLivenessManagement`]
 /// * [`SessionManagement`]
 #[async_trait]
-#[rustfmt::skip] // Workaround for https://github.com/rust-lang/rustfmt/issues/5321
```
LinZhihao-723, line 20: I assume it's fixed by adding `ExecutionManagerLivenessManagement`?

LinZhihao-723 (changes requested):

LinZhihao-723 (approved):
For the PR title, how about:
```
feat(spider-storage): Add MariaDB implementation for execution manager liveness tracking.
```

PR #315, opened 2026-04-28
feat(spider-tdl): Add `#[task]` proc macro for task declaration:

LinZhihao-723:
* Add `spider-tdl-derive` crate housing the `#[task]` macro.
* Re-export `#[task]` from `spider-tdl` behind the optional `derive` feature.

# Description

As the title suggests. This PR is one of the PR series #302 that implements the task proc-macro.

# Checklist

  breaking change.

# Validation performed

* [x] Add unit tests to assert the generated tasks match the expected.
* [x] Add integration tests to assert the macro-converted tasks are usable by the TDL task framework.

## Summary by CodeRabbit

* **New Features**
  * Added a #[task] procedural macro attribute for concise task definitions, automatic parameter (de)serialization, standardized return shaping and error handling.
  * Introduced an optional "derive" feature and a new derive crate to enable compile-time validation and feature-gated task re-exporting.
  * Tasks support overridable names via attribute metadata.

* **Tests**
  * Added end-to-end integration tests covering macro-generated tasks, serialization roundtrips, execution and error propagation.

### DIFF
--- a/components/spider-tdl-derive/src/task_macro.rs
+++ b/components/spider-tdl-derive/src/task_macro.rs
@@ -1,0 +1,602 @@
+//! Implementation of the `#[task]` attribute macro.
+//!
+//! The macro replaces the annotated function with three items:
+//!
+//! 1. A unit marker struct that shares the function's identifier and visibility.
+//! 2. A private params struct holding the non-context parameters, with
+//!    `#[derive(serde::Deserialize)]` so the runtime can rebuild it from wire bytes.
+//! 3. An `impl spider_tdl::Task` for the marker struct that wires the params back into the
+//!    user-authored function body.
+//!
+//! See the crate-level docs of [`spider_tdl_derive`] for usage examples.
+
+use proc_macro2::TokenStream;
+use quote::{format_ident, quote};
+use syn::{
+    FnArg,
+    GenericArgument,
+    ItemFn,
+    LitStr,
+    Pat,
+    PathArguments,
+    ReturnType,
+    Token,
+    Type,
+    parse::{Parse, ParseStream},
+};
+
+/// Parsed representation of the `#[task(...)]` attribute arguments.
+pub struct TaskAttr {
+    /// The name alias of the task.
+    name: Option<LitStr>,
+}
+
+impl Parse for TaskAttr {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        if input.is_empty() {
+            return Ok(Self { name: None });
+        }
+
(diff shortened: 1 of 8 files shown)


components/spider-tdl-derive/src/task_macro.rs
```diff
@@ -0,0 +1,605 @@
...
+
+    let first_param = func
+        .sig
+        .inputs
+        .first()
+        .expect("validated that function has at least one parameter");
+    let FnArg::Typed(first_pat_type) = first_param else {
+        unreachable!("`self` parameters are rejected by `validate_no_self`");
+    };
+    let first_param_type = &first_pat_type.ty;
```
sitaowang1998: Since both `expect` and `unreachable` should be unreachable, why not merge into one?

components/spider-tdl-derive/src/task_macro.rs
```diff
@@ -0,0 +1,605 @@
...
+        ));
+    };
+
+    let last_segment = type_path
+        .path
+        .segments
+        .last()
+        .expect("return type path should have at least one segment");
+
+    if last_segment.ident != "Result" {
```
sitaowang1998, line 305: I am worried about just comparing string with `Result`. What if user use type alias?

LinZhihao-723: Yeah, that's a good concern which I also noticed.
Unfortunately, it's hard to resolve to the aliased symbol in the proc macro. My opinions for this one are:
- We already enforce the generic parameter size to be 2, meaning that both Ok and Err type should be explicitly given.
- In that case, it is a really weird case that a user creates an alias for a result type. I suspect no one would do that in reality.

Since, for now, our code only serves CLP (and potentially through a TDL compiler), I ended up leaving this verbose comparison here as a sanity check. We can iterate in the future if we found it would cause any real problems.

components/spider-tdl-derive/src/task_macro.rs
LinZhihao-723: Sorry, I don't get what u mean. Did you mean to use unreachable to replace `.expect`?

sitaowang1998: I am thinking of something like
```suggestion
    let Some(FnArg::Typed(first_pat_type)) = func.sig.inputs.first() else {
        unreachable!("validated that function has at least one parameter which is not `self`");
    };
    let first_param_type = &first_pat_type.ty;
```

sitaowang1998 (approved):

PR #319, opened 2026-05-07
feat(spider-storage): Add `ServiceState` and job level functions.

sitaowang1998:
# Description

This PR:
- Adds `ServiceState` object that holds all storage services.
- Adds job level functions in `SerivceState`.
- Extracts common mock objects in `test_mocks.rs`.

# Checklist

  breaking change.

# Validation performed

- [x] Adds new unit tests.
- [x] GitHub workflows pass.

## Release Notes

* **New Features**
  * Added job management service API with registration, state tracking, and cancellation
  * Added task instance lifecycle operations (creation, completion, and failure handling)
  * Added job output retrieval and improved error reporting for job not found scenarios

[Review Change Stack](https://app.coderabbit.ai/change-stack/y-scope/spider/pull/319)


sitaowang1998:
@coderabbit review

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,682 @@
...
+    DbConnectorType: DbStorage,
+    TaskInstancePoolConnectorType: TaskInstancePoolConnector,
+> {
+    db: DbConnectorType,
+    session_id: SessionId,
+    job_cache: Arc<JobCache<ReadyQueueSenderType, DbConnectorType, TaskInstancePoolConnectorType>>,
+    ready_queue_sender: ReadyQueueSenderType,
+    _ready_queue_receiver: ReadyQueueReceiverHandle,
+    task_instance_pool_connector: TaskInstancePoolConnectorType,
+}
```
LinZhihao-723, line 48: We shouldn't make this whole struct cloneable: basically, it means we need to clone 6 arc, which is not the most efficient way to implement the state. Can we:
1. Move the data into a dedicated struct called `ServiceStateInner`. Notice that in this inner struct, the job cache is not wrapped by an arc.
2. The actual `ServiceState` holds an arc of `ServiceStateInner`. `clone` now should only need to clone one arc.
3. All methods are still implemented for `ServiceState`, but operating on the inner data.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,682 @@
...
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`JobCache::insert`]'s return values on failure.
+    /// * Forwards [`ExternalJobOrchestration::register`]'s return values on failure.
+    /// * Forwards [`SharedJobControlBlock::create`]'s return values on failure.
+    pub async fn register_job(
+        &self,
+        resource_group_id: ResourceGroupId,
+        task_graph: &spider_core::task::TaskGraph,
```
LinZhihao-723: Since this will be the serviceable layer, shouldn't the input be a serialized task graph? This call should first deserialize the task graph to make sure it's valid.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,682 @@
...
+        job_inputs: Vec<TaskInput>,
+    ) -> Result<JobId, StorageServerError> {
+        let job_id = self
+            .db
+            .register(resource_group_id, task_graph, &job_inputs)
```
LinZhihao-723: I realized we forgot to address #284 so that the task graph vs. inputs are not validated before registration... we should probably fix that first before we implement this registration method.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,682 @@
...
+    /// Gets the job control block from the cache and starts it by calling
+    /// [`SharedJobControlBlock::start`].
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * [`StorageServerError::JobNotFound`] if the job is not in the cache.
+    /// * Forwards [`SharedJobControlBlock::start`]'s return values on failure.
+    pub async fn submit_job(&self, job_id: JobId) -> Result<(), StorageServerError> {
```
LinZhihao-723: It's a bit confusing to use the term "submit". We should probably just call it "start" to match our current protocol.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,682 @@
...
+    pub async fn cancel_job(&self, job_id: JobId) -> Result<JobState, StorageServerError> {
+        let jcb = self
+            .job_cache
+            .get(job_id)
+            .ok_or(StorageServerError::JobNotFound(job_id))?;
+        let state = jcb.cancel().await?;
+        if state.is_terminal() {
+            // OK if already removed by a concurrent operation.
+            drop(self.job_cache.remove(job_id));
+        }
```
LinZhihao-723: I don't think we need to drop the job immediately. We can have a GC task running in the background later. This is because if you look at the implementation of `get_job_state`, it's actually by-passing the cache layer JCB to directly access the DB. I think the most proper way is to treat the entire thing as a real cache behavior: we try to access the JCB to query the state first; if the JCB doesn't exist, we fall back to the DB layer. Keeping the JCB in the cache for a little longer may accelerate state/result access.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,682 @@
...
+    /// The state of the job on success.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`ExternalJobOrchestration::get_state`]'s return values on failure.
+    pub async fn get_job_state(&self, job_id: JobId) -> Result<JobState, StorageServerError> {
+        Ok(self.db.get_state(job_id).await?)
+    }
```
LinZhihao-723, line 221: Please refer to https://github.com/y-scope/spider/pull/319/changes#r3204181094.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,682 @@
...
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`ExternalJobOrchestration::get_outputs`]'s return values on failure.
+    pub async fn get_job_outputs(
+        &self,
+        job_id: JobId,
+    ) -> Result<Vec<TaskOutput>, StorageServerError> {
+        Ok(self.db.get_outputs(job_id).await?)
```
LinZhihao-723: Please refer to https://github.com/y-scope/spider/pull/319/changes#r3204181094.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,682 @@
...
+    ///
+    /// Returns an error if:
+    ///
+    /// * [`StorageServerError::JobNotFound`] if the job is not in the cache.
+    /// * Forwards [`SharedJobControlBlock::create_task_instance`]'s return values on failure.
+    pub async fn create_task_instance(
+        &self,
+        job_id: JobId,
+        task_id: TaskId,
+        execution_manager_id: ExecutionManagerId,
```
LinZhihao-723, line 279: Shall we take the input session ID and match the session inside this method?
Similar for other two task instance APIs.

components/spider-storage/src/state/service.rs
LinZhihao-723, line 1: We should introduce tracing logging in this PR since these service calls should be logged.

LinZhihao-723 (changes requested):
Reviewed service implementation and brought up some design concerns. Will go for a detailed review in the next round.

components/spider-storage/src/state/service.rs
sitaowang1998: Job control block does not directly store the outputs/error of the job. Should we add support for that in JCB or keep it as it is now?

components/spider-storage/src/state/service.rs
sitaowang1998: Do we assume that task graph and inputs are serialized in the same way across RESTful and gRPC? If not, I think it is better to leave serialization to the request handler.

LinZhihao-723: The task graph will be serialized using JSON, right?

components/spider-storage/src/state/service.rs
LinZhihao-723: - For output, yes, I think it's better if we can directly fetch from memory if the job is still in the cache. This is also a performance optimization.
- For error, I think it's optional. I think the error can always be read from the DB.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,1051 @@
...
+    DbConnectorType: DbStorage,
+    TaskInstancePoolConnectorType: TaskInstancePoolConnector,
+> {
+    db: DbConnectorType,
+    session_id: SessionId,
+    job_cache: JobCache<ReadyQueueSenderType, DbConnectorType, TaskInstancePoolConnectorType>,
+    ready_queue_sender: ReadyQueueSenderType,
+    _ready_queue_receiver: ReadyQueueReceiverHandle,
+    task_instance_pool_connector: TaskInstancePoolConnectorType,
+}
```
LinZhihao-723: * Relocate after public ones.
* Type parameters annotation is missing.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,1051 @@
...
+    ReadyQueueSenderType: ReadyQueueSender,
+    DbConnectorType: DbStorage,
+    TaskInstancePoolConnectorType: TaskInstancePoolConnector,
+> ServiceState<ReadyQueueSenderType, DbConnectorType, TaskInstancePoolConnectorType>
+{
+    /// Creates a new `ServiceState` from its constituent parts.
+    ///
+    /// # Returns
+    ///
+    /// A newly created `ServiceState`.
```
LinZhihao-723:
```suggestion
    /// Factory function.
    ///
    /// # Returns
    ///
    /// A newly created [`ServiceState`] from its constituent parts.
```
Symbol annotation.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,1051 @@
...
+    }
+
+    /// Registers a job in the database and inserts its control block into the cache.
+    ///
+    /// Deserializes the serialized task graph and job inputs, validates their consistency,
+    /// then persists the job and creates its control block.
+    ///
+    /// If [`SharedJobControlBlock::create`] or [`JobCache::insert`] fails after the DB record has
+    /// been created, the DB record is **not** deleted.
+    ///
```
LinZhihao-723:
```suggestion
```
I don't think these two paragraph are needed.

components/spider-core/src/types/io.rs
```diff
@@ -22,3 +22,39 @@ pub struct ExecutionContext {
...
+///
+/// Returns `rmp_serde::decode::Error` if any input fails to deserialize.
+pub fn deserialize_job_inputs(
+    inputs: &SerializedJobInputs,
+) -> Result<Vec<TaskInput>, rmp_serde::decode::Error> {
+    inputs
+        .iter()
+        .map(|bytes| rmp_serde::from_slice(bytes))
+        .collect()
+}
```
LinZhihao-723: Looking at how this function is used, I think it's wrong.
The task inputs given at creation time should already be serialized as msgpack bytes individually (each input is serialized as an independent payload). We should not deserialize the byte payload.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,1051 @@
...
+        self.inner.job_cache.insert(jcb)?;
+        debug!("Inserted JCB into job cache.");
+
+        Ok(job_id)
+    }
+
+    /// Starts a job for execution.
+    ///
+    /// Gets the job control block from the cache and starts it by calling
+    /// [`SharedJobControlBlock::start`].
```
LinZhihao-723:
```suggestion
```
A lot of these AI-generated docs are useless. We should remove them and make the docstring's short description clean.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,1051 @@
...
+            .get(job_id)
+            .ok_or(StorageServerError::JobNotFound(job_id))?;
+        debug!("JCB found in cache, starting job.");
+        jcb.start().await?;
+        Ok(())
+    }
+
+    /// Cancels a job.
+    ///
+    /// If the job is in the cache, delegates to [`SharedJobControlBlock::cancel`].
```
LinZhihao-723:
```suggestion
```
Same.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,1051 @@
...
+    #[instrument(skip(self), fields(job_id = ?job_id))]
+    pub async fn cancel_job(&self, job_id: JobId) -> Result<JobState, StorageServerError> {
+        let jcb = self
+            .inner
+            .job_cache
+            .get(job_id)
+            .ok_or(StorageServerError::JobNotFound(job_id))?;
+        debug!("JCB found in cache, cancelling job.");
+        let state = jcb.cancel().await?;
+        Ok(state)
```
LinZhihao-723: Do we want to make the cancellation operation idempotent?
The job might not always appear in the cache: for example, if it is cancelled in the previous session and the storage service restarts, it might not need to be loaded into the cache again. That means, we shouldn't fall back to job-not-found immediately but also check the DB.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,1051 @@
...
+    ///
+    /// # Returns
+    ///
+    /// The execution context for the created task instance on success.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * [`StorageServerError::StaleSession`] if the session has changed.
```
LinZhihao-723, line 271: This is not a direct return right?

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,1051 @@
...
+    ///
+    /// # Returns
+    ///
+    /// The current job state after the operation on success.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * [`StorageServerError::StaleSession`] if the session has changed.
```
LinZhihao-723, line 311: Same.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,1051 @@
...
+        skip(self, session_id, task_outputs),
+        fields(job_id = ?job_id, task_instance_id = ?task_instance_id)
+    )]
+    pub async fn succeed_task_instance(
+        &self,
+        session_id: SessionId,
+        job_id: JobId,
+        task_instance_id: TaskInstanceId,
+        task_index: TaskIndex,
+        task_outputs: Vec<TaskOutput>,
```
LinZhihao-723: The succeed request won't provide a vector of task output. It will send wire-format serialized bytes. We should assume task_outputs are Vec<u8> and deserialize it into Vec<TaskOutput> using [this](https://github.com/y-scope/spider/blob/aadb9eb657523b83ce88c77a5761ed726eb0828e/components/spider-tdl/src/wire.rs#L260) method.

components/spider-storage/src/state/service.rs
```diff
@@ -0,0 +1,1051 @@
...
+    /// * [`StorageServerError::JobNotFound`] if the job is not in the cache.
+    /// * Forwards [`SharedJobControlBlock::create_task_instance`]'s return values on failure.
+    #[instrument(skip(self, session_id), fields(job_id = ?job_id, task_id = ?task_id))]
+    pub async fn create_task_instance(
+        &self,
+        session_id: SessionId,
+        job_id: JobId,
+        task_id: TaskId,
+        execution_manager_id: ExecutionManagerId,
+    ) -> Result<ExecutionContext, StorageServerError> {
```
LinZhihao-723, line 280: Similar to task output, I need to update `ExecutionContext` to serialize task inputs as wire-format serialized raw bytes. I will do it in a separate PR.

LinZhihao-723 (changes requested):
Haven't looked into test cases yet. I will do some direct push to shape the logging. The current way of logging is not making things clear: it's not obvious what fields are logged by reading the log print statement.

components/spider-storage/src/state/service.rs
sitaowang1998: I think this is a discrepancy from the `register_job` signature. Either the storage layer accepts all inputs/outputs as a vector of byte arrays, or they should all accept wire-format serialized bytes.

LinZhihao-723: It doesn't matter what's inside the function. The service layer should take the wire-format serialized raw bytes and deserialize the buffer into `Vec<TaskOutput>`.

LinZhihao-723 (approved):
I'm done with my editing. Please check my changes (especially the logging statements).

PR #325, opened 2026-05-14
feat(spider-task-executor): Add executor binary with bincode wire protocol and integration tests.

LinZhihao-723:
# Description

## Wire protocol (`spider-task-executor/src/protocol.rs`)

A new `protocol` module on the `spider-task-executor` library defines the three wire types the execution manager will use to drive the child:

* `Request::Execute { tdl_context, raw_ctx, raw_inputs }` and `Request::Shutdown`. The parent owns the hard timeout in its entirety; the executor has no notion of timeouts and the request carries no deadline.
* `Response::Result { outcome, elapsed_us }` — exactly one per `Execute`. `elapsed_us` is the in-FFI wall-clock measured by the executor and is what the overhead instrument uses to separate executor-side cost from parent-side IPC.
* `ExecutorOutcome::Success { outputs } | Failure { error }` — `outputs` is the wire-format `TaskOutputsSerializer` buffer ready to forward to storage; `error` is the msgpack-encoded `ExecutorError`.

Stderr is not carried over the protocol; how the spawner disposes of the executor's stderr (inherit / pipe / log file) is a parent-side decision.

## Executor binary (`spider-task-executor/src/bin/spider_task_executor.rs`)

Single-threaded tokio runtime; requests are processed strictly sequentially with exactly one task running for the lifetime of the process. Tokio is here only to match the async I/O surface the execution manager uses (`tokio_util::codec::LengthDelimitedCodec`); the executor itself has no concurrency requirements.

Key shape decisions:

* The FFI call runs **inline** on the runtime thread — no second OS thread, no `oneshot`, no `tokio::select!`. The previous design dispatched the FFI on a `std::thread` so the runtime could `select!` an in-process timer against the FFI; with the timer responsibility consolidated on the parent, none of that scaffolding earns its keep.
* `SPIDER_TDL_PACKAGE_DIR` is validated once at startup. If unset the binary exits non-zero before processing any request, which surfaces a deployment misconfiguration immediately rather than per-request.
* Package resolution: `${SPIDER_TDL_PACKAGE_DIR}/<package>/lib<package>.so`. The first request for a package dlopens the library; subsequent requests reuse the cached `TdlPackage`.
* Tracing init: JSON, ANSI off, env-filtered, written to **stderr** so it doesn't pollute the framed-stdout protocol channel. Both `tracing` and `tracing-subscriber` are pulled in with `default-features = false` and only the features actually used (`std` for the macros; `fmt`, `env-filter`, `json` for the subscriber).

## `ExecutorError` is now wire-friendly

`ExecutorError` derives `serde::Serialize`/`Deserialize` so the binary can ship it across the protocol as `Failure { error: rmp_serde::to_vec(&err) }` and the EM can decode it back to a typed value. Three variants used to wrap external types that don't implement `Serialize` (`libloading::Error`, `std::str::Utf8Error`, `rmp_serde::decode::Error`); they now carry the `Display` rendering of the source error as a `String`. Explicit `From` impls preserve the lossless `?` propagation in `manager.rs`. The wildcard `matches!(err, ExecutorError::InvalidLibrary(_))` pattern in the existing `huntsman/tdl-integration` tests still compiles unchanged.

### `tests/huntsman/integration-test-tasks` (TDL package)

A cdylib + rlib package (`crate-type = ["cdylib", "rlib"]`) registered under TDL package name `integration_test_tasks`. The dual crate-type lets the bench reference compile-time constants (notably `INSTRUMENT_SLEEP_US`) while keeping the cdylib for `dlopen` from the executor.

Tasks:

* `fibonacci(index: u64) -> u64` — naive recursion; correctness check.
* `always_fail() -> Result` — returns `TdlError::ExecutionError`.
* `always_panic() -> !` — panics; the panic crosses the `extern "C"` FFI boundary and aborts the process, which is exactly the crash signal the parent test asserts on.
* `instrument(items: Vec<String>) -> Vec<String>` — sleeps for a fixed `INSTRUMENT_SLEEP_US` (50µs) and echoes the payload back. Used by the overhead bench.

### `tests/huntsman/task-executor` (executor integration tests)

A library crate that provides an `ExecutorHandle` harness (spawn the binary, frame requests on stdin, decode responses from stdout) and two `[[test]]` binaries.

The harness panics with descriptive `.expect(...)` messages on protocol / I/O / decode failures rather than threading errors through every helper — these are infrastructure, not production code, and a panic with backtrace points at the failure site immediately. (This pattern surfaced one subtle bug during development: a stale `.so` left over from a task rename produced `TaskNotFound("instrument")` in the panic message verbatim, which was the entire diagnosis.)

`tests/executor.rs` covers:

* `fibonacci_returns_correct_value` — encodes a single `u64` input; asserts `Success` and that the decoded `u64` equals 55.
* `always_fail_reports_task_error` — asserts `Failure` whose msgpack-decoded `ExecutorError` is `TaskError(TdlError::ExecutionError(_))` and whose message contains the task name.
* `always_panic_crashes_the_process` — sends `Execute`, expects stdout EOF before any frame arrives, then waits for the child to exit non-zero.

`tests/overhead_instrument.rs` runs the `instrument` task ten times against a long-lived executor (so dlopen happens once during a discarded warm-up) and writes a markdown table at `${SPIDER_TEST_INSTRUMENT_OUTPUT_DIR}/task_executor_overhead.md`. With the work portion held constant at 50µs the table separates four metrics:

| Metric | What it measures |
|---|---|
| `E2E (parent)` | `Instant`-to-`Instant` around `send(Execute)` → `recv(Response::Result)` |
| `Executor FFI` | `elapsed_us` reported by the executor (sleep + in-FFI input/output serde) |
| `Executor internal (FFI - sleep)` | In-executor input/output serde alone |
| `IPC overhead (E2E - FFI)` | Parent-side framing + bincode + pipe traversal |

Sample output from a local run:

```
| Metric                          | Count | Avg (µs) | P50 (µs) | P95 (µs) | P99 (µs) |
| E2E (parent)                    | 10    | 476.00   | 479.64   | 602.06   | 602.06   |
| Executor FFI                    | 10    | 157.80   | 164.00   | 175.00   | 175.00   |
| Executor internal (FFI - sleep) | 10    | 107.80   | 114.00   | 125.00   | 125.00   |
| IPC overhead (E2E - FFI)        | 10    | 318.20   | 315.64   | 429.06   | 429.06   |
```

## Taskfile changes (`taskfiles/test.yaml`)

`spider-huntsman-unit-tests-executor` now:

1. Builds three artifacts in separate `cargo build` invocations (combining `--package <cdylib>` with `--bin <name>` would silently exclude the cdylibs from the target selection): `huntsman-complex`, `integration-test-tasks`, and `spider-task-executor --bin spider-task-executor`.
2. Stages the cdylibs under `build/tdl_packages/<package>/lib<package>.so` — the standard layout the executor binary reads via `${SPIDER_TDL_PACKAGE_DIR}/<package>/lib<package>.so`.
3. Sets `SPIDER_TDL_PACKAGE_DIR`, `SPIDER_TASK_EXECUTOR_BIN`, and (relocates) `SPIDER_TDL_PACKAGE_COMPLEX` to point at the staged paths. Existing `huntsman-complex` consumers see no behavior change because their tests load the `.so` by absolute path.
4. Invokes `cargo nextest run --all --all-features --run-ignored all --release`. The new tests are gated `#[ignore]` so plain `cargo test` (which doesn't go through the taskfile) can't accidentally try to run them without env vars set.

# Checklist

  breaking change.

# Validation performed

* [x] Add integration tests to test task executions that:
  * Return results.
  * Return errors.
  * Panic/crash.

## Summary by CodeRabbit

* **New Features**
  * Added a standalone task-executor subprocess with a framed request/response protocol that returns serialized outcomes and elapsed execution time.
  * Added an integration test TDL package with sample tasks (fibonacci, fail, panic, sleep-and-echo).

* **Tests**
  * Added end-to-end integration tests for success, failure, and crash scenarios.
  * Added an instrumentation test measuring round-trip latency, in-executor FFI time and IPC overhead.

* **Chores**
  * Expanded workspace and test setup to build and stage test packages and the executor binary.

[Review Change Stack](https://app.coderabbit.ai/change-stack/y-scope/spider/pull/325?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


components/spider-task-executor/src/bin/spider_task_executor.rs
```diff
@@ -0,0 +1,146 @@
...
+                .with_target(false)
+                .with_file(true)
+                .with_line_number(true)
+                .json(),
+        )
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .with_ansi(false)
+        .with_writer(std::io::stderr)
+        .init();
+}
```
sitaowang1998, line 51: I think many more components will need to initialize tracing later one. How about putting this into a util file that is shared among components?

taskfiles/test.yaml
```diff
@@ -209,24 +209,43 @@ tasks:
...
+        # artifact to keep the target selection unambiguous.
+        cargo build --release --package huntsman-complex
+        cargo build --release --package integration-test-tasks
+        cargo build --release --package spider-task-executor --bin spider-task-executor
+        mkdir -p "{{.G_TDL_PACKAGES_DIR}}/complex" \
+                 "{{.G_TDL_PACKAGES_DIR}}/integration_test_tasks"
+        cp "{{.G_RUST_RELEASE_DIR}}/libhuntsman_complex.so" \
+           "{{.G_TDL_PACKAGES_DIR}}/complex/libcomplex.so"
+        cp "{{.G_RUST_RELEASE_DIR}}/libintegration_test_tasks.so" \
+           "{{.G_TDL_PACKAGES_DIR}}/integration_test_tasks/libintegration_test_tasks.so"
```
sitaowang1998, line 248: Maybe use soft link instead of coping `so` files around?

tests/huntsman/integration-test-tasks/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub fn always_panic(_ctx: TaskContext) -> Result<u64, TdlError> {
+        panic!("always_panic: intentional panic")
+    }
+
+    /// Sleeps for a fixed [`INSTRUMENT_SLEEP_US`] microseconds, then echoes the input back.
+    ///
+    /// The fixed-cost body lets the overhead bench subtract the known sleep from the executor's
+    /// reported FFI duration, isolating the in-executor input/output serde overhead.
+    #[task(name = "instrument")]
+    pub fn instrument(_ctx: TaskContext, items: Vec<String>) -> Result<Vec<String>, TdlError> {
```
sitaowang1998: `instrument` is not a descriptive name. How about `sleep_and_echo`?

components/spider-task-executor/src/bin/spider_task_executor.rs
LinZhihao-723: I was thinking the same though. But I realized other components may require different log rotation strategies and output sources: for example, in this component logs are only written to stderr, but it might not be true for the EM service.

taskfiles/test.yaml
LinZhihao-723: I think our preference is to copy things when possible. Soft links can be more confusing when we need to debug it...

sitaowang1998: But the objective of the tests is to find out if anything could go wrong instead of avoiding them. We need to make sure that our library loading works with valid soft links.

LinZhihao-723: I'm not sure if I got it. Why is soft link loading necessary to test?
The current design is to make the overall TDL package like a Python package, so all dylibs, including their dependencies, should be located in one directory. Soft link is not a part of the loading requirement afaik.
And in that case, the directory containing the actual dylib is simulating our real use-case.

sitaowang1998: Because I know some C and C++ libraries have problems with following soft links when reading files. I am not sure if any rust library has this problem, but it is good to guard against this.

LinZhihao-723: - I don't think this is a related concern to the current test.
  - None of our testing libs are C/C++ libs.
  - It is more likely to be a problem of the libraries themselves or how people package their dependencies.
- The purpose of my design above is to avoid the soft link from scratch: all dependencies (dylibs) are packaged into the same directory with their rpath modified to the proper locations.

I'd strongly prefer not to include any soft link issues in our considerations at this stage.

sitaowang1998 (approved):

PR #326, opened 2026-05-20
feat(spider-execution-manager): Add single-process supervisor pool for the task executor.

LinZhihao-723:
# Description

This PR depends on #325.

## New crate

This PR introduces a new crate `spider-execution-manager`. Its first module, `process_pool`, is a single-process supervisor that owns one `spider-task-executor` subprocess and serializes access to it. The public surface is one factory (`ProcessPool::new`) and one async API (`ProcessPool::execute(request, hard_timeout) -> Result<Outcome, InternalError>`). Multi-slot pools are deliberately deferred — the EM main loop only needs one process today, and the smaller surface keeps the consumer's integration minimal.

## Lifecycle and concurrency

The pool spawns the initial executor at construction. Concurrent `execute` callers queue on a single `tokio::Mutex` around the handle slot; the guard is held for the entire call, so the mutex is the entire concurrency gate (no separate semaphore). The parent is the sole authority for the hard timeout — if either the parent's timer fires or the child exits / closes stdout, the dead handle is dropped (`Command::kill_on_drop(true)` SIGKILLs the OS process), a fresh executor is spawned in the same critical section, and the next caller sees a healthy process. The four channel-level failure modes inside the protocol exchange (framed write failure, framed read failure, stdout EOF, response decode failure) all collapse into `Outcome::ExecutorCrash`, which feeds the same respawn branch as `Outcome::Timeout`.

## Logging and observability

Each spawn allocates a monotonic `executor_id` and routes the child's stderr to `<log_dir>/<em_id>-<executor_id>.log` in create-or-append mode (rather than inheriting the parent's stderr). Per-spawn filenames mean each respawn naturally rotates onto a fresh file; a size/time rotator at the executor side is the documented next step. All pool-emitted tracing events carry `executor_id` as a typed `u64` field so JSON subscribers render it as a number; ID newtypes that lack `Display` use debug formatting.

## Error model

`InternalError` is the pool's own failure surface, distinct from a task-level `Outcome`. Four variants cover the cases where the pool can't serve the current call: `NotRunning` (a prior respawn failed and the pool is unrecoverable), `ExecutorCreationFailure` (any I/O step during spawn), `EncodeTaskContext`, and `EncodeTaskInputs`. All non-`NotRunning` variants use `thiserror`'s `#[from]` so the body propagates with `?` throughout. There are no panics in either the happy path or the error path of `execute`.

## Cancellation safety (known limitation)

`execute` holds the mutex guard across `.await` points. If the caller's future is dropped mid-flight (e.g. a higher-layer timeout, `tokio::select!` losing the branch, ctrl-C), the guard releases but the executor keeps running and will eventually push a response into the stdout pipe — which the next `execute` call would read as the first frame of its own request. The protocol no longer carries `instance_id`, so there's no automatic mismatch detection. The EM main loop will own these futures to completion, so this isn't reachable in practice today; documented as a follow-up to harden via a Drop guard that flags the handle dirty on cancellation.

## Test coverage

End-to-end tests added at `tests/huntsman/task-executor/tests/test_process_pool.rs`, against the real task-executor binary using the existing harness:

* `fibonacci_succeeds` — happy path.
* `always_fail_reports_task_error` — in-task failure surfaces a decodable `ExecutorError`.
* `always_panic_returns_crash_then_respawns` — panicking task returns `Outcome::ExecutorCrash`; a follow-up `execute` on the same pool succeeds, proving the respawn.
* `hard_timeout_kills_then_respawns` — long-running task plus a short hard timeout returns `Outcome::Timeout`; a follow-up `execute` succeeds, proving the SIGKILL + respawn.

The respawn assertion is functional — the follow-up call returning `Outcome::Success` is the visible evidence that a fresh process is alive and serving. No internal pool state is inspected.

# Checklist

  breaking change.

# Validation performed

* [x] Add test coverage as discussed above.

## Summary by CodeRabbit

* **New Features**
  * Added execution manager component with process pool support for handling task execution

* **Tests**
  * Added comprehensive integration tests for process pool functionality, including success scenarios, error handling, and recovery cases

* **Chores**
  * Updated and standardized dependency configurations across workspace crates for consistency

[Review Change Stack](https://app.coderabbit.ai/change-stack/y-scope/spider/pull/326?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


sitaowang1998 (commented):
We likely need to cancel a running task, thus the process pool should return a handle, which supports both cancellation of a running task and waiting for outcome.

sitaowang1998:
With off-line discussion, we decided that cancel is not in the scope of first stage implementation of `Spider`.

sitaowang1998 (approved):

LinZhihao-723:
> With off-line discussion, we decided that cancel is not in the scope of first stage implementation of `Spider`.

As discussed offline, cancellation support involves both protocol-level support and EM-level support, but neither are included in the current design. We will defer cancellation implementation to future PRs.

PR #327, opened 2026-05-20
feat(spider-execution-manager): Add scheduler, storage, and liveness client traits.

LinZhihao-723:
# Description

This PR depends on #326.

This PR adds the network-client trait surface for the execution manager. Three async traits cover the EM's outbound traffic to the rest of the cluster; concrete implementations bind to the corresponding remote services and will land in follow-up PRs.

# Checklist

  breaking change.

# Validation performed

## Release Notes

* **New Features**
  * Added execution manager client interfaces for liveness registration, task scheduling, and result tracking.
  * Enabled task coordination across execution manager nodes.

* **Chores**
  * Added async-trait dependency to support asynchronous client operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/y-scope/spider/pull/327?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


components/spider-execution-manager/src/client/liveness.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// # Returns
+    ///
+    /// The freshly assigned execution manager id and the current storage session id on success.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * [`LivenessResponseError::Transport`] if the connection was lost or timed out.
+    /// * [`LivenessResponseError::IllegalId`] if storage rejected the registration.
```
sitaowang1998: Why `IllegalId` for the storage rejection? Isn't id generated by the storage?

sitaowang1998 (changes requested):
The `Liveness` and `Scheduler` error does not have a catch-all branch:
* `Liveness` is currently implemented in the storage, so it might return more errors like storage.
* We don't have a concrete scheduler implementation yet, but having a catch-all is more future proof.

components/spider-execution-manager/src/client/liveness.rs
LinZhihao-723: True, this is a mistake. It should not return an illegal ID error.

LinZhihao-723:
> The `Liveness` and `Scheduler` error does not have a catch-all branch:
> 
> * `Liveness` is currently implemented in the storage, so it might return more errors like storage.
> * We don't have a concrete scheduler implementation yet, but having a catch-all is more future proof.

We probably don't need catch-all errors, depending on the type of server we implement. For example, the server side may return RESTful error codes, and our error types should be derived from all possible error codes. And in some sense, we should avoid using catch-all errors 'cuz the main loop implementation may need to know the exact error type to determine whether it should keep serving or terminate the runtime.

sitaowang1998 (approved):

PR #328, opened 2026-05-22
feat(spider-execution-manager): Add liveness actor with session ID tracker; Refactor integration tests to extract common helpers into `test-utils`.

LinZhihao-723:
# Description

This PR depends on #327.

This PR adds the liveness actor and the shared session-tracker primitive that the rest of the execution-manager runtime will plug into.

**`spider_core::session::SessionTracker`** — a forward-only counter wrapping `Arc<AtomicU64>` for the runtime's view of storage's current session id. Cloneable, with `current()` / `try_advance()` semantics: writers always move the stored value forward via a CAS loop, and reads observe the latest committed value. Lives in `spider-core` so the future scheduler service can reuse the same primitive.

**`spider_execution_manager::liveness`** — a tokio-actor driving the periodic storage heartbeat:

- A `tokio::time::interval` ticks every `heartbeat_interval`; each tick calls `LivenessClient::heartbeat(em_id)` and forwards storage's reply to the shared `SessionTracker`. The interval uses `MissedTickBehavior::Skip` as a defensive guard against starvation-induced burst-replay.
- A `LivenessCommand::Refresh` lets the rest of the runtime ask for an off-schedule heartbeat. The command does **not** advance the tracker directly — storage's heartbeat reply is the only source of truth for the current session id, so the actor always re-checks rather than trusting the caller's observation.
- `interval.reset()` runs at the end of every heartbeat call, so a `Refresh`-triggered heartbeat naturally rate-limits the next scheduled tick. Two consecutive heartbeats are never closer together than `heartbeat_interval`.
- Terminal errors (`MarkedDead` / `IllegalId`) cancel the actor's `CancellationToken`, which the rest of the runtime will observe to tear everything down.
- A storage reply with a session ID *older* than the locally tracked value is treated as a protocol invariant violation and also cancels the runtime.

This PR also creates `test-utils` under `tests/huntsman` to collect all reusable test helpers.

# Checklist

  breaking change.

# Validation performed

* [x] Add unit tests to cover basic actor behaviors.

## Release Notes

* **New Features**
  * Added session tracking to maintain runtime session progression.
  * Introduced an execution manager with liveness heartbeats and automatic shutdown on terminal errors.
  * Added process-pool management to drive external task executors and recover from crashes/timeouts.

* **Tests**
  * Large expansion of end-to-end integration tests covering success, task errors, crashes and timeouts.
  * New shared test utilities and subprocess harness for deterministic integration testing.

* **Chores**
  * Workspace reorganized with additional test and component crates added.


components/spider-execution-manager/src/client/liveness.rs
```diff
@@ -52,6 +52,7 @@ pub trait LivenessClient: Send + Sync {
     /// Returns an error if:
     ///
     /// * [`LivenessResponseError::Transport`] if the connection was lost or timed out.
+    /// * [`LivenessResponseError::IllegalId`] if storage rejected the registration.
```
sitaowang1998: Bad merge?

components/spider-core/src/session.rs
```diff
@@ -0,0 +1,107 @@
+//! Forward-only session tracker shared across services.
```
sitaowang1998: Forward-only is a confusing term. How about monotonic?

components/spider-execution-manager/src/liveness.rs
```diff
@@ -0,0 +1,398 @@
...
+        match self.client.heartbeat(self.em_id).await {
+            Ok(session_id) => {
+                let previous = self.session_tracker.current();
+                if previous != session_id {
+                    if self.session_tracker.try_advance(session_id) {
+                        tracing::info!(
+                            from = previous,
+                            to = session_id,
+                            "Session advanced by heartbeat."
+                        );
```
sitaowang1998, line 139: I think apart from bumping the id internally, we need a way to signal that heartbeat has been updated. Otherwise the user have to poll the session id.

sitaowang1998 (changes requested):

components/spider-execution-manager/src/client/liveness.rs
LinZhihao-723: Yeah, it's a merge mistake. My bad.

components/spider-execution-manager/src/liveness.rs
LinZhihao-723: I'm not sure if I got what u mean.
Which component do we need to signal for?

sitaowang1998: Does execution manager need to stop running tasks when the session id changes? Or should we just let them run and reject them when they finish?

LinZhihao-723: Similar to the process pool design: to keep things simple, we should just let it run and reject when they finish.
Adding a signal protocol is non-trivial, and the benefit is limited: a session bump should be rare in reality.

sitaowang1998 (approved):

PR #329, opened 2026-05-26
feat(spider-execution-manager): Add the runtime that drives the main task-dispatch loop.

LinZhihao-723:
# Description

This PR adds `spider_execution_manager::runtime`, the last piece of the execution-manager: a generic owner of the process pool, liveness actor, and clients that pulls task assignments from the scheduler, dispatches them through the executor pool, and reports outcomes back to storage.

**`Runtime<SchedulerClientType, StorageClientType>`** — generic over the two client types the loop calls every iteration. Storage carries `+ Clone + 'static` because the outcome reporter is spawned onto a detached tokio task and needs an owned client. The liveness client is only generic at the `create` factory (the runtime hands its only `Arc` to the actor at boot and never touches it again).

**Bootstrap (`Runtime::create`)** — registers with storage via `LivenessClient::register`, seeds a `SessionTracker` with the returned session id, builds the initial `ProcessPool`, and spawns the liveness actor. By the time `create` returns, the actor has already sent its first heartbeat and the runtime is ready to serve.

**Main loop (`Runtime::run` / `Runtime::main_loop`)** — each iteration:

1. **Pull an assignment.** `SchedulerClient::next_task` is raced against the cancellation token via `tokio::select!` with `biased`, so shutdown is observed promptly even when the scheduler is idle. Any `SchedulerError` is treated as fatal and propagated via `?` (the client is responsible for absorbing transient transport blips internally; a surfaced error means it's already given up).
2. **Stale-session triage.** The bundle's `session_id` is compared against the shared `SessionTracker::current()`: a strictly-older bundle is logged and dropped; a strictly-newer bundle nudges the liveness actor via `LivenessHandle::refresh` so the tracker re-syncs through the next heartbeat. The tracker is never advanced from the main loop — the liveness actor is the only writer.
3. **Register the task instance.** `StorageClient::register_task_instance` is racing-aware (cancellation interrupts the await). `StaleSession` triggers a liveness refresh and drops the assignment; `InvalidInput` is treated as a contract bug and is fatal; every other storage error is logged and the assignment is dropped.
4. **Execute.** An `ExecuteRequest` is built from the registration's `ExecutionContext` (the `hard_timeout` is derived from `timeout_policy.hard_timeout_ms`) and submitted to `ProcessPool::execute`. The pool already enforces the hard timeout, so the call is awaited to completion — losing the in-flight outcome to a cancellation-race would prevent step 5 from reporting it. Pool `InternalError`s bubble up via `?` (executor binary missing, encode failures, etc.) and bail the runtime.
5. **Report outcome on a background task.** A fire-and-forget `tokio::spawn(report_outcome(...))` decouples result reporting from the next iteration's scheduler poll. The reporter consumes the outcome end-to-end: builds a `Report` (`Success(Option<Vec<u8>>)` or `Failure(String)` with a message derived from the `Outcome` variant), then transfers ownership into `StorageClient::report_task_success` / `report_task_failure`. No retry logic in this layer — same rationale as the scheduler call.

**Shutdown semantics** — `run(self)` wraps `main_loop` with explicit teardown: cancel the token (covering the scheduler-error / process-pool-error exit paths where the loop bailed without external cancellation), await `liveness_join`, then return `result`. A `tokio_util::sync::DropGuard` field guarantees the token is cancelled even when the `run` future itself is dropped at an outer await point.

# Checklist

  breaking change.

# Validation performed

* [x] Add test cases to cover the basic behavior of the runtime.

## Release Notes

* **New Features**
  * Added execution manager runtime with configuration support, task assignment polling, and outcome reporting.
  * Extended scheduler response types to include resource group identification.

* **Tests**
  * Added comprehensive integration test suite for the execution manager runtime with mock clients.
  * Expanded test utilities with mock scheduler and storage implementations for testing.

* **Chores**
  * Updated workspace configuration to include additional test member.


components/spider-execution-manager/src/runtime.rs
```diff
@@ -0,0 +1,480 @@
...
+    pub async fn run(self) -> Result<(), RuntimeError> {
+        tracing::info!(em_id = ? self.em_id, "Runtime main loop starting.");
+        let result = self.main_loop().await;
+        tracing::info!(em_id = ? self.em_id, "Runtime main loop exited. Shutting down.");
+        self.cancellation_token.cancel();
+        if let Err(err) = self.liveness_join.await {
+            tracing::warn!(err = ? err, "Liveness actor task did not exit cleanly.");
+        }
+        result
+    }
```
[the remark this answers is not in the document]
LinZhihao-723: That's an expected behavior.

components/spider-execution-manager/src/runtime.rs
```diff
@@ -0,0 +1,493 @@
...
+                biased;
+                () = self.cancellation_token.cancelled() => return Ok(()),
+                result = self.scheduler_client.next_task(self.em_id) => {
+                    match result {
+                        Ok(assignment) => assignment,
+                        Err(e) => {
+                            tracing::warn!(err = ? e, "Scheduler returned an error. Retrying.");
+                            continue;
+                        }
+                    }
```
sitaowang1998, line 211: Should we add a random sleep right now as the benchmark shows that it is useful under load?

sitaowang1998 (commented):
LGTM generally.

components/spider-execution-manager/src/runtime.rs
LinZhihao-723: We might want to do this on the client side, though. The trait currently only returns if the scheduler actually returns an assignment.

sitaowang1998 (dismissed):

sitaowang1998 (approved):

PR #333, opened 2026-06-03
feat(huntsman): Add protobuf scaffolding and gRPC StorageClient for task instance management:

sitaowang1998:
* Add `spider-proto` component for protobuf source files.
* Add `spider-proto-rust` component for the generated Rust code.
* Add tasks to build the generated Rust code.
* Add a GitHub workflow to verify the generated Rust code is up-to-date.
* Implement a gRPC-backed `StorageClient` for the execution manager.

# Description

This PR:
- Adds a new `spider-proto` component to hold gRPC protocols, with tonic's protobuf compilation in `build.rs`.
- Moves existing execution manager liveness protocol from `spider-execution-manager` to the new component.
- Adds a new task to compile protobuf and a new task to compile non-test rust components.
- Adds GitHub workflow to guard against outdated compiled protobuf.

# Checklist

  breaking change.

# Validation performed

- [x] Protobuf compiles successfully. 
- [x] Existing rust tests pass.
- [x] GitHub workflows pass, with new protobuf check.

## Release Notes

* **New Features**
  * Added gRPC-based task instance management service enabling remote task registration, success reporting, and failure handling with structured error codes for improved reliability.

* **Build & Infrastructure**
  * Implemented automated protobuf code generation validation to ensure generated artefacts remain in sync with schema definitions.


components/spider-proto-rust/build.rs
```diff
@@ -0,0 +1,33 @@
...
+    println!("cargo:rerun-if-env-changed={GENERATED_ENV_VAR}");
+
+    if let Some(generated_dir) = env::var_os(GENERATED_ENV_VAR) {
+        let generated_dir = Path::new(&generated_dir);
+        if generated_dir.exists() {
+            fs::remove_dir_all(generated_dir)?;
+        }
+        fs::create_dir_all(generated_dir)?;
+        compile_protos(generated_dir)?;
+    }
```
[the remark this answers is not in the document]
sitaowang1998: We only support build using build task, where environment variables are set.

LinZhihao-723 (changes requested):
Didin't check the protobuf details yet. Here're some high-level structuring suggestion:
1. I think we should split the protobuf source files and generated files into the following structure:
  ```
  - spider-proto
    - storage.proto
  - spider-proto-rust
    - src
      - lib.rs
      - generated
        - storage.rs
  ```
  - protobuf source files are language-independent; it makes more sense to store all source files outside of a Rust crate.
  - We use `spider-proto-rust` to hold the generated Rust code. If we need the protobuf to be defined in other languages, for example, Go, we could add `spider-proto-go`.
2. I don't think we should move the storage client trait into `spider-proto-rust`. This client trait is specific to a certain component. Its implementation may require the generated protobuf code, but in that case, it should import from `spider-proto-rust` crate and make the implementation local to the use-case-crate.
3. In `storage.proto`: we define the package name to be `spider.storage`, making the generated file `spider.storage.rs`, which doesn't follow the convention for naming a regular source file in Rust (actually, in any common programming language, lol). We should rename the package name to either `spider_storage`, or simply `storage`.

LinZhihao-723 (commented):
Rewrote `build.rs` for `spider-proto-rust` in the latest commit:
- Instead of using a hardcoded relative path, the new version will derive the root of the crate and `spider-proto` at build.rs' ruintime.
- The generated directory should be fixed to be `spider-proto-rust/src/generated`.
- Since the build depends on the generated file, it is impossible to check whether the generated code has changed: they will always be marked dirty (as explained in the script in detail). If we first generate the code, and then remove the generated directory or change anything inside and rerun, it won't work. This is a known issue and we might want to come back and fix it later. @coderabbitai Please create an issue for this.
- We probably don't want the codegen to be triggered every time we build the library (we prefer to use the pre-generated code, committed as a part of `spider-proto-rust`). Therefore, we expose an env var to control the codegen behavior through `SPIDER_PROTO_RUST_GENERATE_FROM_SOURCE`.
- In our Rust build task, it shouldn't rely on the codegen task since we prefer to use the committed code as discussed in the previous bullet point.

LinZhihao-723:
Our offline discussion makes the following conclusions: the best way to solve the code generation problem is to not do it inside cargo's build script; instead, we make a build binary and use a dedicated task to control the codegen. The generated code has a stable destination (in this case, `components/spider-proto-rust/src/generated), we can use task's checksum to determine whether we need to rerun the codgen.
We'd like to defer the implementation of this solution to a future PR. @coderabbitai, can you create an issue to track this?

components/spider-proto/storage/storage.proto
```diff
@@ -0,0 +1,65 @@
+syntax = "proto3";
+
+package storage;
+
+service ExecutionManagerStorage {
```
LinZhihao-723: `ExecutionManagerStorage` -> `ExecutionManagerToStorage` or `FromExecutionManager`?

components/spider-proto/storage/storage.proto
```diff
@@ -0,0 +1,65 @@
...
+  rpc RegisterTaskInstance(RegisterTaskInstanceRequest) returns (RegisterTaskInstanceResponse);
+  rpc ReportTaskSuccess(ReportTaskSuccessRequest) returns (StorageOperationResponse);
+  rpc ReportTaskFailure(ReportTaskFailureRequest) returns (StorageOperationResponse);
+}
+
+message RegisterTaskInstanceRequest {
+  bytes job_id = 1;
+  bytes task_id = 2;
```
LinZhihao-723: We probably need to merge #331 first: currently we have two `TaskId` definitions: one is in the storage, which is the actual definition we use; the other is in `spider-core`, which should be deprecated since it's no longer used.
The actual task ID we use is an enum. We need to define how to serialize that in gRPC.

components/spider-proto/storage/storage.proto
```diff
@@ -0,0 +1,65 @@
...
+}
+
+message ReportTaskSuccessRequest {
+  bytes job_id = 1;
+  bytes task_id = 2;
+  bytes execution_manager_id = 3;
+  uint64 session_id = 4;
+  oneof output_payload {
+    bytes serialized_outputs = 5;
+  }
```
LinZhihao-723: Do we need `oneof`? I think `serialized_outputs` maybe enought.

components/spider-proto/storage/storage.proto
```diff
@@ -0,0 +1,65 @@
...
+}
+
+message StorageOperationResponse {
+  oneof result {
+    Empty ok = 1;
+    StorageError error = 2;
+  }
+}
+
+message Empty {}
```
LinZhihao-723: How about `Void`?

components/spider-proto/storage/storage.proto
```diff
@@ -0,0 +1,65 @@
...
+
+message StorageError {
+  enum Kind {
```
LinZhihao-723: How about `ErrCode`?

LinZhihao-723 (changes requested):
Some comments on the proto definition. Let's have an agreement on these before we proceed.

components/spider-proto/storage/storage.proto
sitaowang1998: As discussed offline, we will name it `TaskInstanceManagementService`. We prefer to name the service based on its functionality, not on the sender and receiver.

LinZhihao-723 (approved):
I'm done with all my editing. Directly modified the PR title to make it a multi-line commit message.

PR #332, opened 2026-06-01
feat(spider-scheduler): Add channel-based dispatch queue implementation.

LinZhihao-723:
# Description

This PR depends on #330 and #331.

This PR implements an async-channel-backed dispatch queue. It also adds unit tests to cover its basic behaviors.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* [x] Ensure all workflows pass.
* [x] Added unit tests to cover the enqueue/dequeue operations, as well as session bumping.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Release Notes

* **Chores**
  * Established new scheduler component crate with configured dependencies
  * Implemented dispatch queue system foundation with comprehensive test coverage
  * Added tests for concurrent operations, load balancing, and session management scenarios
  * Configured asynchronous runtime support for system operations

PR #334, opened 2026-06-05
feat(spider-scheduler): Add round-robin scheduler core implementation.

LinZhihao-723:
# Description

This PR depends on #332.

This PR implements a round-robin scheduling algorithm.

The scheduler operates in discrete, interval-paced ticks. Each tick consumes the results of an asynchronous inbound-queue poll (all three lanes are polled concurrently per round, throttled by the remaining buffer capacities), loads new tasks into the internal buffers, and then makes scheduling decisions until the dispatch queue reaches capacity. Key behaviors:

* **Round-robin fairness**: a fixed-capacity active job pool is scheduled via a rotation of slots; each full cycle dispatches at most one task per active job, plus at most one commit task and one cleanup task.
* **Pending jobs**: jobs beyond the active pool's capacity are buffered in FIFO order and promoted whenever an active job runs out of schedulable tasks.
* **Deduplication**: storage may deliver duplicate task entries; every buffered task is unique, so each task reaches the dispatch queue at most once.
* **Finalizing jobs**: a commit-ready or cleanup-ready entry marks its job as finalizing — the job leaves the active/pending sets, its buffered tasks are discarded, and any further regular tasks for it are ignored. The finalizing gate persists after the finalizing task dispatches and is retired by a (currently hard-coded) 6-hour expiry.
* **Session bumps**: when storage reports a newer session, the scheduler clears all placement state, bumps the dispatch queue's session (draining stale assignments), and drops entries from lanes still reporting an older session.
* **Observability**: `tracing` instrumentation across the scheduling flow (startup/shutdown, session bumps, job state transitions, finalizing-gate changes, and decision-loop summaries).

The module is laid out as `round_robin/{mod.rs, implementation.rs, tests.rs}`, with internals exposed to the sibling test module via `pub(super)` and only `RoundRobinConfig`/`RoundRobinCore` exported.

# Checklist

  breaking change.

# Validation performed

* [x] Add unit tests to cover round-robin behaviors:
  * Black-box tests that assert on the task assignments generated by the scheduling algorithm.
  * White-box tests that assert internal data structures match our expectations.
* [x] Test on AWS with the storage service to make sure the scheduling overhead is negligible in real workloads, and not observable in the task execution critical path.

## Release Notes

* **New Features**
  * Added a new scheduler core abstraction and a round-robin scheduler implementation with configurable queue capacities and timing controls.
  * Introduced session-aware scheduling with commit-ready and cleanup-ready dispatch lanes, pending-to-active promotion, and expiration handling for finalizing work.
  * Exposed the new implementation via the library public API.
* **Bug Fixes**
  * Improved error reporting with additional scheduler error variants for clearer failure modes.
* **Documentation**
  * Added in-code documentation describing fairness, buffering, and scheduling behaviour.
* **Tests**
  * Added comprehensive round-robin scheduling test coverage.


components/spider-scheduler/src/dispatch_queue.rs
```diff
@@ -0,0 +1,668 @@
...
+}
+
+/// A cloneable writer handle for the dispatching queue, implementing [`DispatchQueueSink`] using
+/// an async channel.
+///
+/// # NOTE
+///
+/// The current implementation assumes that `enqueue` and `bump_session_id` will not be called
+/// concurrently: `bump_session_id` must be called before consequent `enqueue` calls to make session
+/// ID consistent with the enqueued assignments.
```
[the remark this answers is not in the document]
LinZhihao-723: This is outside of the current PR's scope. And the design is intended.

components/spider-scheduler/src/core_impl/round_robin/implementation.rs
```diff
@@ -0,0 +1,1117 @@
...
+    /// * Forwards [`Self::retire_active_job`]'s return values on failure.
+    fn enqueue_commit_ready_entries(
+        &mut self,
+        commit_ready_entries: Vec<InboundEntry>,
+    ) -> Result<(), SchedulerError> {
+        for inbound_entry in commit_ready_entries {
+            if !self
+                .buffered_tasks
+                .insert((inbound_entry.job_id, inbound_entry.task_id))
+            {
```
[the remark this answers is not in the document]
LinZhihao-723: That is expected.

components/spider-scheduler/src/core_impl/round_robin/implementation.rs
```diff
@@ -0,0 +1,1117 @@
...
+    ///   * `dispatch_queue_capacity`
+    ///   * `ready_task_capacity`
+    ///   * `commit_ready_task_capacity`
+    ///   * `cleanup_ready_task_capacity`
+    pub fn make_core<
+        SchedulerStorageClientType: SchedulerStorageClient + 'static,
+        DispatchQueueSinkType: DispatchQueueSink,
+    >(
+        self,
+    ) -> Result<RoundRobinCore<SchedulerStorageClientType, DispatchQueueSinkType>, SchedulerError>
```
sitaowang1998, line 93: Making core's constructor a method of the config is weird.

components/spider-scheduler/src/core_impl/round_robin/implementation.rs
```diff
@@ -0,0 +1,1117 @@
...
+    /// than 6 hours. This timeout is currently hard-coded but may be made configurable through
+    /// [`RoundRobinConfig`] in the future.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`SystemTime::elapsed`]'s return values on failure.
+    fn retire_expired_finalizing_jobs(&mut self) -> Result<(), SchedulerError> {
+        const EXPIRATION_TIME: Duration = Duration::from_hours(6);
```
sitaowang1998: Let's make this configurable.

LinZhihao-723 (commented):
Based on the offline discussion with @sitaowang1998, the following changes are made as a part of code review actions:
- Rename `active_job_queue` to `rr_queue` (round-robin queue) for clarity, since it contains round-robin slots for commit-ready and cleanup-ready tiers as well.
  - Use `rr_` instead of `round_robin_` to fix the clippy error in terms of replicating the struct name at the beginning of a member variable name.
- Make each commit-ready round-robin slot enqueues up to # active jobs of commit tasks: this mirrors the fairness bound for active jobs (in an imperfect way).
- Update the mod-level docstring with a limitation section, referring to issues #344 and #345.

components/spider-scheduler/src/core_impl/round_robin/implementation.rs
LinZhihao-723: hmmm, why is this weird?

The pattern I desire is to make all supported core configs serde-compatible and make an enum of them. The enum will replicate the `make_core` to convert the given config (regardless of the underlying specified core) into `dyn SchedulerCore` without knowing what algorithm it is. That provides the scheduler binary a clean way to:
- Let the user (CLP) specify what scheduling algorithm to use through the config.
- Start with the core without knowing what the core is, as long as it's supported.

In this way, it implicitly binds the corresponding core type with the config, so any use case of the config doesn't need to explicitly know which core implementation should be created from this config: you could make `RoundRobinCore::new(RoundRobinConfig)`, but you need to explicitly know `RoundRobinConfig` maps to `RoundRobinCore`.

Btw this is a common practice. Some examples:

- sqlx's config options exposes `connect` method to convert the config into the actual connection: https://docs.rs/sqlx/latest/sqlx/trait.ConnectOptions.html#tymethod.connect
- tokio runtime's builder exposes `build` method to construct the runtime from the config: https://github.com/tokio-rs/tokio/blob/7892f6020d9c914a41d0c350693fb71937d43c03/tokio/src/runtime/builder.rs#L1062

The difference between these examples and `make_core` is that `make_core` takes the ownership of the config, making it not reusable, which is on purpose since in our use case, the core should be constructed only once.

Let me know if you have other concerns.

sitaowang1998 (approved):

PR #339, opened 2026-06-08
feat(spider-storage): Add support for recovering job cache on storage service restart.

sitaowang1998:
# Description

This PR adds recovery of jobs in `Running`, `CommitReady` and `CleanupReady` states when the storage runtime restarts. All recoverable jobs are fetched from the database and filled into the cache layer and the ready queue.

# Checklist

  breaking change.

# Validation performed
- [x] Tests added that restarts the storage runtime.
- [ ] GitHub workflows pass.

## Release Notes

* **New Features**
  * Added execution manager runtime for per-node task coordination and lifecycle management
  * Implemented task executor subprocess with framed binary communication protocol
  * Added job recovery from persistent storage on runtime startup
  * Introduced session tracking for improved request reliability
  * Added gRPC storage client for task instance management

* **Bug Fixes & Improvements**
  * Updated ID system for improved performance and consistency
  * Enhanced workflow checks to validate generated protocol code integrity
  * Improved development setup with modular Ubuntu installation scripts

* **Testing**
  * Added comprehensive integration test suite for execution manager runtime
  * Added integration tests for task executor subprocess communication
  * Added process pool recovery and timeout handling tests
  * Added job recovery validation tests

* **Chores**
  * Updated Cargo workspace to include new components and test crates
  * Enhanced build tasks for protocol code generation and multi-threaded testing
  * Reorganized development dependency installation scripts


components/spider-storage/src/cache/job.rs
```diff
@@ -47,6 +47,20 @@ pub struct SharedJobControlBlock<
         Arc<JobControlBlock<ReadyQueueSenderType, DbConnectorType, TaskInstancePoolConnectorType>>,
 }
 
+/// Persistent job state used to recover a job control block.
```
LinZhihao-723: Why is the struct again defined in the middle of nowhere?
Ask your coding agent to enforce the rule: https://github.com/LinZhihao-723/claude-instruction/blob/main/rust/coding-style.md#symbol-ordering

components/spider-storage/src/cache/task.rs
```diff
@@ -172,6 +172,30 @@ impl TaskGraph {
...
+    ) -> Result<(), InternalError> {
+        if persisted_outputs.len() != self.outputs.len() {
+            return Err(InternalError::TaskOutputsLengthMismatch(
+                self.outputs.len(),
+                persisted_outputs.len(),
+            ));
+        }
+        for (output_reader, output) in self.outputs.iter().zip(persisted_outputs) {
+            *output_reader.writer().write().await = Some(output);
+        }
```
LinZhihao-723: Have you read this change carefully (assume it's done by your coding agent)?
The purpose of having the reader/writer wrapper around the RW lock is to split read/write access. If we need to mutate the output, we should design a new API, instead of performing write through a "reader".

components/spider-storage/src/db/protocol.rs
```diff
@@ -11,6 +11,23 @@ use spider_core::{
...
+    pub id: JobId,
+    /// The owning resource group.
+    pub resource_group_id: ResourceGroupId,
+    /// The source-of-truth database state.
+    pub state: JobState,
+    /// The original job submission.
+    pub job_submission: ValidatedJobSubmission,
+    /// The committed job outputs, if the job has reached the commit phase.
+    pub job_outputs: Option<Vec<TaskOutput>>,
+}
```
LinZhihao-723, line 23: What's the difference between this struct and `JobRecoveryContext`?

LinZhihao-723 (changes requested):

components/spider-storage/src/cache/task.rs
sitaowang1998: This is actually after my fight with codex. The original is to directly add a `Vec<OutputWriter>` to the `TaskGraph`. What do you propose for the new API?

LinZhihao-723 (commented):
Reviewed with the following changes:

## DB changes

- `RecoverableJob` -> `RecoverableJobContext`
  - Removed field level comments (looks redundant, the field names are self-explanatory)
  - Drop `job_` prefix from field names
- Index jobs on their states in the DB table.
- Add a struct `RecoverableJobRowProjection` to make `get_recoverable_jobs` more readable.

## Cache changes

- We shouldn't send task-ready in JCB recovery: the service is not runnable yet; the task-ready can fill the underlying channel, making a deadlock on startup.
- No alias on `UnexpectedJobState`: we should be explicit whether this error is internal or not.
  - This is my fault from the original JCB PR.
- Add a dedicated error code `InvalidRecoverableJobContext`.
- The major design issue with https://github.com/y-scope/spider/pull/339#discussion_r3383142317 is that we shouldn't mutate the task graph after construction; instead, we should make dedicated factory methods for the cases we need.
  - Optimized a bit further by making the task graph empty for commit-ready and cleanup-ready; there's no need to reload the task graph: these tasks should be irrelevant anyways.

## Service changes

- Add `resend_ready_tasks` in the service level.
  - Will be needed by the scheduler and the recovery step.
  - Tested in the recovery tests.

## Tests

- Change the file name to `runtime_recovery_test.rs`.
- Some style-wise refactoring.
- For `restarted_storage_cache_recovers_commit_ready_job` and `restarted_storage_cache_recovers_cleanup_ready_job`: make sure that the regular task registration would fail with the error code showing the job is already in a terminal state.
- Make `find_entry_for_job` return a vector of ready queue entries, and make assertions outside.
- For `restarted_storage_cache_recovers_running_job_from_start` and `restarted_storage_cache_recovers_commit_ready_job`, create a new runtime in the end and make sure the job state and outputs are persisted (accessable via `service`, through DB operations).
- Add `ignore` section to tests since MariaDB is required.

# Others

- Remove `uuid` dependency from `spider-storage`'s `Cargo.toml`.

---

Two more questions for you:

- Why do we not add `Ready` jobs to the cache? How are we supposed to handle it if the user registers a job and then calls start, but the service crashed and restarted in between?
- Currently, recovery tests are running together with other tasks that may produce jobs in the DB. These jobs might also be recovered, and if that happens, polling 32 items from the ready queue may not be able to guarantee to retrieve the expected entry. My solution is to apply `#[serial_test::file_serial]` to all tests inside `runtime_recovery_test.rs` and `jcb_test.rs`.

components/spider-storage/src/state/runtime.rs
LinZhihao-723, line 1: Why do we need unit tests in this file? `runtime_recovery_test.rs` does the same job, right?

components/spider-storage/tests/runtime_recovery_test.rs
```diff
@@ -45,14 +45,10 @@ async fn restarted_storage_cache_does_not_recover_ready_job() -> anyhow::Result<
...
-    assert!(
-        matches!(start_result, Err(StorageServerError::JobNotFound(id)) if id == job_id),
-        "ready job should not be recovered into cache"
-    );
+    recovered_service.start_job(job_id).await?;
     assert_eq!(
         recovered_service.get_job_state(job_id).await?,
-        JobState::Ready
+        JobState::Running
     );
```
LinZhihao-723, line 52: Probably worth replicating the steps in `restarted_storage_cache_recovers_running_job_from_start` to make sure the job can terminate with all tasks runnable.

LinZhihao-723 (changes requested):

LinZhihao-723 (commented):
- Sorry for not being clear in [this](https://github.com/y-scope/spider/pull/339#discussion_r3398708424) comment: we could keep the existing unit tests; but the read job recovery unit tests are redundant.
- `succeed_ready_job` is confusing: the job is not in "ready"; it should be in the running state. Renamed it for clarity.

LinZhihao-723 (approved):

PR #340, opened 2026-06-10
feat(huntsman): Complete storage gRPC protocol and its clients:

sitaowang1998:
* Add protobuf definitions for the storage gRPC protocol.
* Implement the storage client for `spider-scheduler`.
* Implement the liveness client for `spider-execution-manager`.

# Description

This PR:
* Adds remaining gRPC protocol for storage interface.
* Adds task id and job state conversion between gRPC and `spider-core`.
* Adds gRPC implementation of execution manager's liveness client.
* Adds gRPC implementation of scheduler's storage client.

# Checklist

  breaking change.

# Validation performed

* [x] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Added gRPC services for job orchestration, inbound task polling, resource-group management, execution-manager liveness (register/heartbeat) and session retrieval.
  * Introduced gRPC-backed clients for liveness and scheduler storage to enable remote polling and job-state operations.

* **Error Handling**
  * Replaced generic storage errors with domain-specific error types and richer error mappings for clearer failure categories.

* **Tests**
  * Added unit tests covering protocol conversions, response/error mapping, registration and heartbeat flows.


LinZhihao-723 (changes requested):
Reviewed all changes and applied the following code-level changes:


# proto

- `JobManagementService` -> `JobOrchestrationService` to match the protocol trait's name
  - `JobManagementError` -> `JobOrchestrationError`
- `TaskInstanceError` -> `TaskInstanceManagementError`. `TaskInstanceError` sounds like the error inside the task instance itself.
- `ResourceGroupError` -> `ResourceGroupManagementError`.
- `serialized_inputs` should be `bytes`, not `repeated bytes`. The inputs should be submitted in a serialized form, which allows the client and server to determine how the bytes should be decoded.
- `SchedulerStorageService` -> `InboundQueueService` to match the term we use in the scheduler.
  - `SchedulerStorageError` -> `InboundQueueResponseError`
- Remove `IsExecutionManagerAlive` and `GetDeadExecutionManagers` from `ExecutionManagerLivenessService`: these operations are internal to the storage service, we don't need gRPC methods for them.
- Change `PollReadyTaskRequest`'s `wait_ns` to `wait_ms`: None of our clients use nanoseconds.

# spider-proto-rust

- Create a dedicated error type under `error.rs` mod, export it through the library layer. Use this error type for `id.rs` and `job.rs` to replace the use of string errors.
  - Client implementation should be changed accordingly.

# spider-execution-manager

- `storage_error_to_liveness_error` should be renamed to reflect the error type renaming.
- Use `impl From<storage::ExecutionManagerLivenessError> for LivenessResponseError` to replace the use of `storage_error_to_liveness_error`. This was done in the storage client in the previous PR.
- Use constants to replace magic numbers/strings in the test cases.

# spider-scheduler

- Restructure the client mod in the following structure (to match the execution manager trait):
  - storage_client
    - mod.rs (containing the client trait)
    - grpc.rs (containing the gPRC client implementation)
- Drop `inbound_queue_response_error_to_client_error`.
- Doesn't make sense to return `Transport` error for request out-of-range in `poll_ready_task_request`. Use `InvalidInput` instead.
- Docstrings are not inconsistent with other `_to_result` methods in `spider-execution-manager`.
- The session tracker doesn't make any sense: no methods will update it, and used for all requests. If we look at the storage service in the [dev branch](https://github.com/y-scope/spider/blob/61e4a49f5cd157d1a6929b046ce73aedb507d86f/components/spider-storage/src/state/service.rs#L229), retrieving the job state doesn't need the session ID. Thus, the session ID is dropped from the proto source; the session tracker is removed from the client.
- Use constants to replace magic numbers/strings in the test cases.

---

Besides some naming issues, I need to point out that some changes above are **not supposed to be my job** for reviewing this PR:

* Considering we already have a gRPC PR reviewed and merged, I'd assume you should read my changes and make sure these changes are propagated to the consequent PRs. But part of my editing of this review is still repeating what I've done in the last PR.
* Some of the code generated by the coding agent doesn't make sense. You probably need to spend more time reviewing them yourself.

Please review my changes and be careful with the pending gRPC PR for the scheduler-em communication.

LinZhihao-723 (approved):
For the PR title, how about:
```
 feat(huntsman): Complete storage gRPC protocol and its clients

* Add protobuf definitions for the storage gRPC protocol.
* Implement the storage client for `spider-scheduler`.
* Implement the liveness client for `spider-execution-manager`.
```

PR #341, opened 2026-06-12
feat(spider-storage): Add background garbage collection to remove expired terminated jobs from the cache.

sitaowang1998:
# Description

This PR adds storage cache garbage collection. The garbage collection works as follow:
1. When a job terminates, push the job id to an unbounded buffer as a queue.
2. The gc actor polls from buffer and push the job id into the queue with a timestamp.
3. The gc actor periodically runs gc cycle to remove all jobs in the queue with timestamp before the configured retention period.

# Checklist

  breaking change.

# Validation performed

- [x] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Added execution manager runtime for coordinating task execution across distributed processes.
  * Added liveness heartbeat system for execution manager registration and monitoring.
  * Added automatic job recovery capability on storage service restart.
  * Implemented job cache garbage collection for terminated job cleanup.
  * Added gRPC-based storage service interface for task instance management.

* **Architecture Changes**
  * Migrated task identifiers from UUID to numeric format.
  * Changed task instance inputs representation to serialized bytes.
  * Added session tracking for distributed state coordination.

### DIFF
--- a/components/spider-storage/src/cache/error.rs
+++ b/components/spider-storage/src/cache/error.rs
@@ -81,6 +81,9 @@ pub enum InternalError {
     #[error("invalid config: {0}")]
     TaskInstancePoolInvalidConfig(&'static str),
 
+    #[error("invalid config: {0}")]
+    JobCacheGcInvalidConfig(&'static str),
+
     #[error("ready queue channel is closed")]
     ReadyQueueChannelClosed,
 
--- a/components/spider-storage/src/cache/error.rs
+++ b/components/spider-storage/src/cache/error.rs
@@ -75,13 +75,13 @@ pub enum InternalError {
     #[error("task instance pool corrupted: {0}")]
     TaskInstancePoolCorrupted(String),
 
-    #[error("invalid config: {0}")]
+    #[error("invalid ready-queue config: {0}")]
     ReadyQueueInvalidConfig(&'static str),
 
-    #[error("invalid config: {0}")]
+    #[error("invalid task instance pool config: {0}")]
     TaskInstancePoolInvalidConfig(&'static str),
 
-    #[error("invalid config: {0}")]
+    #[error("invalid job cache GC config: {0}")]
     JobCacheGcInvalidConfig(&'static str),
 
     #[error("ready queue channel is closed")]
--- a/components/spider-storage/src/state.rs
+++ b/components/spider-storage/src/state.rs
@@ -1,10 +1,12 @@
 pub mod error;
 pub mod job_cache;
+pub mod job_cache_gc;
 pub mod runtime;
 pub mod service;
 
 pub use error::StorageServerError;
 pub use job_cache::JobCache;
+pub use job_cache_gc::{JobCacheGcConfig, JobCacheGcHandle, create_job_cache_gc};
 pub use runtime::{Runtime, create_runtime};
 pub use service::ServiceState;
 
--- a/components/spider-storage/src/state/job_cache.rs
+++ b/components/spider-storage/src/state/job_cache.rs
@@ -108,6 +108,19 @@ impl<
         self.jobs.write().await.remove(&job_id)
     }
 
+    /// Removes multiple job control blocks from the cache.
+    ///
+    /// # Returns
+    ///
+    /// The number of job control blocks that existed and were removed.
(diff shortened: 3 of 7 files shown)


LinZhihao-723 (changes requested):
Reviewed with the requested changes.

There is still a gap between the current state of this PR and what we would expect from a ready-to-review PR. Most of these issues should be catchable by the coding agent, since I've used it to catch issues such as incorrect symbol ordering or outdated docstring updates.

It may be helpful to compile a list of common issues found in previously reviewed PRs and ask the agent to perform an additional pass against that checklist before submitting future PRs for review.

# Job cache GC implementations

- For the GC handle, I'm not sure if it's the coding agent decision or yours, but I don't think it makes sense to.
  - We should not:
    - Have the GC handle optional: in reality, we shouldn't need to support the GC not set up.
    - Have the GC handle to return the job ID on error: the error reporting is meaningless; if the GC actor fails, the cancellation token should be fired; printing the error message is not helpful. It also doesn't make sense to return the job ID as the error type.
  - I'm not sure if this is your decision or not: having both `new` and `new_with_job_cache_gc` makes no sense except it keeps the existing tester code untouched. But it turned out that it doesn't need too many lines of change to update the existing code and feed the tester with a handle created explicitly from a channel (which the `job_cache_gc.rs`'s tester does).
  - Changes:
    - Make the GC handle required, make only one `new` method that enforces the GC handle to be given.
    - Make the GC handle fire-and-forget: if the channel has been closed, just leave it. If the GC coroutine is corrupted, it should be handled by calling the cancellation token to shut down the service. A job that is already terminated should pass through.
- This is relatively minor: `enqueue_terminated_job` smells too AI: if we need a method to abstract the enqueue method, the method should probably just take the job ID and handles the enqueue time inside the call. Taking both the job ID and the enqueue time to make it simply a wrapper of `VecDeque::push_back` doesn't help: should be just removed and replaced by an inline push_back method.
  - I kept the `enqueue_terminated_job` method but remove the `enqueue_at` parameter. The enqueue time is now determined by the method itself.
- Also minor: adding `baised` to the select statement to prefer the GC cycle execution.
- `create_job_cache_gc` is in a wrong place as it's ordered after the private symbols. It's tuple-return docstring doesn't match our guideline either.
- Naming inconsistency: the struct to hold the job metadata is `TerminatedJob`, but the actual queue is called `pending_jobs`. Removed to `terminated_jobs` to match the terminology.
- Update the logging for job retirement: we log the operation by the job ID to removal for a better tracking purpose. If it becomes a bottleneck we may switch this print to debug level.

# job_cache.rs

- In `job_cache.rs`, the new "types" should be defined in the private symbols section. I also don't think you should manually implement `Clone` trait -> derive from `Clone` is the normal practice. **Check the agent-generated code before asking for a review. These errors are (or should be) obvious if you skim the diff.**
- Minor: when reading the name `job_cache_remove_batch_removes_existing_jobs_once`, it looks like "once" indicates the test case feeds the removal batch with duplicate job IDs, but it doesn't. Updated the test case to duplicate the job ID and make sure the removal works as expected.

# service.rs

- In `service.rs`'s tester, `cancel_job_enqueues_terminal_job_for_cache_gc` uses `recv` to wait for the job to be enqueued. This works fine for now, but if the test fails, for example, the internal service never enqueues the terminated job, that recv will be just blocked forever. The right thing is to use `try_recv` and don't block.

# runtime.rs

- In `wait_for_background_task`, it still returns `InternalError::TaskInstancePoolCorrupted` while it's now supposed to be a generic helper for waiting a background task.
- I found the current runtime shutdown doesn't make sense when we need to handle multiple background tasks: the stop API should only tell whether the stop operation is successful or not; it should not return the error of the background tasks: these tasks should have their own handler for making their failure visible (by firing the cancellation token). Inside the stop, we should only log the error if any.
  - I also restructured the stop layout with `tokio::join!` macro and the timeout API.
- Docstring of `create_runtime` is not updated.
- The signature of `create_runtime` is not updated: it should accept the job cache GC config, not using the default.
  - For better integration with the coming service executable, I grouped all config needed into a struct `RuntimeConfig`. We could add serde support to it later for reading the config from a file/JSON string.

# Others

- Just realized in the cache error, we have `invalid config` context for three different errors. Add the config name details to differentiate them.

PR #342, opened 2026-06-14
feat(huntsman): Add protobuf defition and Rust binding for scheduler service; Implement gRPC scheduler client for execution manager.

sitaowang1998:
# Description

This PR:
- Adds gRPC proto for scheduler service for execution manager.
- Moves shared proto types to common proto file.
- Adds gRPC scheduler client in execution manager.

# Checklist

  breaking change.

# Validation performed

- [x] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Added scheduler gRPC support, including heartbeat and shutdown.
  * Extended task polling to include previous assignment context.
* **Changes**
  * Introduced shared protobuf types for task IDs and empty responses, applied across scheduler and storage APIs.
  * Expanded scheduler assignment responses to include scheduler ID and session details for follow-up storage calls.
* **Tests**
  * Updated conversion/packing coverage for the new protobuf contracts and response unpacking.
* **Chores**
  * Regenerated protobufs to include the new common and scheduler definitions.


components/spider-proto/scheduler/scheduler.proto
```diff
@@ -0,0 +1,46 @@
+syntax = "proto3";
+
+package scheduler;
+
+import "common/common.proto";
+
+service SchedulerService {
+  rpc NextTask(NextTaskRequest) returns (NextTaskResponse);
```
LinZhihao-723, line 8: The waiting time is missing. The gRPC call should take a waiting time that the request can be blocked in the scheduler for how long, like we did in the benchmark.
This parameter was not explicitly exposed in the execution manager because the execution manager will need to pool the scheduler infinitely to get a task assignment anyway. This waiting time should be a client option given at construction time.

components/spider-proto-rust/src/unpack/mod.rs
```diff
@@ -31,3 +43,23 @@ pub trait RequestUnpack {
...
+    ///
+    /// # Returns
+    ///
+    /// The unpacked response on success.
+    ///
+    /// # Errors
+    ///
+    /// Returns the failure message on failure.
+    fn unpack(self) -> Result<Self::Unpacked, String>;
+}
```
LinZhihao-723: I don't think this trait makes sense: this is literally just the standard `TryFrom` trait, right? We already have a couple of examples:
- https://github.com/y-scope/spider/blob/284052adcafc29e64b43a0b5f7ed9604f65a7b0a/components/spider-proto-rust/src/job.rs#L21
- https://github.com/y-scope/spider/blob/284052adcafc29e64b43a0b5f7ed9604f65a7b0a/components/spider-proto-rust/src/io.rs#L10

components/spider-proto-rust/build.rs
```diff
@@ -10,7 +10,11 @@ const SPIDER_PROTO_RUST_GENERATED_DIR: &str = "src/generated";
 const SPIDER_PROTO_ROOT: &str = "spider-proto";
 
 /// The protobuf source files to compile, relative to [`SPIDER_PROTO_ROOT`].
-const SPIDER_PROTO_SOURCE_FILES: &[&str] = &["storage/storage.proto"];
+const SPIDER_PROTO_SOURCE_FILES: &[&str] = &[
+    "common/common.proto",
+    "storage/storage.proto",
+    "scheduler/scheduler.proto",
```
LinZhihao-723: Ordering?

components/spider-execution-manager/src/client/grpc/scheduler.rs
```diff
@@ -0,0 +1,123 @@
...
+///
+/// The protobuf representation of `record`.
+const fn task_assignment_record_to_protocol(
+    record: TaskAssignmentRecord,
+) -> scheduler::TaskAssignmentRecord {
+    scheduler::TaskAssignmentRecord {
+        id: record.id.get(),
+        from: record.from.get(),
+    }
+}
```
LinZhihao-723: - Can we implement `From` for `scheduler::TaskAssignmentRecord`?
- Why named "protocol"?

components/spider-execution-manager/src/client/grpc/scheduler.rs
```diff
@@ -0,0 +1,123 @@
...
+                prev_assignments: prev_assignments
+                    .into_iter()
+                    .map(task_assignment_record_to_protocol)
+                    .collect(),
+            })
+            .await
+        {
+            tracing::warn!(
+                em_id = ?em_id,
+                error = ?error,
```
LinZhihao-723: As we always did in other logging messages, there should be a space between `?`/`%` and the actual value being logged.

LinZhihao-723 (changes requested):

components/spider-proto-rust/src/unpack/scheduler.rs
LinZhihao-723, line 1: Have you read the example I gave in the previous? https://github.com/y-scope/spider/pull/342#discussion_r3478695903
Answer the following questions:
- Why do these helpers still land in the `unpack` mod while the ones I gave in the example don't?
- Why, using a string to report errors instead of using the error enum in the crate like the other examples do?
- Why use the request unpacking pattern to implement a response transformation? In particular:
  - Why logging?
  - Why call an unpack function that returns a lightweight gRPC status on error? Shouldn't we use [this](https://github.com/y-scope/spider/blob/284052adcafc29e64b43a0b5f7ed9604f65a7b0a/components/spider-proto-rust/src/id.rs#L23) helper? Like, do you think it makes sense to yourself to call a method that returns a server-specific error code on the client side?

I think there are two problems here:

- The current change looks like feeding my comments into the coding agent and blindly committing whatever it changes. That's not what I would expect.
- The above example helpers have been there for a while. The reason I asked you to carefully read what I've done in previous PRs was to avoid the current situation: you should understand the coding choices I made in those PRs, and follow the same standard in your PRs as well.

tests/huntsman/em-runtime/tests/test_runtime.rs
```diff
@@ -128,6 +128,7 @@ fn runtime_config(heartbeat_interval: Duration) -> RuntimeConfig {
         em_ip: "127.0.0.1".parse().expect("parse loopback"),
         heartbeat_interval,
         scheduler_heartbeat_interval: heartbeat_interval,
+        scheduler_poll_wait: HEARTBEAT_INTERVAL,
```
LinZhihao-723: Do you think it makes sense if you ever read this line to see a heartbeat interval assigned to a polling wait?

components/spider-execution-manager/src/runtime.rs
```diff
@@ -264,7 +270,11 @@ impl<
                 () = self.cancellation_token.cancelled() => return Ok(()),
                 result = self.scheduler_client.next_task(
                     self.em_id,
-                    self.prev_assignments.pop_front()
+                    self.prev_assignments.pop_front(),
+                    self.scheduler_poll_wait
+                        .as_millis()
+                        .try_into()
+                        .expect("scheduler poll wait fits in a u64 millisecond budget"),
```
LinZhihao-723: Do you think a fatal error here makes sense, while you lost all the cleanup and observability logging on top of `main_loop`?

LinZhihao-723: I doubt we should store a duration in the runtime config:
- The interval variables are stored in duration forms because they're directly used by the runtime.
- This polling timeout is not. It should be just a u64 passed all the way to the gRPC.

LinZhihao-723 (changes requested):

LinZhihao-723 (commented):
Changes in https://github.com/y-scope/spider/pull/342/changes/38d0fd1bc2202faa501b67e0d669e9ce2d75a5d2: when a new error enum is added, it should follow the convention as the other enums:
- I'm ok with `#[error("next task response is missing its result")]`, but still prefer to be consistent with others.
- But `#[error("scheduler assignment is missing its task id")]` is wrong. The error enum itself is general to a missing task ID. Task ID may be carried by many responses, not just the scheduler assignment. Thus, the error message should not include `scheduler assignment`.

LinZhihao-723 (approved):
Directly modified the PR title. Notice that `spider-grpc` is not a valid field name since it doesn't correspond to any components in the system.

PR #346, opened 2026-06-16
feat(huntsman): Add the skeleton for gRPC storage server.

sitaowang1998:
# Description

This PR:
* Removes two unused RPCs from the gRPC protocol.
* Adds `task_instance_id` to task result and error report gRPC protocol.
* Updates execution manager to accommodate the changed protocol.
* Adds gRPC service layer skeleton.
* Adds gPRC server binary with cli.

> [!Warning]
> The resource group deletion is not implemented in the MariaDB storage layer.

# Checklist

  breaking change.

# Validation performed

* [ ] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Added gRPC server for storage service with support for job orchestration, task management, resource groups, and scheduler registration.
  * Configuration now supports YAML-based setup with network endpoint and runtime settings.
  * Enhanced task reporting with improved instance tracking.

* **Removals**
  * Removed `DeleteExpiredTerminatedJobs` and `ResendReadyTasks` endpoints.
  * Removed `DeleteResourceGroup` endpoint.
  * Simplified job submission by removing session parameter.

### DIFF
--- a/components/spider-proto/storage/storage.proto
+++ b/components/spider-proto/storage/storage.proto
@@ -150,6 +150,7 @@ message ReportTaskSuccessRequest {
   uint64 execution_manager_id = 3;
   uint64 session_id = 4;
   bytes serialized_outputs = 5;
+  uint64 task_instance_id = 6;
 }
 
 message ReportTaskFailureRequest {
@@ -158,6 +159,7 @@ message ReportTaskFailureRequest {
   uint64 execution_manager_id = 3;
   uint64 session_id = 4;
   string error_message = 5;
+  uint64 task_instance_id = 6;
 }
 
 message AddResourceGroupRequest {
--- a/components/spider-proto/storage/storage.proto
+++ b/components/spider-proto/storage/storage.proto
@@ -9,9 +9,6 @@ service JobOrchestrationService {
   rpc GetJobState(JobIdRequest) returns (JobStateResponse);
   rpc GetJobOutputs(JobIdRequest) returns (JobOutputsResponse);
   rpc GetJobError(JobIdRequest) returns (JobErrorResponse);
-  rpc DeleteExpiredTerminatedJobs(DeleteExpiredTerminatedJobsRequest)
-      returns (DeleteExpiredTerminatedJobsResponse);
-  rpc ResendReadyTasks(ResendReadyTasksRequest) returns (JobManagementOperationResponse);
 }
 
 service TaskInstanceManagementService {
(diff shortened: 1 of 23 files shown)


LinZhihao-723 (changes requested):
Partial review results for the bin entrypoint:
- Instead of exposing everything as cli arguments and reconstruct the runtime config, we should probably start with CLP-package style configuration: we make the runtime config serde-compatible, with all subconfig being the fields inside `RuntimeConfig`. We will need an extra level of `ServerConfig`, containing the details of where the gprc server will be hosted. The cli should only parser the path of this config to start the service.
  - Pending: we may need to discuss with the cloud team to understand their current way for passing the config to the hosting container.
  - For now, the DB credentials are still passed explicitly as parameters. We might need to use environment variables as what CLP-package does.
- The logger is not properly configured: it's missing the basic setup and the unblocking writer. Made a helper for it (and we probably want to make a dedicated crate for the bin helpers outside of `spider-core` lib, maybe called `spider-utils`, to make the helper reusable for other bins).

LinZhihao-723 (changes requested):
Changes:

- Removed session ID from job submission request: the user-facing API doesn't need to expose the session ID.
- As discussed offline, the previous implementation has messy error handling and logging setup. I have removed the service implementation and will resume it in the next few PRs.
- Move `logging` to a new crate `spider-utils` for sharing across different components.
- Make the service listen to the cancellation token: if the token is cancelled, we should stop the service. This allows the error handler to fire the cancellation token on internal errors and thus gracefully terminate the runtime.

Since we removed the service implementation, for the PR title, how about:

```
feat(huntsman): Add the skeleton for gRPC storage server.
```

components/spider-proto/storage/storage.proto
```diff
@@ -9,9 +9,6 @@ service JobOrchestrationService {
   rpc GetJobState(JobIdRequest) returns (JobStateResponse);
   rpc GetJobOutputs(JobIdRequest) returns (JobOutputsResponse);
   rpc GetJobError(JobIdRequest) returns (JobErrorResponse);
-  rpc DeleteExpiredTerminatedJobs(DeleteExpiredTerminatedJobsRequest)
-      returns (DeleteExpiredTerminatedJobsResponse);
-  rpc ResendReadyTasks(ResendReadyTasksRequest) returns (JobManagementOperationResponse);
```
sitaowang1998, line 14: @coderabbit `ResendReadyTask` should not be delete, same as the `DeleteResourceGroup` below. However, we might need to put `ResendReadyTask` in another trait later, as it is used by scheduler, while other functions in this trait are used by client.

sitaowang1998: Please open an issue to track this.

sitaowang1998: @coderabbit Please open an issue to track this

LinZhihao-723 (approved):

PR #350, opened 2026-06-18
feat(spider-execution-manager): Update `Runtime` implementation to track the consumed task assignment records.

LinZhihao-723:
# Description

This PR updates the implementation of `Runtime` to reflect the latest changes in our em-to-scheduler protocol: the runtime now keeps track of a queue of previously consumed task assignments. These records are reported back to the scheduler on every task dispatch or on exit. The runtime also sets up a background task to send heartbeats to the scheduler periodically.

# Checklist

  breaking change.

# Validation performed

* [x] Update unit tests to make sure that on a graceful termination, the scheduler should receive feedbacks of all assigned tasks.
* [x] Add new unit test cases to make sure if the runtime is cancelled before a the storage responses for task registration, the task won't be marked as consumed.

## Release Notes

* **New Features**
  * Enhanced task assignment lifecycle tracking and acknowledgement
  * Added scheduler heartbeat monitoring for improved system coordination

* **Improvements**
  * Refined error handling and reporting during task registration
  * Optimized shutdown procedure with better resource cleanup

* **Tests**
  * Expanded test coverage for heartbeat synchronization and cancellation scenarios

### DIFF
--- a/components/spider-execution-manager/src/runtime.rs
+++ b/components/spider-execution-manager/src/runtime.rs
@@ -7,9 +7,10 @@ use spider_core::{
     types::{
         id::{ExecutionManagerId, JobId, SessionId, TaskId},
         io::ExecutionContext,
+        scheduler::TaskAssignmentRecord,
     },
 };
-use tokio::task::JoinHandle;
+use tokio::{select, task::JoinHandle};
 use tokio_util::sync::{CancellationToken, DropGuard};
 
 use crate::{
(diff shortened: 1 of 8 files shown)


components/spider-execution-manager/src/runtime.rs
```diff
@@ -196,14 +257,17 @@ impl<
...
+    async fn main_loop(&mut self) -> Result<(), RuntimeError> {
         loop {
-            let assignment = tokio::select! {
+            let response = tokio::select! {
                 biased;
                 () = self.cancellation_token.cancelled() => return Ok(()),
-                result = self.scheduler_client.next_task(self.em_id) => {
+                result = self.scheduler_client.next_task(
+                    self.em_id,
+                    self.prev_assignments.pop_front()
```
sitaowang1998, line 267: Why not report all received assignments?

LinZhihao-723: I'm not sure yet what the multi-executor implementation would look like, so I made the `next_task` only send one processed task assignment, assuming the task assignment is from its own previous execution. I think it should be fair for now since we only have one assignment per EM. We could promote this param to a vector in the future, though, if we found it makes more sense for a multi-executor case.

sitaowang1998 (approved):

PR #351, opened 2026-06-19
feat(spider-storage): Add support for compressing job submission and job outputs.

sitaowang1998:
# Description

This PR adds:
- Job task graph and input compression in database and gRPC protocol.
- Conditional job output compression in database and gRPC protocol.

# Checklist

  breaking change.

# Validation performed

- [x] GitHub workflows pass.

## Release Notes

* **New Features**
  * Added zstd support for task-graph and job-input payloads, including end-to-end encoding/decoding.
  * Job outputs are now exchanged and stored as typed binary payloads with explicit encoding selection (e.g., raw or zstd).

* **Refactor**
  * Updated job registration, recovery, and cache handling to operate on compressed task-graph JSON and compressed job-input blobs.

* **Bug Fixes**
  * Improved handling and surfacing of binary-payload decoding and decompression failures.

* **Tests**
  * Updated and extended unit and runtime recovery tests to cover the new compressed payload flow and round-trips.


LinZhihao-723 (changes requested):
Reviewed `spider-core`. Changes:

- Move `ZSTD_LEVEL` into `encode_zstd_bytes`. The constant should be located at the end of the public symbols though.

As discussed offline, I think the way for task output serialization should not happen in the protobuf level. I will first send a PR to refactor the wire format and then propose the desired way in this PR.

components/spider-storage/src/cache/error.rs
LinZhihao-723, line 1: Moved to `JobSubmissionError`.

components/spider-storage/src/cache/job_submission.rs
LinZhihao-723, line 1: Moved to `job_submission` mod on the crate level.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -778,20 +776,21 @@ impl RecoverableJobRowProjection {
...
-            TaskGraph::from_json(&self.serialized_task_graph).map_err(DbError::task_graph_de)?;
-        let inputs: Vec<TaskInput> =
-            rmp_serde::from_slice(&self.serialized_job_inputs).map_err(DbError::value_de)?;
-        let submission = ValidatedJobSubmission::create(task_graph, inputs)
-            .map_err(|e| DbError::CorruptedDbState(e.to_string()))?;
+        let submission = ValidatedJobSubmission::create(
+            self.compressed_serialized_task_graph,
+            self.compressed_serialized_job_inputs,
+        )
+        .map_err(|e| DbError::CorruptedDbState(e.to_string()))?;
```
LinZhihao-723, line 787: Map `JobSubmissionError` as `CorruptedDbState` on recovery.

components/spider-storage/src/state/service.rs
```diff
@@ -111,27 +110,22 @@ impl<
...
-            .into_iter()
-            .map(TaskInput::ValuePayload)
-            .collect();
-        let job_submission =
-            ValidatedJobSubmission::create(task_graph, inputs).map_err(CacheError::from)?;
+        let job_submission = ValidatedJobSubmission::create(
+            compressed_serialized_task_graph,
+            compressed_serialized_inputs,
+        )
+        .map_err(|e| StorageServerError::BadRequest(e.to_string()))?;
```
LinZhihao-723, line 128: Forward as `BadRequest`.

components/spider-storage/Cargo.toml
LinZhihao-723, line 1: These dependencies no longer needed.

components/spider-storage/src/state/error.rs
```diff
@@ -15,9 +12,6 @@ pub enum StorageServerError {
     #[error(transparent)]
     Db(#[from] DbError),
 
-    #[error(transparent)]
-    Task(#[from] task::Error),
-
```
LinZhihao-723, line 20: No longer needed since it's been moved to `JobSubmissionError`.

LinZhihao-723 (commented):
Reviewed with the following changes.

## Input compression

The job submission is now a top-level mod in `spider-storage`. It has its own error enum. The validation step happens in the factory function of `ValidatedJobSubmission`, which takes the compressed task graph and inputs to construct the in-memory representation of a task graph.

The job submission holds the origin compressed task graph and inputs so that it can be submitted to DB without re-compression.

## Output compression

The task outputs are serialized in binary format for gRPC transmission. The serialized form is abstracted in a struct named `SerializedTaskOutputs`. This struct holds an underlying enum to determine how the task outputs are serialized, for now with two options:

- Serialize in wire-format (same as TDL).
- Serialize in zstd-compressed wire-format.

`SerializedTaskOutputs` can be further turned into a raw binary form, where the serailization option is given as the header byte, following the actual payload (wire-frame or zstd-compressed wire-frame). The raw binary format is used in two places:

- DB storage
- gRPC transition to return task outputs to user client

Currently, the compression support is mainly for reducing the network overhead. There're room for further optimization:

- Streaming compression: currently, the compression happens in a batch mode: all task outputs are fetched into a `Vec<TaskOutput>` in the cache layer, and then serialized into wire-frames, then compressed by zstd. A streaming zstd compressor can be implemneted under wire-format serializer to reduce the memory copy needed.
- When the outputs are returned from DB, the outputs are first deserialized into `Vec<TaskOutput>` and then serialized into `SerializedTaskOutputs` on its raw form. Ideally, the client can directly foward the raw bytes read from the database. However, that would require changing the current DB protocol, leading to more files to be touched. This optimization will be deferred to future PRs when the performance benefit becomes important.

@sitaowang1998 Please check my changes carefully before we merge.

LinZhihao-723 (approved):

PR #353, opened 2026-06-23
feat(spider-storage): Implement the job-orchestration gRPC service and migrate it to `Status`-based errors.

LinZhihao-723:
# Description

## Summary

This branch implements the storage server's `JobOrchestrationService` end-to-end and changes how that service reports failures: instead of embedding a structured error in every response body (the old `oneof result { payload; JobOrchestrationError error }` pattern), the job-orchestration RPCs now return a flat success payload and propagate failures through gRPC `Status` codes. The scheduler-side client is updated to match, and a small `RequestUnpack` abstraction is introduced to convert raw proto requests into spider-native types.

The scope is limited to `JobOrchestrationService`. The other services (`InboundQueueService`, `TaskInstanceManagementService`, `ResourceGroupManagementService`, etc.) still use their in-body `oneof` error model and are unchanged.

## Protocol changes (`storage.proto` + regenerated bindings)

These are wire-breaking changes to `JobOrchestrationService`; both the server and the in-repo scheduler client are updated together in this branch.

- Renamed `SubmitJob` / `SubmitJobRequest` / `SubmitJobResponse` to `RegisterJob` / `RegisterJobRequest` / `RegisterJobResponse`.
- Flattened every `JobOrchestrationService` response: removed the `oneof result { ...; JobOrchestrationError error }` wrapper from `RegisterJobResponse`, `JobStateResponse`, `JobOutputsResponse`, and `JobErrorResponse`. Each now carries only its success payload, and errors travel over the gRPC `Status` channel.
- Removed the `JobOrchestrationError` message and its `ErrCode` enum, plus the now-unused `JobManagementOperationResponse`.
- Regenerated `components/spider-proto-rust/src/generated/storage.rs` accordingly (net reduction of ~173 lines).

## Request unpacking (`spider-proto-rust/src/unpack.rs`, new)

A new `RequestUnpack` trait converts a raw proto request into its spider-native form, returning a lightweight `UnpackError` (a `Code` + `String`) rather than a full `tonic::Status`. Using a small error type keeps the request `Result` well under clippy's `result_large_err` threshold.

- `RegisterJobRequest` unpacks into `(ResourceGroupId, String, Vec<u8>)`, validating that the serialized task graph is UTF-8.
- `JobIdRequest` unpacks into `JobId`.
- `impl From<UnpackError> for Status` lets handlers `?`-forward an unpack failure directly into their `Result<_, Status>` return type. This impl lives in `spider-proto-rust` because `UnpackError` is local there, which satisfies the orphan rule.
- Added a `tracing` dependency and `pub mod unpack` to the crate.

## Storage server handlers (`spider-storage/src/grpc.rs`)

- Implemented all `JobOrchestrationService` RPCs against `ServiceState`: `register_job`, `start_job`, `cancel_job`, `get_job_state`, `get_job_outputs`, and `get_job_error`. Each unpacks its request via `unpack()?`, calls the service layer, and returns a flat response on success or a `Status` on failure.
- Added `job_orchestration_service_error_handler`, the single chokepoint that maps a `StorageServerError` to a `Status` and logs it.
- `GrpcServiceState` now holds the runtime's `CancellationToken` (its constructor takes it, and the binary passes it in) so that a fatal error can terminate and restart the service.

### Error mapping

`job_orchestration_service_error_handler` classifies each `StorageServerError`:

- `Cache(Internal(_))` is unrecoverable by design: it fires the cancellation token to restart the service and returns `UNAVAILABLE`.
- A database error (`Db(_)` or `Cache(Db(_))`) is further classified: resource-group-not-found / invalid-password to `UNAUTHENTICATED` (without leaking whether the group exists), job-not-found to `NOT_FOUND`, invalid job-state transitions to `FAILED_PRECONDITION`, and anything else to an opaque `INTERNAL`.
- `JobNotFound` to `NOT_FOUND`.
- A malformed task graph, malformed inputs, or bad request (`Task` / `Tdl` / `BadRequest`) to `INVALID_ARGUMENT`.
- Everything else (including stale-session and stale-state, which cannot occur in this deployment) to an opaque `INTERNAL`.

## Error-model refactor (`state/error.rs`, `cache/error.rs`, `state/job_cache.rs`)

Moved `JobAlreadyExists` out of `StorageServerError` and into `InternalError`. A duplicate job ID is a DB-to-cache inconsistency (the database mints the ID, so a client can never trigger it), which means it is a fatal cache-internal condition rather than a client-facing `ALREADY_EXISTS`. `JobCache::insert` now returns it through `Cache(Internal(...))`, and its tests were updated to match.

## Scheduler client (`spider-scheduler/src/storage_client/grpc.rs`)

- Rewrote `job_state` to consume the flat `JobStateResponse { state }` and to map the call's `tonic::Status` back into a `StorageClientError`: `NOT_FOUND` to `JobNotFound`, `INVALID_ARGUMENT` to `InvalidInput`, and any other code (including the server's `UNAVAILABLE` and `INTERNAL`) to `Server`. The mapping is inlined into the `map_err` closure.
- Removed `job_orchestration_error_to_client_error`, since the `JobOrchestrationError` proto type it decoded no longer exists.
- The poll paths (`poll_ready`, `poll_commit_ready`, `poll_cleanup_ready`) are unchanged; the `InboundQueueService` still uses in-body `oneof` errors.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **Refactor**
  * Job submission API updated: `SubmitJob` operation replaced with `RegisterJob` for improved clarity and streamlined error responses.
  * Enhanced error reporting with more specific gRPC status codes for better troubleshooting.
  * Improved service shutdown coordination for increased operational reliability.


components/spider-proto-rust/src/unpack/storage.rs
```diff
@@ -0,0 +1,66 @@
...
+    /// Returns a [`UnpackError`] on failure.
+    fn unpack(self) -> Result<Self::Unpacked, UnpackError>;
+}
+
+/// Unpacks [`RegisterJobRequest`] into a tuple containing:
+///
+/// * The resource group ID.
+/// * The serialized task graph.
+/// * The serialized inputs.
+impl RequestUnpack for RegisterJobRequest {
```
sitaowang1998, line 16: I think we should move `impl RequestUnpack` into separate files. There will be many requests that we need to unpack later.

LinZhihao-723: Fixed.

sitaowang1998 (approved):

components/spider-storage/src/grpc.rs
```diff
@@ -50,11 +56,126 @@ impl<
...
+            }
+
+            _ => {
+                tracing::error!(
+                    error = % error,
+                    service = SERVICE_NAME,
+                    tag,
+                    "Unexpected internal error."
+                );
+                Status::internal("internal error")
```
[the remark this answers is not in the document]
LinZhihao-723: This should be fine. The official clippy linter doesn't complain about this.

PR #358, opened 2026-06-26
fix(spider-storage): Reserve `UNAVAILABLE` gRPC status code for network connection errors.

LinZhihao-723:
# Description

`UNAVAILABLE` was being returned by user code in two storage handlers, which made it ambiguous: a genuine transport failure (lost or unestablished connection) and an application-level error both surfaced as `UNAVAILABLE`, so the execution-manager client could not tell them apart. Per the [gRPC status-code guidance](https://grpc.io/docs/guides/status-codes/), only a small set of codes are appropriate for user code to return, and `UNAVAILABLE` is best left to the transport. This PR remaps the two offending cases onto application-appropriate codes and updates the execution-manager client to treat `UNAVAILABLE` as a transport error. It closes the follow-up flagged at the end of #354.

## Server changes (`spider-storage/src/grpc.rs`)

- `job_orchestration_service_error_handler`: a fatal cache-internal error now returns `INTERNAL` instead of `UNAVAILABLE` (it still fires the cancellation token and restarts the service).
- `task_instance_management_service_error_handler`: a request from a stale session now returns `NOT_FOUND` instead of `UNAVAILABLE`. The other mappings are unchanged.
- Updated the handler doc comments to match.

## Client changes (`spider-execution-manager/src/client/grpc/storage.rs`)

- `status_to_error` now maps `Code::Unavailable` to `StorageResponseError::Transport`, so a lost or unestablished connection is reported as a transport failure rather than as an application error.
- Stale-session detection is remapped to follow the server: `Code::NotFound` → `StorageResponseError::StaleSession`. This is required for correctness — without it, the server's new `NOT_FOUND` stale-session response would fall through to the generic `Server` case and stale-session handling would silently break.
- Updated the unit tests: replaced `status_maps_unavailable_to_stale_session` with `status_maps_not_found_to_stale_session`, and added `status_maps_unavailable_to_transport`.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gRPC error responses for storage-related issues, so some failures now return more appropriate status codes.
  * Session-related errors are now reported more consistently, helping clients distinguish missing, stale, and transport-related problems.
  * Updated error messages to better reflect the actual issue returned to callers.

### DIFF
--- a/components/spider-execution-manager/src/client/grpc/storage.rs
+++ b/components/spider-execution-manager/src/client/grpc/storage.rs
@@ -135,15 +135,17 @@ impl StorageClient for GrpcStorageClient {
 ///
 /// The [`StorageResponseError`] for `status`'s code:
 ///
-/// * [`StorageResponseError::StaleSession`] for `UNAVAILABLE`.
+/// * [`StorageResponseError::StaleSession`] for `NOT_FOUND`.
 /// * [`StorageResponseError::CacheStale`] for `FAILED_PRECONDITION`.
 /// * [`StorageResponseError::InvalidInput`] for `INVALID_ARGUMENT`.
+/// * [`StorageResponseError::Transport`] for `UNAVAILABLE` (a lost or unestablished connection).
 /// * [`StorageResponseError::Server`] for any other code.
 fn status_to_error(status: &Status) -> StorageResponseError {
     match status.code() {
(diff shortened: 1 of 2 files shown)


LinZhihao-723 (commented):
@sitaowang1998 Please check this PR since it may affect the error handling in the coming service implementations.

sitaowang1998 (approved):

PR #359, opened 2026-06-26
feat(huntsman): Replace gRPC protocol custom error code with gRPC status.

sitaowang1998:
# Description

This PR:
- Removes custom error code in gRPC protocol.
- Adds gRPC status to error conversion.

# Checklist

  breaking change.

# Validation performed

- [x] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * None.
* **Bug Fixes**
  * Improved gRPC error handling by mapping status codes to clearer client-friendly errors for liveness, registration/heartbeat, and scheduler inbound-queue polling.
  * Updated execution manager and scheduler response parsing to handle the new protobuf shapes and properly detect missing expected data.
* **Breaking Changes**
  * Updated storage/proto RPC contracts for task completion, resource-group verification, and related scheduler/reporting calls to use success-only responses and simplified response structures.
* **Tests**
  * Refreshed unit tests to match the new protobuf shapes and validate the updated status/response mappings.


components/spider-scheduler/src/storage_client/grpc.rs
```diff
@@ -137,19 +136,20 @@ impl SchedulerStorageClient for GrpcSchedulerStorageClient {
...
+/// The [`StorageClientError`] for `status`'s code:
+///
+/// * [`StorageClientError::InboundClosed`] for `UNAVAILABLE`.
+/// * [`StorageClientError::InvalidInput`] for `INVALID_ARGUMENT`.
+/// * [`StorageClientError::Server`] for any other code.
+fn inbound_status_to_error(status: &Status) -> StorageClientError {
+    match status.code() {
+        Code::Unavailable => StorageClientError::InboundClosed,
+        Code::InvalidArgument => to_invalid_input_error(status.message()),
+        _ => StorageClientError::Server(status.message().to_owned()),
```
LinZhihao-723: I don't think it would make too full sense to implement this handler before we have a complete server side implementation, but **have you read [this](https://github.com/y-scope/spider/pull/358#pullrequestreview-4582752172) carefully**? Like the whole purpose of that PR is to reserve `UNAVAILABLE` for network transmission error, and I explicitly said this may affect what u will need to implement in the other PRs.

If this is intentional, you should document it in the PR description.

components/spider-execution-manager/src/client/grpc/liveness.rs
```diff
@@ -109,16 +109,20 @@ impl From<storage::ExecutionManagerLivenessError> for LivenessResponseError {
...
 ) -> Result<RegistrationResponse, LivenessResponseError> {
-    match response.result {
-        Some(register_execution_manager_response::Result::Registration(registration)) => {
+    match response.registration {
+        Some(registration) => {
+            if registration.session_id == 0 {
+                return Err(LivenessResponseError::Transport(
+                    "register execution manager response carried a zero session id".to_owned(),
+                ));
+            }
```
LinZhihao-723: - Can you explain which policy assumes that 0 is an invalid session ID?
- Why is it considered a transport error instead of another fatal error, if that violates the system's assumption?

components/spider-proto/storage/storage.proto
```diff
@@ -27,7 +27,7 @@ service InboundQueueService {
 
 service ResourceGroupManagementService {
   rpc AddResourceGroup(AddResourceGroupRequest) returns (ResourceGroupIdResponse);
-  rpc VerifyResourceGroup(VerifyResourceGroupRequest) returns (ResourceGroupOperationResponse);
+  rpc VerifyResourceGroup(VerifyResourceGroupRequest) returns (common.Void);
```
LinZhihao-723, line 30: Since we're using `common.Void` in scheduler protobuf definitions for void returns already, we should be consistent and avoid defining more void types.

LinZhihao-723 (changes requested):

components/spider-execution-manager/src/client/grpc/liveness.rs
sitaowang1998: * This is more a implementation detail. MariaDB auto-increment column starts at 1, so it will never be 0.
* This is my bad. Should I extend illegal id error to fit all illegal ids?

LinZhihao-723: No, I don't think we should handle this ID at all since by policy 0 is a valid ID; it's just impossible if it's backed by mariaDB.

components/spider-scheduler/src/storage_client/grpc.rs
```diff
@@ -142,10 +142,9 @@ impl SchedulerStorageClient for GrpcSchedulerStorageClient {
 ///
 /// The [`StorageClientError`] for `status`'s code:
 ///
-/// * [`StorageClientError::Transport`] for `UNAVAILABLE` (a network-level failure).
+/// * [`StorageClientError::Transport`] for `UNAVAILABLE` (a lost or unestablished connection).
```
LinZhihao-723, line 145: If similar wording already exists in the code base, we'd prefer to be consistent: https://github.com/y-scope/spider/blob/76a7adcf86262afe01bb85e66a4bfddaad76d297/components/spider-execution-manager/src/client/grpc/storage.rs#L150

Ask your coding agent to follow this rule.

components/spider-scheduler/src/storage_client/grpc.rs
```diff
@@ -142,10 +142,9 @@ impl SchedulerStorageClient for GrpcSchedulerStorageClient {
...
 /// * [`StorageClientError::InvalidInput`] for `INVALID_ARGUMENT`.
-/// * [`StorageClientError::Server`] for any other code (including `INTERNAL`, which the storage
-///   server sends when the inbound queue is closed).
+/// * [`StorageClientError::Server`] for any other code.
```
LinZhihao-723, line 147: This really sounds like the conventional comments claude would add if you explicitly ask for a behavior change, but do you think this from-no-where explanation makes sense in general?
You are expected to check and remove these comments.

components/spider-scheduler/src/storage_client/mod.rs
```diff
@@ -39,8 +39,7 @@ pub trait SchedulerStorageClient: Send + Sync + Clone {
     ///
     /// Returns an error if:
     ///
-    /// * [`StorageClientError::Server`] if the regular-task lane is closed and can no longer yield
-    ///   entries, or the storage server returns another error.
```
LinZhihao-723, line 43: Similar to https://github.com/y-scope/spider/pull/359/changes/79bcb043616d4c5ff2eb3fbdef489aaec9cc30fd#r3488578792 but even worse: what is "another error" referring to?
If you want to be general, then just be general. The same file has an example of how to describe a general case:  https://github.com/y-scope/spider/blob/76a7adcf86262afe01bb85e66a4bfddaad76d297/components/spider-scheduler/src/storage_client/mod.rs#L117.

LinZhihao-723 (changes requested):
Please confirm you have read the following comments.

components/spider-scheduler/src/storage_client/grpc.rs
sitaowang1998: I have added the instruction. However, I am only sure it will trigger and look for existing convention when I pose a direct question. When running a long task, it seem to skip checking on convention. Maybe I can add to review instruction as well.

LinZhihao-723: - Yes, adding to the review instruction would be good.
- You should also have a sense for checking existing ones manually.

LinZhihao-723 (approved):

PR #368, opened 2026-06-30
feat(spider-scheduler): Add the runtime that wires the scheduler core, dispatch queue, and execution manager registry together.

LinZhihao-723:
# Description

This PR adds the `spider-scheduler` runtime: the assembly layer that constructs and supervises the scheduler's moving parts behind a single factory, mirroring the existing `spider-storage` runtime.

`create_runtime` is generic over the storage client (`SchedulerStorageClient`) so it accepts a real gRPC client in production and a mock in tests. Given a `RuntimeConfig` and a storage client, it:

* Creates a cancellation token, passes it to the scheduler core and the execution manager registry, and returns it to the caller so an external failure can cancel every background task.
* Registers the scheduler with storage via the new `SchedulerStorageClient::register` call, advertising the host/port execution managers should reach it on, and adopts the `SchedulerId` storage assigns.
* Builds the real channel-based dispatch queue with `create_dispatch_queue`, hands the sink to the core, and surfaces the reader through the returned service.
* Spawns the scheduler core coroutine and retains its `JoinHandle`.
* Returns the `Runtime`, the `SchedulerServiceState` built from the scheduler ID, the registry, and the dispatch queue reader, and the cancellation token.

`Runtime::stop` cancels the token, joins the core within a configurable timeout, logs the core's exit result, and drains any task assignments still buffered on the reschedule queue.

A serde-deserializable `SchedulerConfig` enum selects and configures the scheduling algorithm (currently only round-robin). Its `make_core` method dispatches to the selected variant to produce a `dyn SchedulerCore`, so the runtime stays agnostic to the concrete algorithm.

## Supporting changes

* Added `SchedulerStorageClient::register` to the trait and its gRPC implementation, routed through the scheduler-registration connection pool.
* Changed `SchedulerCore::run`'s receiver from `self` to `self: Box<Self>` so the core is object-safe and can be held as a `Box<dyn SchedulerCore>`.
* Added the `SchedulerRuntimeError` error type covering storage-client registration failures and stop-timeout failures.
* Exported the new `config`, `runtime` modules and their public types from the crate root.

## Summary by CodeRabbit

* **New Features**
  * Added scheduler runtime support, including startup, background execution, and graceful shutdown handling.
  * Added scheduler service operations for assigning work, sending heartbeats, and shutting down execution managers.
  * Expanded scheduler configuration support with selectable scheduling setup and queue sizing.

* **Bug Fixes**
  * Improved shutdown behaviour so pending work is tracked, drained, and rescheduled more reliably.
  * Updated scheduler processing to work correctly with background task execution and cancellation.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable scheduler runtime that starts the scheduler in the background and supports graceful shutdown with a timeout.
  * Added scheduler registration so the service can announce its address and receive an assigned scheduler ID.

* **Bug Fixes**
  * Improved shutdown behaviour to stop cleanly when possible and report a timeout if the scheduler does not exit in time.
  * Expanded scheduling support to better handle the current round-robin configuration.

### DIFF
--- a/components/spider-scheduler/src/config.rs
+++ b/components/spider-scheduler/src/config.rs
@@ -1,0 +1,56 @@
+//! The scheduler core configuration that selects and configures the scheduling algorithm.
+
+use serde::Deserialize;
+
+use crate::{
+    core::SchedulerCore,
+    core_impl::RoundRobinConfig,
+    dispatch_queue::DispatchQueueSink,
+    storage_client::SchedulerStorageClient,
+};
+
+/// The configuration that selects and configures the scheduler core's scheduling algorithm.
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum SchedulerConfig {
+    /// The round-robin scheduling algorithm.
+    RoundRobin(RoundRobinConfig),
+}
+
+impl SchedulerConfig {
+    /// Creates a ready-to-run scheduler core from the selected configuration.
+    ///
+    /// # Type Parameters
+    ///
+    /// * `SchedulerStorageClientType` - The storage client the core polls and registers through.
+    /// * `DispatchQueueSinkType` - The dispatch sink that task assignments are written to.
+    ///
+    /// # Returns
+    ///
+    /// A boxed [`SchedulerCore`] configured by the selected variant.
+    #[must_use]
+    pub fn make_core<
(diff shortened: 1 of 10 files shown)


LinZhihao-723 (commented):
Just realized we didn't wire the reschedule queue to the core yet. Will add it in a separate PR.

sitaowang1998 (approved):

PR #376, opened 2026-07-07
build(huntsman): Add Dockerfile and GitHub CI for Spider Huntsman images.

20001020ycx:
# Description

This PR adds a multi-stage Dockerfile that builds container images for the Spider Huntsman services, plus a CI workflow to publish them to GHCR. It produces three per-service images:
  - storage → spider_storage_grpc_server
  - scheduler → spider_scheduler_grpc_server
  - worker → spider_execution_manager + spider-task-execution
 
The worker image ships without any TDL .so package. So we expect user-defined TDL package is layered on top of this published worker image during the docker compose or k8s integration.

Configuration for all three images is supplied at the docker compose or k8s integration rather than baked into the images.

## Notes for reviewers
 - The image-build workflow will publish images on push/tags, pull request only builds images for validation. The image publication process will run after merge this PR to main. You can preview the final artifacts on my [fork](https://github.com/20001020ycx/spider).
  - Here I would also like to discuss on the tagging conventions assumed in this PR:
    - Every commit pushed to main publishes an image tagged with its commit SHA (:sha-<commit>), and updates the rolling :main tag to point at that latest build.
    - Each release tag on spider-huntsman also publishes an image (e.g. :spider-huntsman-v0.1.0).

> [!Note]
> This PR does not add new tests.

# Checklist

  breaking change.

# Validation performed

All CIs pass, including the new CI that builds the image using the Dockerfile added in this PR.

**Published images (pulled from the fork's GHCR):**
```shell
docker login ghcr.io -u <user>
base=ghcr.io/20001020ycx/spider; tag=sha-c4dd00f
docker pull $base/storage:$tag
docker pull $base/scheduler:$tag
docker pull $base/worker:$tag

docker run --rm $base/storage:$tag                             spider_storage_grpc_server   --help
docker run --rm $base/scheduler:$tag                           spider_scheduler_grpc_server --help
docker run --rm $base/worker:$tag spider_execution_manager     --help
```

## Summary by CodeRabbit

* **New Features**
  * Added automated Docker image builds and publishing for the storage, scheduler, and worker services.
  * Added container support for multiple release images, including tag-based builds and manual runs.
* **Chores**
  * Improved build automation to run only when relevant files change and to manage concurrent runs more safely.
  * Standardized image naming and access rules for container publishing.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.git/
+build/
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,0 @@
-.git/
-build/
-target/
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 .git/
 build/
+target/
--- a/.github/workflows/spider-image-build.yaml
+++ b/.github/workflows/spider-image-build.yaml
@@ -20,7 +20,7 @@ jobs:
       packages: "write"
     strategy:
       matrix:
-        service: ["storage", "scheduler", "execution-manager-and-task-executor"]
+        service: ["storage", "scheduler", "worker"]
     steps:
       - uses: "actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd"  # v6.0.2
         with:
--- a/.github/workflows/spider-image-build.yaml
+++ b/.github/workflows/spider-image-build.yaml
@@ -1,0 +1,56 @@
+name: "spider-image-build"
+
+on:
+  push:
+    branches: ["main"]
+    tags: ["v*"]
+  workflow_dispatch:
+
+concurrency:
+  group: "${{github.workflow}}-${{github.ref}}"
+  cancel-in-progress: true
+
+permissions: {}
+
+jobs:
+  build-and-push:
+    runs-on: "ubuntu-latest"
+    permissions:
+      contents: "read"
+      packages: "write"
+    strategy:
+      matrix:
+        service: ["storage", "scheduler", "execution-manager-and-task-executor"]
+    steps:
+      - uses: "actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd"  # v6.0.2
+
+      - name: "Lowercase the image repository"
+        id: "repo"
+        run: >-
+          echo "name=${GITHUB_REPOSITORY,,}" >> "$GITHUB_OUTPUT"
+
+      - uses: "docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2"  # v3.10.0
+
+      - uses: "docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121"  # v4.1.0
+        with:
+          registry: "ghcr.io"
+          username: "${{github.actor}}"
+          password: "${{secrets.GITHUB_TOKEN}}"
+
+      - uses: "docker/metadata-action@902fa8ec7d6ecbf8d84d538b9b233a880e428804"  # v5.7.0
+        id: "meta"
+        with:
+          images: "ghcr.io/${{steps.repo.outputs.name}}/${{matrix.service}}"
+          tags: |
+            type=ref,event=branch
+            type=ref,event=tag
+            type=sha
+
+      - uses: "docker/build-push-action@471d1dc4e07e5cdedd4c2171150001c434f0b7a4"  # v6.15.0
+        with:
+          context: "."
+          file: "tools/docker/Dockerfile"
+          target: "${{matrix.service}}"
+          push: true
+          tags: "${{steps.meta.outputs.tags}}"
+          labels: "${{steps.meta.outputs.labels}}"
--- a/.github/workflows/spider-image-build.yaml
+++ b/.github/workflows/spider-image-build.yaml
@@ -1,16 +1,33 @@
 name: "spider-image-build"
 
 on:
+  pull_request:
+    paths: &monitored_paths
+      - ".cargo/**"
+      - ".dockerignore"
+      - ".github/workflows/spider-image-build.yaml"
+      - ".gitmodules"
+      - "Cargo.lock"
+      - "Cargo.toml"
+      - "components/**"
+      - "taskfile.yaml"
+      - "taskfiles/**"
+      - "tools/docker/**"
+      - "tools/scripts/**"
+      - "tools/yscope-dev-utils"
   push:
     branches: ["main"]
+    paths: *monitored_paths
     tags: ["spider-huntsman-v*"]
   workflow_dispatch:
 
+permissions: {}
+
 concurrency:
   group: "${{github.workflow}}-${{github.ref}}"
-  cancel-in-progress: true
-
-permissions: {}
+  # Cancel in-progress jobs for efficiency. Exclude the `main` branch to allow uninterrupted
+  # publishing of container images.
+  cancel-in-progress: "${{github.ref != 'refs/heads/main'}}"
 
 jobs:
   build-and-push:
@@ -26,7 +43,7 @@ jobs:
         with:
           submodules: "recursive"
 
-      - name: "Lowercase the image repository"
+      - name: "Sanitize repository name"
         id: "repo"
         run: >-
           echo "name=${GITHUB_REPOSITORY,,}" >> "$GITHUB_OUTPUT"
@@ -34,6 +51,7 @@ jobs:
       - uses: "docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2"  # v3.10.0
 
       - uses: "docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121"  # v4.1.0
+        if: "github.event_name != 'pull_request'"
         with:
           registry: "ghcr.io"
           username: "${{github.actor}}"
@@ -53,6 +71,6 @@ jobs:
           context: "."
           file: "tools/docker/Dockerfile"
           target: "${{matrix.service}}"
-          push: true
+          push: "${{github.event_name != 'pull_request'}}"
           tags: "${{steps.meta.outputs.tags}}"
           labels: "${{steps.meta.outputs.labels}}"
--- a/.github/workflows/spider-image-build.yaml
+++ b/.github/workflows/spider-image-build.yaml
@@ -4,7 +4,6 @@ on:
   pull_request:
     paths: &monitored_paths
       - ".cargo/**"
-      - ".dockerignore"
       - ".github/workflows/spider-image-build.yaml"
       - ".gitmodules"
       - "Cargo.lock"
--- a/.github/workflows/spider-image-build.yaml
+++ b/.github/workflows/spider-image-build.yaml
@@ -3,7 +3,7 @@ name: "spider-image-build"
 on:
   push:
     branches: ["main"]
-    tags: ["v*"]
+    tags: ["spider-huntsman-v*"]
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/spider-image-build.yaml
+++ b/.github/workflows/spider-image-build.yaml
@@ -23,6 +23,8 @@ jobs:
         service: ["storage", "scheduler", "execution-manager-and-task-executor"]
     steps:
       - uses: "actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd"  # v6.0.2
+        with:
+          submodules: "recursive"
 
       - name: "Lowercase the image repository"
         id: "repo"
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -1,0 +1,25 @@
+# syntax=docker/dockerfile:1
+
+ARG RUST_VERSION=1.96
+ARG DEBIAN_VERSION_CODENAME=bookworm
+
+FROM rust:${RUST_VERSION}-${DEBIAN_VERSION_CODENAME} AS builder
+
+WORKDIR /spider
+COPY . .
+
+RUN cargo build --release \
+    --bin spider_storage_grpc_server \
+    --bin spider_scheduler_grpc_server \
+    --bin spider_execution_manager \
+    --bin spider-task-executor
+
+FROM debian:${DEBIAN_VERSION_CODENAME}-slim AS storage
+COPY --from=builder /spider/target/release/spider_storage_grpc_server /usr/local/bin/
+
+FROM debian:${DEBIAN_VERSION_CODENAME}-slim AS scheduler
+COPY --from=builder /spider/target/release/spider_scheduler_grpc_server /usr/local/bin/
+
+FROM debian:${DEBIAN_VERSION_CODENAME}-slim AS execution-manager-and-task-executor
+COPY --from=builder /spider/target/release/spider_execution_manager /usr/local/bin/
+COPY --from=builder /spider/target/release/spider-task-executor /usr/local/bin/
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -5,7 +5,17 @@ ARG UBUNTU_VERSION_CODENAME=jammy
 FROM ubuntu:${UBUNTU_VERSION_CODENAME} AS builder
 
 WORKDIR /spider
-COPY --link . .
+
+# Whitelist only the sources required to build the Rust ("Huntsman") binaries; the C++ and spider-py
+# components are deliberately excluded.
+COPY --link Cargo.toml Cargo.lock taskfile.yaml ./
+COPY --link .cargo ./.cargo
+COPY --link components ./components
+COPY --link examples/huntsman ./examples/huntsman
+COPY --link taskfiles ./taskfiles
+COPY --link tests/huntsman ./tests/huntsman
+COPY --link tools/scripts ./tools/scripts
+COPY --link tools/yscope-dev-utils ./tools/yscope-dev-utils
 
 RUN tools/scripts/lib_install/ubuntu/install-dev-huntsman.sh
 RUN sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b /usr/local/bin
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -5,7 +5,7 @@ ARG UBUNTU_VERSION_CODENAME=jammy
 FROM ubuntu:${UBUNTU_VERSION_CODENAME} AS builder
 
 WORKDIR /spider
-COPY . .
+COPY --link . .
 
 RUN tools/scripts/lib_install/ubuntu/install-dev-huntsman.sh
 RUN sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b /usr/local/bin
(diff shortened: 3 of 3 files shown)


tools/docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+FROM rust:${RUST_VERSION}-${DEBIAN_VERSION_CODENAME} AS builder
+
+WORKDIR /spider
+COPY . .
+
+RUN cargo build --release \
+    --bin spider_storage_grpc_server \
+    --bin spider_scheduler_grpc_server \
+    --bin spider_execution_manager \
+    --bin spider-task-executor
```
sitaowang1998: Should we use build task in case we add something extra in build task later?

.github/workflows/spider-image-build.yaml
```diff
@@ -0,0 +1,56 @@
+name: "spider-image-build"
+
+on:
+  push:
+    branches: ["main"]
+    tags: ["v*"]
```
sitaowang1998: I am not sure about how we will tag new huntsman version. @LinZhihao-723 Will `v*` be used for version tagging?

sitaowang1998 (changes requested):
We should have a task to build images to be used locally and in GitHub CI.

sitaowang1998:
The new yaml files should be included in the existing `lint:yml` task.

20001020ycx:
> We should have a task to build images to be used locally and in GitHub CI.

 I agree with the Dockerfile change you mentioned, but I'm not sure about a separate task to build images in GitHub CI.
  - I'd imagine such a task would just run docker build — but then it also has to define its own tagging, whereas the build and tagging are already natively integrated into the CI via build-push-action/metadata-action. A separate build task would mean re-implementing the tagging and SHA hashing that we get from those for free. I believe CLP follows the same pattern proposed in this PR, and I did the same for Presto — but I'm open to any proposal if I'm missing something

sitaowang1998:
> > We should have a task to build images to be used locally and in GitHub CI.
> 
> I agree with the Dockerfile change you mentioned, but I'm not sure about a separate task to build images in GitHub CI.
> 
>     * I'd imagine such a task would just run docker build — but then it also has to define its own tagging, whereas the build and tagging are already natively integrated into the CI via build-push-action/metadata-action. A separate build task would mean re-implementing the tagging and SHA hashing that we get from those for free. I believe CLP follows the same pattern proposed in this PR, and I did the same for Presto — but I'm open to any proposal if I'm missing something

I agree. The only benefit for using task in to make sure the published image's content is identical to the locally built one. However, the extra complexity added is not worth it. For reference, clp directly use docker actions to build and publish images.

.github/workflows/spider-image-build.yaml
LinZhihao-723: Since we also have `spider-py` and wolf, shall we have a complete prefix like `spider-huntsman-v*`?

LinZhihao-723:
> > We should have a task to build images to be used locally and in GitHub CI.
> 
> I agree with the Dockerfile change you mentioned, but I'm not sure about a separate task to build images in GitHub CI.
> 
> * I'd imagine such a task would just run docker build — but then it also has to define its own tagging, whereas the build and tagging are already natively integrated into the CI via build-push-action/metadata-action. A separate build task would mean re-implementing the tagging and SHA hashing that we get from those for free. I believe CLP follows the same pattern proposed in this PR, and I did the same for Presto — but I'm open to any proposal if I'm missing something

Without a task to build the Docker image, how do we expect to test with the latest image with unmerged local changes?
And since our integration test will depend on the image, how do we expect the integration test to figure out which image to use?

20001020ycx:
> > > We should have a task to build images to be used locally and in GitHub CI.
> > 
> > 
> > I agree with the Dockerfile change you mentioned, but I'm not sure about a separate task to build images in GitHub CI.
> > 
> > * I'd imagine such a task would just run docker build — but then it also has to define its own tagging, whereas the build and tagging are already natively integrated into the CI via build-push-action/metadata-action. A separate build task would mean re-implementing the tagging and SHA hashing that we get from those for free. I believe CLP follows the same pattern proposed in this PR, and I did the same for Presto — but I'm open to any proposal if I'm missing something
> 
> Without a task to build the Docker image, how do we expect to test with the latest image with unmerged local changes? And since our integration test will depend on the image, how do we expect the integration test to figure out which image to use?

Fair point, but just to be clear: I am NOT saying we shouldn't not have a taskfile for building images locally, I am saying it should NOT be used in the GitHub CI.

What you're suggesting belongs to the docker compose integration work, which we should do — but in a follow-up, not this PR. Let me use CLP to illustrate why.

For k8s integration, following the convention from CLP repo's helm chart: 
```
 image:
    clpPackage:
      repository: "ghcr.io/y-scope/clp/clp-package"
      pullPolicy: "Always"
      tag: "main"
```
the clpPackage image is published by the GitHub CI flow with the exact mechanism implemented in this PR.  The chart consumes the image generated by GHCR by default.

 For CLP's integration tests, which run against docker compose rather than k8s, there is a taskfile: `task docker-images:package`. It builds the image locally and records its image ID, which docker compose then consumes to bring up the services and run the tests against them. So the taskfile is needed there because the integration test runs a locally-built image, not the CI-published one.

Therefore the taskfile is not on the critical path of the k8s (Helm) integration — that path uses the CI-published image. I understand it is needed for the docker compose path described in the integration-testing plan, and I'll add it as a follow-up PR.

.github/workflows/spider-image-build.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+jobs:
+  build-and-push:
+    runs-on: "ubuntu-latest"
+    permissions:
+      contents: "read"
+      packages: "write"
+    strategy:
+      matrix:
+        service: ["storage", "scheduler", "execution-manager-and-task-executor"]
```
sitaowang1998: My bad. Should have caught it in the first round. `execution-manager-and-task-executor` is too long for a service name, and we might pair multiple task executors with a single execution manager later. How about:

```suggestion
        service: ["storage", "scheduler", "worker"]
```

.github/workflows/spider-image-build.yaml
```diff
@@ -0,0 +1,58 @@
...
+          submodules: "recursive"
+
+      - name: "Lowercase the image repository"
+        id: "repo"
+        run: >-
+          echo "name=${GITHUB_REPOSITORY,,}" >> "$GITHUB_OUTPUT"
+
+      - uses: "docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2"  # v3.10.0
+
+      - uses: "docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121"  # v4.1.0
```
sitaowang1998, line 52: Why we are using this version? All other actions uses the same version as in clp. Only this one uses a newer version, but not the newest. We should decide to either replicate all version from clp, or to use latest versions for all (unless they are release just yesterday.).

sitaowang1998 (changes requested):

.github/workflows/spider-image-build.yaml
20001020ycx: Was about to discuss with you guys in what naming I should give, I think this rings, thanks for the catch. Also updated all references of execution-manager-and-task-executor to worker, including the PR description.

.github/workflows/spider-image-build.yaml
20001020ycx: sry, I am confused, I think these two docker action are using the same version as CLP repo. Please let me know if I am missing something:
docker/setup-buildx-action: 
https://github.com/y-scope/clp/blob/dc104612f62b876d3fac665ad8eab72f277c7210/.github/actions/clp-build-runtime-image/action.yaml#L30

docker/login-action:
https://github.com/y-scope/clp/blob/dc104612f62b876d3fac665ad8eab72f277c7210/.github/actions/clp-build-runtime-image/action.yaml#L33

.github/workflows/spider-image-build.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+jobs:
+  build-and-push:
+    runs-on: "ubuntu-latest"
+    permissions:
+      contents: "read"
+      packages: "write"
+    strategy:
+      matrix:
+        service: ["storage", "scheduler", "worker"]
```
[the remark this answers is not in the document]
20001020ycx: this is fixed in the latest commit right?

.github/workflows/spider-image-build.yaml
sitaowang1998: I was taking reference to https://github.com/y-scope/clp/blob/dc104612f62b876d3fac665ad8eab72f277c7210/.github/actions/clp-core-build-containers/action.yaml. Seem like clp itself is not consistent. I'm ok with using either version.

sitaowang1998 (dismissed):

tools/docker/Dockerfile
LinZhihao-723, line 1: As discussed offline, I think we should use a whitelist when copying the necessary files into the Docker build.
- This is probably easier to maintain in the long-term.
- We can get rid of unnecessary components, for example, spider-py and C++ components.

@junhaoliao What do you think?

.dockerignore
```diff
@@ -0,0 +1,3 @@
+.git/
+build/
+target/
```
LinZhihao-723: - If we're using the whilelist solution, we probably don't need this file.
- I'm not sure if we should add this to our *ignore file. Officially this target directory shouldn't exist when building with our taskfiles.

junhaoliao:
```suggestion
# Git
.git/

# Build
.task/
build/
target/
```

tools/docker/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+FROM runtime-base AS storage
+COPY --from=builder /spider/build/rust-targets/release/spider_storage_grpc_server /usr/local/bin/
+
+FROM runtime-base AS scheduler
+COPY --from=builder /spider/build/rust-targets/release/spider_scheduler_grpc_server /usr/local/bin/
+
+FROM runtime-base AS worker
+COPY --from=builder /spider/build/rust-targets/release/spider_execution_manager /usr/local/bin/
+COPY --from=builder /spider/build/rust-targets/release/spider-task-executor /usr/local/bin/
```
LinZhihao-723: Sigh... the binary naming convention needs to be fixed. Will have a PR soon.

LinZhihao-723 (changes requested):

tools/docker/Dockerfile
```diff
@@ -0,0 +1,26 @@
+# syntax=docker/dockerfile:1
+
+ARG UBUNTU_VERSION_CODENAME=jammy
+
+FROM ubuntu:${UBUNTU_VERSION_CODENAME} AS builder
+
+WORKDIR /spider
+COPY . .
```
junhaoliao: to promote layer reuse

```suggestion
COPY --link . .
```

tools/docker/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+RUN tools/scripts/lib_install/ubuntu/install-dev-huntsman.sh
+RUN sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b /usr/local/bin
+RUN task build:rust
+
+FROM ubuntu:${UBUNTU_VERSION_CODENAME} AS runtime-base
+RUN useradd --create-home spider-user
+USER spider-user
+
+FROM runtime-base AS storage
+COPY --from=builder /spider/build/rust-targets/release/spider_storage_grpc_server /usr/local/bin/
```
junhaoliao:
```suggestion
COPY --from=builder --link /spider/build/rust-targets/release/spider_storage_grpc_server /usr/local/bin/
```

tools/docker/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+FROM runtime-base AS scheduler
+COPY --from=builder /spider/build/rust-targets/release/spider_scheduler_grpc_server /usr/local/bin/
```
junhaoliao:
```suggestion
COPY --from=builder --link /spider/build/rust-targets/release/spider_scheduler_grpc_server /usr/local/bin/
```

tools/docker/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+FROM runtime-base AS worker
+COPY --from=builder /spider/build/rust-targets/release/spider_execution_manager /usr/local/bin/
```
junhaoliao:
```suggestion
COPY --from=builder --link /spider/build/rust-targets/release/spider_execution_manager /usr/local/bin/
```

tools/docker/Dockerfile
junhaoliao:
```suggestion
COPY --from=builder --link /spider/build/rust-targets/release/spider-task-executor /usr/local/bin/
```

.github/workflows/spider-image-build.yaml
```diff
@@ -0,0 +1,58 @@
+name: "spider-image-build"
+
+on:
+  push:
```
junhaoliao, line 17: shall we ensure the PRs are checked too?

```suggestion
  pull_request:
    paths: &monitored_paths
      - ".cargo/**"
      - ".dockerignore"
      - ".github/workflows/spider-image-build.yaml"
      - ".gitmodules"
      - "Cargo.lock"
      - "Cargo.toml"
      - "components/**"
      - "taskfile.yaml"
      - "taskfiles/**"
      - "tools/docker/**"
      - "tools/scripts/**"
      - "tools/yscope-dev-utils"
  push:
```

.github/workflows/spider-image-build.yaml
```diff
@@ -0,0 +1,58 @@
...
+    branches: ["main"]
+    tags: ["spider-huntsman-v*"]
```
junhaoliao, line 20:
```suggestion
    paths: *monitored_paths
    tags: ["spider-huntsman-v*"]
```

.github/workflows/spider-image-build.yaml
```diff
@@ -0,0 +1,58 @@
...
+  workflow_dispatch:
+
+concurrency:
+  group: "${{github.workflow}}-${{github.ref}}"
+  cancel-in-progress: true
```
junhaoliao:
```suggestion
  # Cancel in-progress jobs for efficiency. Exclude the `main` branch to allow uninterrupted
  # publishing of container images.
  cancel-in-progress: "${{github.ref != 'refs/heads/main'}}"
```

.github/workflows/spider-image-build.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+permissions: {}
```
junhaoliao:
```suggestion
```

.github/workflows/spider-image-build.yaml
```diff
@@ -0,0 +1,58 @@
+name: "spider-image-build"
+
+on:
+  push:
+    branches: ["main"]
+    tags: ["spider-huntsman-v*"]
+  workflow_dispatch:
+
```
junhaoliao, line 22:
```suggestion

permissions: {}

```

.github/workflows/spider-image-build.yaml
```diff
@@ -0,0 +1,58 @@
...
+      packages: "write"
+    strategy:
+      matrix:
+        service: ["storage", "scheduler", "worker"]
+    steps:
+      - uses: "actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd"  # v6.0.2
+        with:
+          submodules: "recursive"
+
+      - name: "Lowercase the image repository"
```
junhaoliao:
```suggestion
      - name: "Sanitize repository name"
```

.github/workflows/spider-image-build.yaml
```diff
@@ -0,0 +1,58 @@
...
+        id: "repo"
+        run: >-
+          echo "name=${GITHUB_REPOSITORY,,}" >> "$GITHUB_OUTPUT"
+
+      - uses: "docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2"  # v3.10.0
+
+      - uses: "docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121"  # v4.1.0
```
junhaoliao, line 52:
```suggestion
      - uses: "docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121"  # v4.1.0
        if: "github.event_name != 'pull_request'"
```

.github/workflows/spider-image-build.yaml
```diff
@@ -0,0 +1,58 @@
...
+            type=ref,event=branch
+            type=ref,event=tag
+            type=sha
+
+      - uses: "docker/build-push-action@471d1dc4e07e5cdedd4c2171150001c434f0b7a4"  # v6.15.0
+        with:
+          context: "."
+          file: "tools/docker/Dockerfile"
+          target: "${{matrix.service}}"
+          push: true
```
junhaoliao:
```suggestion
          push: "${{github.event_name != 'pull_request'}}"
```

tools/docker/Dockerfile
junhaoliao: i agree an allowlist is better than a denylist

junhaoliao (changes requested):

.github/workflows/spider-image-build.yaml
junhaoliao: to match order of declaration in other workflow files - related: https://github.com/y-scope/spider/pull/376/changes#r3539347669

.github/workflows/spider-image-build.yaml
20001020ycx: Make sense, we should also validate if the build succeed in docker environment before publish to main. I thought these are publishing images per pull request at first, but now I get it.

.dockerignore
20001020ycx: the file is deleted as we move to whitelist COPY in dockerfile

20001020ycx: Good catch, done

tools/docker/Dockerfile
20001020ycx: done

20001020ycx:
PR description is updated after the changes discussed in the review.

tools/docker/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+COPY --link .cargo ./.cargo
+COPY --link components ./components
+COPY --link examples/huntsman ./examples/huntsman
+COPY --link taskfiles ./taskfiles
+COPY --link tests/huntsman ./tests/huntsman
+COPY --link tools/scripts ./tools/scripts
+COPY --link tools/yscope-dev-utils ./tools/yscope-dev-utils
+
+RUN tools/scripts/lib_install/ubuntu/install-dev-huntsman.sh
+RUN sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b /usr/local/bin
```
junhaoliao:
```suggestion
RUN sh -c "$(curl --fail --location --silent --show-error https://taskfile.dev/install.sh)" -- -d -b /usr/local/bin
```

junhaoliao: actually see https://github.com/y-scope/spider/pull/376/changes#r3540159724 about pinning the go-task version

```suggestion
RUN sh -c "$(curl --fail --location --silent --show-error https://taskfile.dev/install.sh)" -- \
    -d -b /usr/local/bin "v${GO_TASK_VERSION}"
```

tools/docker/Dockerfile
```diff
@@ -0,0 +1,34 @@
+# syntax=docker/dockerfile:1
+
+ARG UBUNTU_VERSION_CODENAME=jammy
+
+FROM ubuntu:${UBUNTU_VERSION_CODENAME} AS builder
+
```
junhaoliao, line 6:
```suggestion

ARG GO_TASK_VERSION=3.49.1

```

junhaoliao (changes requested):
for the title, how about
```
build(docker): Add Docker images and CI for Spider services.
```

tools/docker/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN useradd --create-home spider-user
+USER spider-user
+
+FROM runtime-base AS storage
+COPY --from=builder --link /spider/build/rust-targets/release/spdier-storage /usr/local/bin/
+
+FROM runtime-base AS scheduler
+COPY --from=builder --link /spider/build/rust-targets/release/spider-scheduler /usr/local/bin/
+
+FROM runtime-base AS worker
```
junhaoliao, line 35: per 233e1983226dacee9f0e359116bdb5b3db21c1a0

does it mean we need to rename these "worker" references too?

tools/docker/Dockerfile
20001020ycx: Done as https://github.com/y-scope/spider/pull/378 is merged

tools/docker/Dockerfile
junhaoliao: please ignore the comment. i missed that the "worker" name here already serves as a stage name because both the execution-manager and task-executor are getting copied

junhaoliao (dismissed):

LinZhihao-723 (approved):
For the PR title, how about:
```
build(huntsman): Add Dockerfile and GitHub CI for Spider Huntsman images.
```

PR #377, opened 2026-07-07
test(huntsman-integration): Add neural network tasks.

sitaowang1998:
# Description

This PR adds in `example/huntsman/nn` three components:
- `core`: Neural network neuron functions.
- `tasks`: Spider task wrapper for neuron functions.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

- [x] Add unit tests for neuron tasks.
- [x] Add a unit test to check that client correctly wire all inner tasks' outputs.
- [x] Add a client that runs a random neural network task graph on a Spider instance on baker and verify the result. The test code is in [another branch](https://github.com/sitaowang1998/spider/tree/e2e-nn-stack-test).
- [x] GitHub workflows pass.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary

* **New Features**
  * Added new neural-network example crates to the workspace: one for deterministic neuron math and one that exposes dense-layer tasks.
  * Provided three exported dense task variants (ReLU, sigmoid, identity) that compute a weighted sum plus bias and apply the selected activation.

* **Tests**
  * Added unit tests validating activations, weighted-sum behavior, and dense-layer outputs (including ReLU clamping).

PR #379, opened 2026-07-08
feat(huntsman): Add the end-to-end integration-test driver.

LinZhihao-723:
# Description

This PR adds the end-to-end integration-test driver as a new `e2e` crate under `tests/huntsman/`. The driver wraps `spider-client`'s `SpiderClient` in a process-wide singleton with the concurrency control that end-to-end scenarios share, and exposes a small surface for driving a job (task graph + inputs) through a live Spider deployment and asserting its terminal outcome.

This PR adds the driver library only; the concrete test scenarios that use it are added in a follow-up.

> [!Note]
> This PR does not add test scenarios. The `e2e` crate builds and lints as a library, and its `tests/` directory is reserved for the scenarios added later.

## New `e2e` crate (`tests/huntsman/e2e/`)

- Added `tests/huntsman/e2e/` as a new workspace member. The crate is named `e2e`: its `src/` is the driver library, and its `tests/` directory is reserved for the scenario suite.
- Depends on `spider-client` and `spider-core` for the gRPC and core types, `tokio` (`sync`/`time` for the concurrency primitives and polling), `tonic` (to build the endpoint), and `anyhow` for the high-level error surface.

## Scenario types (`src/types.rs`)

- `TerminationResult` — the terminal outcome a scenario converges to: `Success(Vec<TaskOutput>)` carrying the job's task outputs, `Failure(String)` carrying the reported error message, or `Cancelled`.
- `JobSubmission` — a single job to drive: an external resource-group id (`String`), the `TaskGraph`, and the job-level `inputs` (`Vec<TaskInput>`).

## The driver (`src/test_driver.rs`)

- `SpiderTestDriver` is a lazily initialized, process-wide singleton (`tokio::sync::OnceCell`). On first access it reads the Spider endpoint from `SPIDER_ENDPOINT` and the concurrency from `SPIDER_CONCURRENCY` (defaulting to `8`), then connects a `SpiderClient` pool sized to that concurrency. It holds the client behind a `RwLock`, a `Semaphore` sized to the concurrency, and a `Mutex<HashMap<String, ResourceGroupId>>` for resource-group resolution.
- `SpiderTestDriver::run` is the non-exclusive entry point: it acquires the client's **shared** read lock and a concurrency permit from the semaphore, then drives the scenario with no failure injection. Multiple `run` scenarios execute concurrently up to the configured limit.
- `SpiderTestDriver::run_exclusive` is the exclusive entry point: it acquires the client's **write** lock so no other scenario runs concurrently, and drives the scenario with a caller-supplied failure-injection coroutine, which is what a failure-recovery scenario uses to perturb a running job.
- In both cases the acquired lock (and, for `run`, the permit) is held for the entire scenario; the client is cloned under the lock and passed by value into the scenario runner.
- Resource groups are resolved outside the scenario runner: the driver maps each external resource-group id to its Spider-assigned `ResourceGroupId`, registering it through the client on first use (with an empty password) and caching the result under the table lock, so subsequent submissions reuse the same id.
- `run_scenario` is the shared core: it submits and starts the job, then concurrently runs the failure-injection coroutine while polling the job to a terminal state, bounding the whole race by `timeout`, and finally hands the `TerminationResult` to the caller's outcome-assertion coroutine.
- `submit_and_start_job` performs both the submit and the start of the job through the client.
- `poll_until_terminal` polls the job state every 10ms via `get_job_state`. On a terminal state it builds the corresponding `TerminationResult`: `Success` retrieves the job outputs, `Failure` retrieves the job error message, and `Cancelled` carries no payload.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **New Features**
  * Added an end-to-end testing harness for running jobs against live deployments.
  * Added shared job submission and termination result types for test scenarios.
  * Added support for concurrent and exclusive test execution.
  * Added configurable endpoint, concurrency, timeout, failure-injection, and job outcome handling.


LinZhihao-723 (commented):
Let's merge #377 first and integrate NN-testing into this crate after that.

tests/huntsman/e2e/src/test_driver.rs
```diff
@@ -0,0 +1,336 @@
...
+///
+/// # Errors
+///
+/// Returns an error if:
+///
+/// * [`anyhow::Error`] if the concurrency value is invalid.
+///
+/// # Panics
+///
+/// Panics if the default value is zero.
```
sitaowang1998: For most libraries/software I have used, 0 for concurrency setting usually means using the number of CPUs as the concurrency level.

tests/huntsman/e2e/src/test_driver.rs
```diff
@@ -0,0 +1,336 @@
...
+fn read_concurrency() -> anyhow::Result<NonZeroUsize> {
+    const CONCURRENCY_ENV_VAR: &str = "SPIDER_CONCURRENCY";
+    const DEFAULT_CONCURRENCY: usize = 8;
+
+    match std::env::var(CONCURRENCY_ENV_VAR) {
+        Ok(value) => value.parse::<NonZeroUsize>().with_context(|| {
```
sitaowang1998, line 327: We should guard against negative values and a crazily-large value.

LinZhihao-723: - Parsing a negative number should fail 'cuz the given type is unsigned.
- I don't think it's necessary for a tester.

tests/huntsman/e2e/src/test_driver.rs
LinZhihao-723: I should document this clearly: the default value is a compile-time constant which should never be 0.

sitaowang1998 (approved):

PR #381, opened 2026-07-08
build(huntsman): Add local build tasks for Spider Huntsman images.

20001020ycx:
# Description

The Taskfile wraps the `docker build` for each Spider service image, following CLP's approach — in particular the image-ID (iid) handling introduced in [y-scope/clp#1335][clp-1335]. For each service, `tools/docker/build.sh`:
  - builds the matching stage of the multi-stage `tools/docker/Dockerfile`,
  - records the built image's digest (iid) to `build/spider-<service>-image.id`,
  - tags the image `spider-<service>:dev-<user>-<short-id>`.
 
The tag naming convention exists to avoid conflicts; that discussion can be traced in [y-scope/clp#1335][clp-1335] and related issues, so we won't expand on it here.

Below, I would like to clarify how docker compose consumes the images this taskfile produces, and how a compose author should resolve the image reference when spinning up the Spider services.

## How docker-compose consumes the local build images

The docker compose file references each service through an env var; an env script can resolves it from the `.id` file for a local build. Using `storage` as an example:

```yaml
services:
  storage:
    image: "${SPIDER_STORAGE_IMAGE_REF:-spider-storage}"
    # The images bake no ENTRYPOINT, so the command names the service binary.
    command: ["spider_storage_grpc_server", "--config", "/etc/spider/storage.yaml"]
```

End to end use case:

```shell
# 1. Build; writes build/spider-storage-image.id (and the other two).
task docker:build

# 2. Point compose at the exact image just built (the env script does this on every `up`):
export SPIDER_STORAGE_IMAGE_REF="$(cat build/spider-storage-image.id)"

# 3. Bring the stack up.
docker compose up
```

# Checklist

  breaking change.

# Validation performed

* `task docker:build-storage`: the image builds, the digest in `build/spider-storage-image.id` matches the built image's ID, the image is tagged `spider-storage:dev-<user>-<short-id>`, and the previously built image is pruned.
* `docker run "$(cat build/spider-storage-image.id)" spider-storage --help` runs, as the non-root `uid=1000(spider-user)`.

[clp-build-sh]: https://github.com/y-scope/clp/blob/main/tools/docker-images/clp-package/build.sh
[clp-1335]: https://github.com/y-scope/clp/pull/1335

## Summary by CodeRabbit

* **New Features**
  * Added Taskfile support for building Docker images for the storage, scheduler, and worker services.
  * Introduced a unified build task to build all service images in sequence.
  * Added a shared Docker build helper with stricter build safety, component-specific builds, and consistent local tagging.
  * Automatically removes the previously built local image for the selected service when a new image is produced.


tools/docker/build.sh
```diff
@@ -0,0 +1,51 @@
...
+    rm -f "$temp_iid_file"
+
+    [[ -z "$prev_image_id" || "$prev_image_id" == "$new_image_id" ]] && return
+
+    docker image inspect "$prev_image_id" >/dev/null 2>&1 || return
+
+    echo "Removing previous image $prev_image_id."
+    docker image remove "$prev_image_id" || true
+}
+trap remove_temp_file_and_prev_image EXIT
```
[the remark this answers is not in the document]
20001020ycx: @junhaoliao should we add `-z "$prev_image_id"` to the guard? Does this mean CLP also have the same bug?

sitaowang1998 (commented):
Before going into details, I have a high level question: Why use a script to manage the download instead of using tasks?

20001020ycx:
> Before going into details, I have a high level question: Why use a script to manage the download instead of using tasks?

This is how clp handles it at https://github.com/y-scope/clp/blob/main/tools/docker-images/clp-package/build.sh, we could it do it in task.yaml but it will be rewriting this build.sh in the yaml itself, which raises the concern in readability.

taskfiles/docker.yaml
```diff
@@ -0,0 +1,23 @@
...
+tasks:
+  build:
+    cmds:
+      - task: "build-storage"
+      - task: "build-scheduler"
+      - task: "build-worker"
+
+  build-storage:
+    deps: [":init"]
+    cmd: "{{.G_BUILD_SCRIPT}} storage"
```
junhaoliao:
```suggestion
    cmd: "'{{.G_BUILD_SCRIPT}}' storage"
```

taskfiles/docker.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+  build-scheduler:
+    deps: [":init"]
+    cmd: "{{.G_BUILD_SCRIPT}} scheduler"
```
junhaoliao:
```suggestion
    cmd: "'{{.G_BUILD_SCRIPT}}' scheduler"
```

taskfiles/docker.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+  build-worker:
+    deps: [":init"]
+    cmd: "{{.G_BUILD_SCRIPT}} worker"
```
junhaoliao:
```suggestion
    cmd: "'{{.G_BUILD_SCRIPT}}' worker"
```

sitaowang1998:
From the perspective of a user, I only need the image name to launch the containers in Docker Compose. If I was given only the iid file, I need to parse it and regenerate the image name. The best output of image building for use in Docker Compose is probably a file containing the image names.

tools/docker/build.sh
```diff
@@ -0,0 +1,51 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+service="${1:?Usage: build.sh <storage|scheduler|worker>}"
+
+remove_temp_file_and_prev_image() {
+    rm -f "$temp_iid_file"
+
+    [[ -z "$prev_image_id" || "$prev_image_id" == "$new_image_id" ]] && return
```
junhaoliao:
```suggestion
    if [[ -z "$new_image_id" ]]; then
        rm -f "$iid_file"
    elif [[ "$prev_image_id" == "$new_image_id" ]]; then
        return
    fi

    [[ -z "$prev_image_id" ]] && return
```

tools/docker/build.sh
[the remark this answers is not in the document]
junhaoliao: right, i agree that:
1. since this is potentially a new build, we should always clean the older image if the new build is faulty, just so we don't mislead developers running services with older images
2. if the image has been cleaned, we should not leave a stale iid file

i proposed https://github.com/y-scope/spider/pull/381/changes#r3553514597 instead then

junhaoliao (changes requested):
for the title, how about simply:

```
build(docker): Add local build tasks for Spider images.
```

(personally i feel the unique tag is some implementation detail that is less important, but feel free to keep it if you think otherwise

tools/docker/build.sh
junhaoliao: we should back port to the CLP repo if you agree with the strategy

20001020ycx:
> From the perspective of a user, I only need the image name to launch the containers in Docker Compose. If I was given only the iid file, I need to parse it and regenerate the image name. The best output of image building for use in Docker Compose is probably a file containing the image names.

My bad in my explanation with you during offline, please look at the example in the PR description. With the iid file, no parsing is required as you can just do `export SPIDER_STORAGE_IMAGE_REF="$(cat build/spider-storage-image.id)"`. `docker compose` uses `SPIDER_STORAGE_IMAGE_REF` as the image tag.

Therefore, you dont need to reformat it to something like `spider-storage-dev-$USER-iid` in docker compose, this format shows up in `docker images` for human readability.

junhaoliao (approved):
note there's also a pr title proposal at https://github.com/y-scope/spider/pull/381#pullrequestreview-4662981302

taskfiles/docker.yaml
```diff
@@ -0,0 +1,23 @@
...
+    deps: [":init"]
+    cmd: "'{{.G_BUILD_SCRIPT}}' storage"
+
+  build-scheduler:
+    deps: [":init"]
+    cmd: "'{{.G_BUILD_SCRIPT}}' scheduler"
+
+  build-worker:
+    deps: [":init"]
+    cmd: "'{{.G_BUILD_SCRIPT}}' worker"
```
LinZhihao-723, line 23: Shall we consider making these tasks private?

LinZhihao-723 (approved):
A nit comment which I'm not sure if it's already a decision from Junhao.
Otherwise lgtm.

taskfiles/docker.yaml
20001020ycx: I think public also makes sense, developer may just run `task docker:build-storage` for faster docker build.

20001020ycx: @junhaoliao what do you think? :)

I am fine with both

sitaowang1998 (approved):

PR #383, opened 2026-07-09
build(helm): Set up Spider Huntsman Helm chart scaffolding and linting.

20001020ycx:
# Description

See PR title.

> [!NOTE]  
This PR is part of the ongoing work Spider Huntsman k8s integration. More PRs will be submitted until the Helm chart is complete and fully functional for Spider Huntsman.

## Note for reviewer:

The taskfile implementation is a close copy from CLP:
- `lint:helm` ← [`clp/taskfiles/lint.yaml#L59`][clp-lint]
- `toolchains:helm` ← [`clp/taskfiles/toolchains.yaml#L24`][clp-toolchains]

The only intentional deviations are the task aliases (`helm-check`/`helm-fix`, to match Spider's `<lang>-check`/`<lang>-fix` naming).

# Checklist

  breaking change.

# Validation performed
- `helm lint` and `yamllint --strict` pass on the committed YAML.
- `task lint:helm-check` resolves the full `ct lint` pipeline and passes.
- `task lint:check` passes.

[clp-lint]: https://github.com/y-scope/clp/blob/251510ab3b41ef4f64ac2435879dcc11e4b49e5b/taskfiles/lint.yaml#L59
  [clp-toolchains]: https://github.com/y-scope/clp/blob/251510ab3b41ef4f64ac2435879dcc11e4b49e5b/taskfiles/toolchains.yaml#L24

## Summary by CodeRabbit

* **New Features**
  * Added an initial Helm chart for deploying the Spider application, including chart metadata and packaging exclusions.
* **Developer Tools**
  * Extended lint workflows to validate Helm charts.
  * Added a dedicated Helm toolchain setup to run chart-testing and install Helm 4 consistently.
  * Enabled safer pipeline failure propagation by turning on `pipefail`.
* **Chores**
  * Added Yamale to development dependencies.

### DIFF
--- a/.github/workflows/code-linting-checks.yaml
+++ b/.github/workflows/code-linting-checks.yaml
@@ -53,6 +53,8 @@ jobs:
 
       - run: "task lint:yml-check"
 
+      - run: "task lint:helm-check"
+
   lint-cpp:
     needs: "filter-relevant-changes"
     if: >-
--- a/.github/workflows/code-linting-checks.yaml
+++ b/.github/workflows/code-linting-checks.yaml
@@ -53,7 +53,7 @@ jobs:
 
       - run: "task lint:yml-check"
 
-      - run: "task lint:helm-check"
+      - run: "task lint:check-helm"
 
   lint-cpp:
     needs: "filter-relevant-changes"
--- a/.github/workflows/code-linting-checks.yaml
+++ b/.github/workflows/code-linting-checks.yaml
@@ -38,6 +38,7 @@ jobs:
     steps:
       - uses: "actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd"  # v6.0.2
         with:
+          fetch-depth: 0
           submodules: "recursive"
 
       - uses: "./tools/yscope-dev-utils/exports/github/actions/install-python"
--- a/taskfiles/lint.yaml
+++ b/taskfiles/lint.yaml
@@ -136,6 +136,7 @@ tasks:
         . "{{.G_LINT_VENV_DIR}}/bin/activate"
         . "{{.G_HELM_TOOLCHAIN_ENV_FILE}}"
         ct lint \
+          --all \
           --chart-dirs="tools/deployment" \
           --chart-yaml-schema="{{.G_HELM_TOOLCHAIN_DIR}}/etc/chart_schema.yaml" \
           --lint-conf="tools/yscope-dev-utils/exports/lint-configs/.yamllint.yml"
--- a/taskfiles/lint.yaml
+++ b/taskfiles/lint.yaml
@@ -15,7 +15,7 @@ tasks:
 
       - task: "cmake-check"
       - task: "cpp-check"
-      - task: "helm-check"
+      - task: "check-helm"
       - task: "py-check"
       - task: "toml-check"
       - task: "yml-check"
@@ -26,7 +26,7 @@ tasks:
 
       - task: "cmake-fix"
       - task: "cpp-fix"
-      - task: "helm-fix"
+      - task: "fix-helm"
       - task: "py-fix"
       - task: "toml-fix"
       - task: "yml-fix"
@@ -126,8 +126,8 @@ tasks:
 
   helm:
     aliases:
-      - "helm-check"
-      - "helm-fix"
+      - "check-helm"
+      - "fix-helm"
     deps:
       - "toolchains:helm"
       - "venv"
--- a/taskfiles/lint.yaml
+++ b/taskfiles/lint.yaml
@@ -136,7 +136,6 @@ tasks:
         . "{{.G_LINT_VENV_DIR}}/bin/activate"
         . "{{.G_HELM_TOOLCHAIN_ENV_FILE}}"
         ct lint \
-          --all \
           --chart-dirs="tools/deployment" \
           --chart-yaml-schema="{{.G_HELM_TOOLCHAIN_DIR}}/etc/chart_schema.yaml" \
           --lint-conf="tools/yscope-dev-utils/exports/lint-configs/.yamllint.yml"
--- a/taskfiles/lint.yaml
+++ b/taskfiles/lint.yaml
@@ -15,6 +15,7 @@ tasks:
 
       - task: "cmake-check"
       - task: "cpp-check"
+      - task: "helm-check"
       - task: "py-check"
       - task: "toml-check"
       - task: "yml-check"
@@ -25,6 +26,7 @@ tasks:
 
       - task: "cmake-fix"
       - task: "cpp-fix"
+      - task: "helm-fix"
       - task: "py-fix"
       - task: "toml-fix"
       - task: "yml-fix"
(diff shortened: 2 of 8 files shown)


taskfiles/lint.yaml
```diff
@@ -15,6 +15,7 @@ tasks:
 
       - task: "cmake-check"
       - task: "cpp-check"
+      - task: "helm-check"
```
sitaowang1998: My bad. `xxx-check` is an obsolete convention. The latest convention is to use `check-xxx`. For example `check-rust`. I am ok with both in this PR. @LinZhihao-723 Shall we create a cleanup PR after this one?

tools/deployment/spider-helm/.helmignore
```diff
@@ -0,0 +1,23 @@
...
+.svn/
+# Common backup files
+*.bak
+*.orig
+*.swp
+*.tmp
+*~
+# Various IDEs
+*.tmproj
+.idea/
```
sitaowang1998, line 21: For RustRover, we probably need to ignore `target/`.

taskfiles/lint.yaml
20001020ycx: okay, thanks for the catch, good to know

taskfiles/toolchains.yaml
```diff
@@ -77,3 +80,39 @@ tasks:
...
+            {{- else -}}
+              unsupported
+            {{- end }}
+          INCLUDE_PATTERNS: ["ct", "etc/chart_schema.yaml"]
+          NUM_COMPONENTS_TO_STRIP: "0"
+      - |-
+        export PATH="{{.G_HELM_TOOLCHAIN_DIR}}:$PATH"
+        export HELM_INSTALL_DIR="{{.G_HELM_TOOLCHAIN_DIR}}"
+        curl --proto "=https" --tlsv1.2 --silent --show-error --fail \
+        https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-4 | bash -s -- --no-sudo
```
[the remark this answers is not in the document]
20001020ycx: I think rabbit is making a reasonable point, @junhaoliao, we should also update this in CLP repo as well, I will go ahead and use the latest version for now.

tools/deployment/spider-helm/.helmignore
20001020ycx: Actually, just dived a bit more on the .helmignore: this ignore file is controlling what files are put into the `.tgz` during helm package, the `.tgz` file is the artifact distributed to k8s user.

So, `helm package` only reads the chart directory and re-bundles its own files into a tarball — it compiles nothing and creates no target/ and no build/. So target/ is not, and never will be, a product of helm package.

I will just use the version inherited from the CLP package for consistency then.

taskfiles/toolchains.yaml
```diff
@@ -77,3 +80,41 @@ tasks:
         vars:
           CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
           INCLUDE_PATTERNS: *output_patterns
+
+  helm:
```
junhaoliao: can we place `helm` before `rust` and order its attributes as variables, environment control, then execution control? that matches CLP and the Taskfile contribution guide, which will make the copied setup easier to compare and maintain

taskfiles/toolchains.yaml
```diff
@@ -77,3 +80,41 @@ tasks:
...
+    run: "once"
```
junhaoliao: Spider does not currently enable `pipefail` globally, while CLP sets it at the root. can we add `set: ["pipefail"]` to the root `taskfile.yaml` instead of scoping it to this installer task? that protects this download pipeline and the other Taskfile pipelines consistently, without also opting Spider into CLP’s separate `u` setting.

taskfiles/lint.yaml
```diff
@@ -171,6 +173,22 @@ tasks:
           taskfile.yaml \
           taskfiles/
 
+  helm:
```
junhaoliao: can we move this unchanged block to immediately before `py-check`? keeping the new non-internal tool linter with the C++/Python lint tasks follows the Taskfile guide's logical ordering and avoids leaving it between `yml` and `toml`

taskfiles/lint.yaml
junhaoliao: (maybe we have planned for it - i'm fine if we want to submit separately

can we call this in code-linting-checks.yaml ?

tools/deployment/spider-helm/values.yaml
```diff
@@ -0,0 +1,2 @@
+nameOverride: ""
+fullnameOverride: ""
```
junhaoliao: mind explaining what those are and how they are used?

junhaoliao (changes requested):
for the title, how about

```
build(helm): Set up Spider Helm chart scaffolding and linting.
```

20001020ycx:
> for the title, how about
> 
> ```
> build(helm): Set up Spider Helm chart scaffolding and linting.
> ```

LOL, this was the initial one that ai suggested. Sure, I will just add Huntsman for clarity then

tools/deployment/spider-helm/values.yaml
20001020ycx: The values.yaml here is only a placeholder so ct lint has a values file to validate against — linting fails if the file is missing, but its content is irrelevant.

nameOverride/fullnameOverride aren't required for linting and nothing in the chart references them yet, so I've dropped them and left the file empty. I will add the actual content in the follow-up PR.

taskfiles/lint.yaml
20001020ycx: Wow, this is a great catch, I thought it is already cover by some high level linting like `task lint:check`. Updated the code-linting-checks.yaml in the new version. Thanks

junhaoliao (dismissed):

.github/workflows/code-linting-checks.yaml
junhaoliao, line 42: sorry for missing this - 

do we need `fetch-depth: 0` for helm check to work?

taskfiles/lint.yaml
LinZhihao-723: I actually prefer `check-xxx`, as we should always use the latest convention when both exist.
fyi, Rust linting is called `fix-rust` and `check-rust`.

.github/workflows/code-linting-checks.yaml
20001020ycx: I think I know what you meant, I believe ct -all shall solve the problem.

I'm assuming you're referring to this error we hit when the workflow ran without --all:
```
Error: failed linting charts: failed identifying charts to process: targetBranch 'origin/main' does not exist
```
--all skips that git-diff with `origin/main` and lints every chart directly, as you can see, lint-common passes are green now.

junhaoliao: i see. do you know if the version bump check would also need to use git diff?

in fact, the `fetch-depth: 0` option is also specified in the official example: https://github.com/helm/chart-testing-action#:~:text=with%3A%0A%20%20%20%20%20%20%20%20%20%20fetch%2Ddepth%3A%200

either way, the CLP repo will not use the `--all` option because it will integrate the spider helm chart as a subchart. then for consistency i think it's better to follow the same in spider?

LinZhihao-723 (dismissed):

junhaoliao (approved):

PR #384, opened 2026-07-09
feat(helm): Add storage and bundled-MariaDB templates.

20001020ycx:
# Description

This PR adds the chart's **templates** and **`values.yaml`**: the storage gRPC service (Deployment + Service + ConfigMap) and a bundled MariaDB (StatefulSet + Service + Secret). Remove `"database"` from `bundled` to connect to an external MariaDB instead.

> [!NOTE]  
Part of the ongoing Spider Huntsman Kubernetes integration. This PR adds only storage + database; the scheduler, execution manager, and task executor follow in later PRs.

> [!NOTE]
> This PR needs review from two teams:
>
> - Cloud Team: Please review the Helm chart itself — the templates, helpers, and packaging. Feel free to also skim through the configuration itself.
> - Spider Team: Please review the user-facing Spider configuration — the `spiderConfig` block in values.yaml and the storage.yaml it renders in configmap.yaml. You can leave the rest of the chart to the Cloud reviewer.

## A few notes for the reviewers:

For Cloud reviewer:
- ` _helpers.tpl` is mostly helm create boilerplate. We only added: componentFullname (63-char-safe names), imageRef, readiness/livenessProbeTimings, and databaseHost/databasePort.

For Spider reviwer:
- `spiderConfig` in values.yaml is the user-facing config in helm chart; the chart renders it into each component's config file. Naming loosely follows the sample [spider.yaml](https://github.com/sitaowang1998/spider/blob/huntsman-dev-e2e/tools/scripts/stack/spider.yaml). I would appreciate any suggestions to make it more aligning with the real Spider configuration workflow.

# Checklist

  breaking change.

# Validation performed

We perform the steps below to verify the storage + database Helm chart in this PR works end-to-end in a `kind` cluster. The goal: submit a task through the `spider-client` Rust SDK and observe it persisted in the database.

1. Start a `kind` cluster:
```
kind create cluster --name spider-test
```

2. Install the chart:
```
helm install spider tools/deployment/spider-helm/
```

3. Wait for both components to become ready (storage restarts a few times while MariaDB boots):
```
kubectl wait --for=condition=ready pod -l app.kubernetes.io/name=spider --timeout=180s
```

4. Confirm storage self-created its tables:
```
kubectl exec spider-database-0 -- mysql -uspider-user -pspider-password spider-db -e "SHOW TABLES;"
```
Expected: `execution_managers`, `jobs`, `resource_groups`, `schedulers`, `sessions`.

5. Port-forward storage and submit a task through the `spider-client` SDK, the below code are phsedu-code for illustration purpose:
```
kubectl port-forward svc/spider-storage 50051:50051
```
The client submits a one-task graph — `math::add(2, 3)`:
```rust
let client = SpiderClient::connect(endpoint, pool_size).await?;             // dial storage
let rg = client.add_resource_group("add-rg", b"secret".to_vec()).await?;    // register a resource group

let mut graph = TaskGraph::new(None, None)?;                                // one-task graph
graph.insert_task(TaskDescriptor {                                          // TDL func math::add(i32, i32) -> i32
    tdl_context: TdlContext { package: "math", task_func: "add" },
    inputs: vec![int32, int32], outputs: vec![int32], ..Default::default()
})?;
let inputs = vec![TaskInput::ValuePayload(rmp_serde::to_vec(&2)?),          // msgpack-encode the operands
                  TaskInput::ValuePayload(rmp_serde::to_vec(&3)?)];
let job_id = client.submit_job(rg, &graph, inputs).await?;                  // storage persists the job
```

6. Confirm the task is persisted in the database:
```
kubectl exec spider-database-0 -- mysql -uspider-user -pspider-password spider-db -e "SELECT COUNT(*) FROM jobs; SELECT COUNT(*) FROM resource_groups;"
```
Expected: `jobs=1`, `resource_groups=1`.

## Summary by CodeRabbit

* **New Features**
  * Added Helm deployment support for Spider storage, including ConfigMap-backed configuration plus a dedicated storage Deployment and gRPC Service.
  * Added optional bundled MariaDB support, including automatic database credentials Secret creation, a headless database Service, and a database StatefulSet with persistent storage.
  * Introduced chart values for image sources/tags, storage endpoint settings, connection limits, and default queue capacity.
* **Chores**
  * Bumped the Helm chart version to **0.1.1**.

### DIFF
--- a/tools/deployment/spider-helm/templates/_helpers.tpl
+++ b/tools/deployment/spider-helm/templates/_helpers.tpl
@@ -1,0 +1,129 @@
+{{/*
+Expands the name of the chart.
+
+@return {string} The chart name (truncated to 63 characters)
+*/}}
+{{- define "spider.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Creates a default fully qualified app name (truncated to 63 chars for the DNS naming spec). If the
+release name already contains the chart name it is used as-is.
+
+@return {string} The fully qualified app name (truncated to 63 characters)
+*/}}
+{{- define "spider.fullname" -}}
+{{- if .Values.fullnameOverride }}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- $name := default .Chart.Name .Values.nameOverride }}
+{{- if contains $name .Release.Name }}
+{{- .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
+{{- end }}
+
+{{/*
+Creates chart name and version as used by the chart label.
+
+@return {string} Chart name and version (truncated to 63 characters)
+*/}}
+{{- define "spider.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Creates common labels for all resources.
+
+@return {string} YAML-formatted common labels
+*/}}
+{{- define "spider.labels" -}}
+helm.sh/chart: {{ include "spider.chart" . }}
+{{ include "spider.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Creates selector labels for matching resources.
+
+@return {string} YAML-formatted selector labels
+*/}}
+{{- define "spider.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "spider.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
+Creates a container image reference from `.Values.image.<component>`.
+
+The tag defaults to the chart's appVersion when the component omits it.
+
+@param {object} root Root template context (required)
+@param {string} component Key under `.Values.image` (e.g. "storage", "mariadb")
+@return {string} Full image reference (repository:tag)
+*/}}
+{{- define "spider.imageRef" -}}
+{{- $img := index .root.Values.image .component -}}
+{{- $tag := $img.tag | default .root.Chart.AppVersion -}}
+{{- printf "%s:%s" $img.repository $tag -}}
+{{- end }}
+
+{{/*
+Creates timings for readiness probes (faster checks for quicker startup).
+
+@return {string} YAML-formatted readiness probe timing configuration
+*/}}
+{{- define "spider.readinessProbeTimings" -}}
+initialDelaySeconds: 6
+periodSeconds: 2
+timeoutSeconds: 2
+failureThreshold: 10
+{{- end }}
+
+{{/*
+Creates timings for liveness probes.
+
+@return {string} YAML-formatted liveness probe timing configuration
+*/}}
+{{- define "spider.livenessProbeTimings" -}}
+initialDelaySeconds: 180
+periodSeconds: 30
+timeoutSeconds: 4
+failureThreshold: 3
+{{- end }}
+
+{{/*
+Gets the database host. When "database" is bundled, this is the in-cluster MariaDB Service; when not
+bundled, it is the external `spiderConfig.database.host`.
+
+@param {object} . Root template context
+@return {string} The database host
+*/}}
+{{- define "spider.databaseHost" -}}
+{{- if has "database" .Values.spiderConfig.bundled -}}
+{{- printf "%s-database" (include "spider.fullname" .) -}}
+{{- else -}}
+{{- .Values.spiderConfig.database.host -}}
+{{- end -}}
+{{- end }}
+
+{{/*
+Gets the database port. When "database" is bundled, this is 3306 (the bundled MariaDB's port);
+otherwise it is the external `spiderConfig.database.port`.
+
+@param {object} . Root template context
+@return {string} The database port
+*/}}
+{{- define "spider.databasePort" -}}
+{{- if has "database" .Values.spiderConfig.bundled -}}
+3306
+{{- else -}}
+{{- .Values.spiderConfig.database.port -}}
+{{- end -}}
+{{- end }}
--- a/tools/deployment/spider-helm/templates/_helpers.tpl
+++ b/tools/deployment/spider-helm/templates/_helpers.tpl
@@ -22,9 +22,9 @@ release name already contains the chart name it is used as-is.
 {{- .Release.Name | trunc 63 | trimSuffix "-" }}
 {{- else }}
 {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
-{{- end }}
-{{- end }}
-{{- end }}
+{{- end }}{{/* if contains $name .Release.Name */}}
+{{- end }}{{/* if .Values.fullnameOverride */}}
+{{- end }}{{/* define "spider.fullname" */}}
 
 {{/*
 Creates chart name and version as used by the chart label.
@@ -99,8 +99,7 @@ failureThreshold: 3
 {{- end }}
 
 {{/*
-Gets the database host. When "database" is bundled, this is the in-cluster MariaDB Service; when not
-bundled, it is the external `spiderConfig.database.host`.
+Gets the bundled MariaDB Service host or external `spiderConfig.database.host`.
 
 @param {object} . Root template context
 @return {string} The database host
@@ -114,8 +113,7 @@ bundled, it is the external `spiderConfig.database.host`.
 {{- end }}
 
 {{/*
-Gets the database port. When "database" is bundled, this is 3306 (the bundled MariaDB's port);
-otherwise it is the external `spiderConfig.database.port`.
+Gets the bundled MariaDB port or external `spiderConfig.database.port`.
 
 @param {object} . Root template context
 @return {string} The database port
--- a/tools/deployment/spider-helm/templates/_helpers.tpl
+++ b/tools/deployment/spider-helm/templates/_helpers.tpl
@@ -27,6 +27,20 @@ release name already contains the chart name it is used as-is.
 {{- end }}{{/* define "spider.fullname" */}}
 
 {{/*
+Creates a fully qualified component name while preserving its suffix within the 63-character limit.
+
+@param {object} root Root template context (required)
+@param {string} component Component name suffix (required)
+@return {string} The fully qualified component name
+*/}}
+{{- define "spider.componentFullname" -}}
+{{- $suffix := printf "-%s" .component -}}
+{{- $maxBaseLength := sub 63 (len $suffix) | int -}}
+{{- $base := include "spider.fullname" .root | trunc $maxBaseLength | trimSuffix "-" -}}
+{{- printf "%s%s" $base $suffix -}}
+{{- end }}
+
+{{/*
 Creates chart name and version as used by the chart label.
 
 @return {string} Chart name and version (truncated to 63 characters)
@@ -106,7 +120,7 @@ Gets the bundled MariaDB Service host or external `spiderConfig.database.host`.
 */}}
 {{- define "spider.databaseHost" -}}
 {{- if has "database" .Values.spiderConfig.bundled -}}
-{{- printf "%s-database" (include "spider.fullname" .) -}}
+{{- include "spider.componentFullname" (dict "root" . "component" "database") -}}
 {{- else -}}
 {{- .Values.spiderConfig.database.host -}}
 {{- end -}}
--- a/tools/deployment/spider-helm/templates/_helpers.tpl
+++ b/tools/deployment/spider-helm/templates/_helpers.tpl
@@ -74,18 +74,22 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
 
 {{/*
-Creates a container image reference from `.Values.image.<component>`.
+Creates a container image reference from .Values.image.
 
-The tag defaults to the chart's appVersion when the component omits it.
+Renders repository@digest when "digest" is set; otherwise, requires "tag" and renders repository:tag.
 
 @param {object} root Root template context (required)
-@param {string} component Key under `.Values.image` (e.g. "storage", "database")
-@return {string} Full image reference (repository:tag)
+@param {string} component Key under .Values.image (e.g., "storage", "database")
+@return {string} Full image reference (repository@digest or repository:tag)
 */}}
 {{- define "spider.imageRef" -}}
 {{- $img := index .root.Values.image .component -}}
-{{- $tag := $img.tag | default .root.Chart.AppVersion -}}
+{{- if $img.digest -}}
+{{- printf "%s@%s" $img.repository $img.digest -}}
+{{- else -}}
+{{- $tag := required (printf "image.%s.tag is required" .component) $img.tag -}}
 {{- printf "%s:%s" $img.repository $tag -}}
+{{- end -}}
 {{- end }}
 
 {{/*
--- a/tools/deployment/spider-helm/templates/_helpers.tpl
+++ b/tools/deployment/spider-helm/templates/_helpers.tpl
@@ -79,7 +79,7 @@ Creates a container image reference from `.Values.image.<component>`.
 The tag defaults to the chart's appVersion when the component omits it.
 
 @param {object} root Root template context (required)
-@param {string} component Key under `.Values.image` (e.g. "storage", "mariadb")
+@param {string} component Key under `.Values.image` (e.g. "storage", "database")
 @return {string} Full image reference (repository:tag)
 */}}
 {{- define "spider.imageRef" -}}
@@ -113,7 +113,7 @@ failureThreshold: 3
 {{- end }}
 
 {{/*
-Gets the bundled MariaDB Service host or external `spiderConfig.database.host`.
+Gets the bundled database Service host or external `spiderConfig.database.host`.
 
 @param {object} . Root template context
 @return {string} The database host
@@ -127,7 +127,7 @@ Gets the bundled MariaDB Service host or external `spiderConfig.database.host`.
 {{- end }}
 
 {{/*
-Gets the bundled MariaDB port or external `spiderConfig.database.port`.
+Gets the bundled database port or external `spiderConfig.database.port`.
 
 @param {object} . Root template context
 @return {string} The database port
--- a/tools/deployment/spider-helm/templates/configmap.yaml
+++ b/tools/deployment/spider-helm/templates/configmap.yaml
@@ -1,0 +1,21 @@
+apiVersion: "v1"
+kind: "ConfigMap"
+metadata:
+  name: {{ include "spider.fullname" . }}-config
+  labels:
+    {{- include "spider.labels" . | nindent 4 }}
+data:
+  storage.yaml: |
+    host: "0.0.0.0"
+    port: {{ .Values.spiderConfig.storage_endpoint.port }}
+    runtime:
+      db_config:
+        host: {{ include "spider.databaseHost" . | quote }}
+        port: {{ include "spider.databasePort" . }}
+        name: {{ .Values.spiderConfig.database.name | quote }}
+        username: {{ .Values.spiderConfig.database.username | quote }}
+        password: {{ .Values.spiderConfig.database.password | quote }}
+        max_connections: {{ .Values.spiderConfig.storage.max_connections }}
+      {{- with omit .Values.spiderConfig.storage "max_connections" }}
+      {{- toYaml . | nindent 6 }}
+      {{- end }}
--- a/tools/deployment/spider-helm/templates/configmap.yaml
+++ b/tools/deployment/spider-helm/templates/configmap.yaml
@@ -13,7 +13,7 @@ data:
         host: {{ include "spider.databaseHost" . | quote }}
         port: {{ include "spider.databasePort" . }}
         name: {{ .Values.spiderConfig.database.name | quote }}
-        max_connections: {{ .Values.spiderConfig.storage.max_connections }}
-      {{- with omit .Values.spiderConfig.storage "max_connections" }}
+        max_connections: {{ .Values.spiderConfig.database.max_connections }}
+      {{- with .Values.spiderConfig.storage }}
       {{- toYaml . | nindent 6 }}
       {{- end }}
--- a/tools/deployment/spider-helm/templates/configmap.yaml
+++ b/tools/deployment/spider-helm/templates/configmap.yaml
@@ -1,7 +1,7 @@
 apiVersion: "v1"
 kind: "ConfigMap"
 metadata:
-  name: {{ include "spider.fullname" . }}-config
+  name: {{ include "spider.componentFullname" (dict "root" . "component" "config") }}
   labels:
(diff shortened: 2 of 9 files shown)


tools/deployment/spider-helm/templates/_helpers.tpl
```diff
@@ -0,0 +1,129 @@
...
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- $name := default .Chart.Name .Values.nameOverride }}
+{{- if contains $name .Release.Name }}
+{{- .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
+{{- end }}
```
junhaoliao: can we label these closures like CLP's ConfigMap does? the nested scopes are otherwise hard to pair at a glance.

```suggestion
{{- end }}{{/* if contains $name .Release.Name */}}
{{- end }}{{/* if .Values.fullnameOverride */}}
{{- end }}{{/* define "spider.fullname" */}}
```

tools/deployment/spider-helm/templates/_helpers.tpl
```diff
@@ -0,0 +1,129 @@
...
+The tag defaults to the chart's appVersion when the component omits it.
+
+@param {object} root Root template context (required)
+@param {string} component Key under `.Values.image` (e.g. "storage", "mariadb")
+@return {string} Full image reference (repository:tag)
+*/}}
+{{- define "spider.imageRef" -}}
+{{- $img := index .root.Values.image .component -}}
+{{- $tag := $img.tag | default .root.Chart.AppVersion -}}
+{{- printf "%s:%s" $img.repository $tag -}}
```
junhaoliao: can we match Spider's actual image tags? `Chart.AppVersion` is `0.1.0`, but the workflow publishes `main`, SHA, and raw release tags, so neither image should inherit it. require a tag unless a digest is set.

```suggestion
Creates a container image reference from .Values.image.

Renders repository@digest when "digest" is set; otherwise, requires "tag" and renders repository:tag.

@param {object} root Root template context (required)
@param {string} component Key under .Values.image (e.g., "storage", "mariadb")
@return {string} Full image reference (repository@digest or repository:tag)
*/}}
{{- define "spider.imageRef" -}}
{{- $img := index .root.Values.image .component -}}
{{- if $img.digest -}}
{{- printf "%s@%s" $img.repository $img.digest -}}
{{- else -}}
{{- $tag := required (printf "image.%s.tag is required" .component) $img.tag -}}
{{- printf "%s:%s" $img.repository $tag -}}
{{- end -}}
```

tools/deployment/spider-helm/templates/_helpers.tpl
```diff
@@ -0,0 +1,129 @@
...
+timeoutSeconds: 4
+failureThreshold: 3
+{{- end }}
+
+{{/*
+Gets the database host. When "database" is bundled, this is the in-cluster MariaDB Service; when not
+bundled, it is the external `spiderConfig.database.host`.
+
+@param {object} . Root template context
+@return {string} The database host
```
junhaoliao:
```suggestion
Gets the bundled MariaDB Service host or external `spiderConfig.database.host`.

@param {object} . Root template context
@return {string} The database host
```

tools/deployment/spider-helm/templates/_helpers.tpl
```diff
@@ -0,0 +1,129 @@
...
+{{- .Values.spiderConfig.database.host -}}
+{{- end -}}
+{{- end }}
+
+{{/*
+Gets the database port. When "database" is bundled, this is 3306 (the bundled MariaDB's port);
+otherwise it is the external `spiderConfig.database.port`.
+
+@param {object} . Root template context
+@return {string} The database port
```
junhaoliao:
```suggestion
Gets the bundled MariaDB port or external `spiderConfig.database.port`.

@param {object} . Root template context
@return {string} The database port
```

tools/deployment/spider-helm/templates/configmap.yaml
```diff
@@ -0,0 +1,21 @@
...
+  storage.yaml: |
+    host: "0.0.0.0"
+    port: {{ .Values.spiderConfig.storage_endpoint.port }}
+    runtime:
+      db_config:
+        host: {{ include "spider.databaseHost" . | quote }}
+        port: {{ include "spider.databasePort" . }}
+        name: {{ .Values.spiderConfig.database.name | quote }}
+        username: {{ .Values.spiderConfig.database.username | quote }}
+        password: {{ .Values.spiderConfig.database.password | quote }}
```
junhaoliao: CLP keeps these separate: it [creates a database Secret](https://github.com/y-scope/clp/blob/3c6ad2605696f95b657a8eb5dfa8af79f8b6537d/tools/deployment/package-helm/templates/database-secret.yaml#L1-L13), [injects its keys with `secretKeyRef`](https://github.com/y-scope/clp/blob/3c6ad2605696f95b657a8eb5dfa8af79f8b6537d/tools/deployment/package-helm/templates/api-server-deployment.yaml#L38-L49), and [mounts the non-secret ConfigMap separately](https://github.com/y-scope/clp/blob/3c6ad2605696f95b657a8eb5dfa8af79f8b6537d/tools/deployment/package-helm/templates/api-server-deployment.yaml#L87-L90). It does not map a Secret into a ConfigMap.

Spider [already creates the Secret](https://github.com/y-scope/spider/blob/656b17e776ecb9454668cba48f8e0fa7fdfb8c39/tools/deployment/spider-helm/templates/database-secret.yaml#L1-L13), but MariaDB consumes it while storage [reads one complete YAML file](https://github.com/y-scope/spider/blob/656b17e776ecb9454668cba48f8e0fa7fdfb8c39/components/spider-storage/src/bin/grpc_server.rs#L33-L38). Kubernetes cannot substitute Secret keys into ConfigMap data, so this is not only wiring. The chart-only fix is to render `storage.yaml` as a Secret and mount it; alternatively, add env/file credential support and use `secretKeyRef` like CLP.

tools/deployment/spider-helm/templates/storage-deployment.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: "apps/v1"
+kind: "Deployment"
+metadata:
+  name: {{ include "spider.fullname" . }}-storage
+  labels:
+    {{- include "spider.labels" . | nindent 4 }}
+    app.kubernetes.io/component: "storage"
+spec:
+  replicas: 1
```
junhaoliao, line 9: non-blocking: can we create an issue to track CLP-style `resources` and `scheduling` overrides for both workloads before production or multi-node use?

tools/deployment/spider-helm/templates/storage-deployment.yaml
```diff
@@ -0,0 +1,43 @@
...
+    metadata:
+      labels:
+        {{- include "spider.labels" . | nindent 8 }}
+        app.kubernetes.io/component: "storage"
+    spec:
+      containers:
+        - name: "storage"
+          image: {{ include "spider.imageRef" (dict "root" . "component" "storage") | quote }}
+          imagePullPolicy: {{ .Values.image.storage.pullPolicy | quote }}
+          command: ["spider-storage", "--config", "/etc/spider/storage.yaml"]
```
junhaoliao, line 24: Kubernetes stops this container with `SIGTERM`, but storage currently handles only `SIGINT`; rollout can bypass `runtime.stop()`. can we handle `SignalKind::terminate()` in a focused follow-up?

tools/deployment/spider-helm/templates/storage-service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: "v1"
+kind: "Service"
+metadata:
+  name: {{ include "spider.fullname" . }}-storage
```
junhaoliao: a 53-character release produces a 60-character `<release>-spider` fullname; appending `-storage` makes this Service name 68 characters, exceeding Kubernetes' 63-character limit. can we reserve space for the suffix with a helper like this?

```gotemplate
{{/*
Creates a fully qualified component name while preserving its suffix within the 63-character limit.

@param {object} root Root template context (required)
@param {string} component Component name suffix (required)
@return {string} The fully qualified component name
*/}}
{{- define "spider.componentFullname" -}}
{{- $suffix := printf "-%s" .component -}}
{{- $maxBaseLength := sub 63 (len $suffix) | int -}}
{{- $base := include "spider.fullname" .root | trunc $maxBaseLength | trimSuffix "-" -}}
{{- printf "%s%s" $base $suffix -}}
{{- end }}
```

then use it consistently for resource names, `serviceName`, Secret/ConfigMap references, and `databaseHost`

tools/deployment/spider-helm/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+  storage:
+    max_connections: 64
+    # task_instance_pool_config:
+    #   execution_manager_stale_cutoff_sec: 30
+    #   gc_interval_sec: 5
+    #   message_channel_capacity: 1024
+    # job_cache_gc_config:
+    #   terminated_job_retention_sec: 300
+    #   gc_interval_sec: 10
```
junhaoliao: mind explaining what these are for? would it be better to leave them uncommented so they can serve as default (fallback) values?

junhaoliao (changes requested):
the rest of the code and the title lgtm. 

sorry i just caught an issue that might cause helm lint not to work (?) at https://github.com/y-scope/spider/pull/383/changes#r3567518399 . shall we fix it in the same PR / before this PR is merged?

junhaoliao:
let's use scope `feat(helm)` in the title for conciseness and consistency

tools/deployment/spider-helm/templates/storage-deployment.yaml
20001020ycx: Good catch, will implement `SIGTERM` handler in the storage grpc server, as well as other components in Spider Huntsman.

20001020ycx: @coderabbitai, can you create an issue to track what Junhao is mentioning?

tools/deployment/spider-helm/templates/storage-deployment.yaml
20001020ycx: @coderabbitai, can you create an issue to track what Junhao is mentioning?

tools/deployment/spider-helm/values.yaml
20001020ycx: These are advanced configurations supported by Spider, following the convention in [spider.yaml](https://github.com/sitaowang1998/spider/blob/huntsman-dev-e2e/tools/scripts/stack/spider.yaml). I left them commented since values.yaml serves as the default template, they're here to showcase what Spider supports, which user can override for their own k8s deployment.

Note this is contrary to clp-package's per-variable assignment: we copy the entire YAML through to the component for compatibility, which avoids maintaining that many variables and I think is a better practice for the helm chart.

tools/deployment/spider-helm/values.yaml
```diff
@@ -1 +1,47 @@
+nameOverride: ""
+fullnameOverride: ""
 
+image:
+  storage:
+    repository: "ghcr.io/y-scope/spider/storage"
+    pullPolicy: "Always"
+    tag: "main"
```
LinZhihao-723: Does it allow us to switch to different versions using tag?

tools/deployment/spider-helm/values.yaml
```diff
@@ -1 +1,47 @@
...
+  mariadb:
+    repository: "mariadb"
+    pullPolicy: "Always"
+    tag: "10.11.16"
```
LinZhihao-723: Does it allow us to switch to MySQL?
If true, shall we rename `mariadb` to make it DB-type anonymous?

tools/deployment/spider-helm/values.yaml
```diff
@@ -1 +1,47 @@
...
+
+  database:
+    # Host and port for the database service. When "database" is in `bundled`, the host is
+    # automatically set to the bundled service name (ignored). When external, set these to the
+    # external service's address.
+    host: "127.0.0.1"
+    port: 3306
+    username: "spider-user"
+    password: "spider-password"
+    root_password: "spider-root-password"
```
LinZhihao-723, line 30: Might be worth to confirm with @junhaoliao for the best practice of passing DB credentials.

LinZhihao-723: Seems to relate to https://github.com/y-scope/spider/pull/384/changes#r3567189030

tools/deployment/spider-helm/values.yaml
LinZhihao-723: These are configurable options for Spider storage service. In CLP, we might need to make these fields configurable, but I'm not sure what's the best way to do it. For example, shall we directly expose this value file in CLP, or shall we map CLP-side config values to fields defined here? May need @junhaoliao's inputs.

tools/deployment/spider-helm/templates/configmap.yaml
20001020ycx: Good catch, I DID had some concern about the security stuff when discussing with Chaoyue. 

I think now I get the problem, and I think it make sense to just use env variable for consistency with CLP. I will make the rust change now before merging this PR so we dont change back and forth.

tools/deployment/spider-helm/values.yaml
20001020ycx: Yes.  an user can switch to a different version by overriding it at helm install time, e.g. `helm install ... --set image.storage.tag=<version>`, or through your own values file (`helm install ... -f values-prod.yaml`).

Once Spider starts publishing versioned image tags, we should update this default to point at a pinned release version instead of main which gets overwritten by every commit (non-stable).

20001020ycx: And just to be clear on the user's interface:
this `values.yaml` is just the default set for spinning up Spider with minimal setup. A user doing more advanced tuning or a Spider developer wanting to test a specific feature set on k8s would provide their own overrides, as shown above, on top of these defaults.

tools/deployment/spider-helm/values.yaml
20001020ycx: Ditto, answered in https://github.com/y-scope/spider/pull/384#discussion_r3581042875

But I think the renaming make sense, will do.

tools/deployment/spider-helm/templates/configmap.yaml
20001020ycx: Discussed feature is implemented in [this PR](https://github.com/y-scope/spider/pull/397)

tools/deployment/spider-helm/values.yaml
20001020ycx: done

tools/deployment/spider-helm/templates/database-secret.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: {{ include "spider.componentFullname" (dict "root" . "component" "database") }}
+  labels:
+    {{- include "spider.labels" . | nindent 4 }}
+    app.kubernetes.io/component: "database"
+type: "Opaque"
+stringData:
+  username: {{ .Values.spiderConfig.database.username | quote }}
+  password: {{ .Values.spiderConfig.database.password | quote }}
+  root_password: {{ .Values.spiderConfig.database.root_password | quote }}
```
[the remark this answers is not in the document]
20001020ycx: outdated, please check

tools/deployment/spider-helm/templates/database-statefulset.yaml
```diff
@@ -0,0 +1,71 @@
...
+    metadata:
+      labels:
+        {{- include "spider.labels" . | nindent 8 }}
+        app.kubernetes.io/component: "database"
+    spec:
+      containers:
+        - name: "database"
+          image: {{ include "spider.imageRef" (dict "root" . "component" "database") | quote }}
+          imagePullPolicy: {{ .Values.image.database.pullPolicy | quote }}
+          env:
```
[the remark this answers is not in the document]
20001020ycx: Deferring to #394, which tracks CLP-style opt-in `resources`/`scheduling` overrides across the chart workloads (now including the bundled database). We intentionally will not hardcode default requests/limits here—CLP renders these blocks only when the user sets them, and ships empty defaults.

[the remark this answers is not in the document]
20001020ycx: tracked in https://github.com/y-scope/spider/issues/394, let's defer this

tools/deployment/spider-helm/templates/database-statefulset.yaml
```diff
@@ -0,0 +1,71 @@
...
+          readinessProbe:
+            {{- include "spider.readinessProbeTimings" . | nindent 12 }}
+            exec: &health-check
+              command: [
+                "mysqladmin", "ping",
+                "--host=127.0.0.1",
+                "--port=3306",
+                "--user=$(MYSQL_USER)",
+                "--password=$(MYSQL_PASSWORD)"
+              ]
```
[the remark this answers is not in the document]
20001020ycx: if a user can do ps, isnt that he is already in the data center/nodes. Leaking the username and pwd is probably least of the concern in this case.

The reason why we hide secrets from configmap is that user can download it using kubctrl, but this requires remote login to the machine, not related to our security concern in this PR

Also, CLP does the same thing.

tools/deployment/spider-helm/values.yaml
```diff
@@ -1 +1,47 @@
...
+    name: "spider-db"
+
+  storage_endpoint:
+    port: 50051
+
+  storage:
+    max_connections: 64
```
junhaoliao: shall we move this under `spiderConfig.database`? that way we don't have to do the `omit` in the configmap. the mapping will be more natural as well

tools/deployment/spider-helm/templates/database-statefulset.yaml
```diff
@@ -0,0 +1,71 @@
...
+            {{- include "spider.livenessProbeTimings" . | nindent 12 }}
+            exec: *health-check
+  volumeClaimTemplates:
+    - metadata:
+        name: "data"
+      spec:
+        accessModes: ["ReadWriteOnce"]
+        resources:
+          requests:
+            storage: "20Gi"
```
hoophalab, line 70: We plan to template this 20Gi later?

tools/deployment/spider-helm/templates/storage-deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+            - name: "grpc"
+              containerPort: {{ .Values.spiderConfig.storage_endpoint.port }}
+          volumeMounts:
+            - name: "config"
+              mountPath: "/etc/spider"
+              readOnly: true
+          readinessProbe:
+            {{- include "spider.readinessProbeTimings" . | nindent 12 }}
+            tcpSocket:
+              port: "grpc"
```
hoophalab, line 47: It appears Spider's gRPC server doesn't enable the health check API. We might need to consider enabling it later, since TCP probes only confirm something is listening on the port and don't verify the service process actually responds correctly.

tools/deployment/spider-helm/templates/storage-deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+                secretKeyRef:
+                  name: {{ include "spider.componentFullname" (dict "root" . "component" "database") }}
+                  key: "password"
+          ports:
+            - name: "grpc"
+              containerPort: {{ .Values.spiderConfig.storage_endpoint.port }}
+          volumeMounts:
+            - name: "config"
+              mountPath: "/etc/spider"
+              readOnly: true
```
hoophalab, line 43: Without `subPath` here (or `items` in `volumes:`), every file in configmap is mounted to this path.
Just to confirm: this is intended since all config to be added in configmap are to be shared by all spider's components?

tools/deployment/spider-helm/values.yaml
junhaoliao: also have we consider this structure?

```
nameOverride: ""
fullnameOverride: ""

image: {}

bundled: []

databse:
   host: ...
   max_connections: ...

storage:
   port: {} # or endpoint_port ?
   task_instance_pool_config:
       execution_manager_stale_cutoff_sec: 30
       gc_interval_sec: 5
       message_channel_capacity: 1024
...
```

cc @sitaowang1998 @LinZhihao-723 please confirm if the above structure makes sense for a consolidated config file

in clp, all clp package related configs were nested under clpConfig for compatibility with the CLP Docker Compose package's clp-config.yaml (as in, users can override the entire clpConfig with the contents of a working docker compose package's clp-config.yaml without many modifications)

here in spider, since we have previously never maintained a centralized config file, i think it's ok to just fields at the root level whenever it makes sense to

hoophalab (dismissed):
Overall looks good. Some minor confirmations/comments.

Verified that all pods started correctly.

tools/deployment/spider-helm/values.yaml
junhaoliao: i'm not sure if i understand the questions and explanations here, but here's the deal:

1. if we want certain CLP configs to be configurable in the CLP Package Helm chart, those fields should be listed and uncommented in the CLP Package Helm chart's `values.yaml`.
2. if we want certain spider configs to be configurable in the Spider Helm chart, those fields should be listed and uncommented in the Spider chart's `values.yaml`.
3. if certain CLP configs need to be mapped into the Spider Helm chart, which will be integrated as a subchart into the CLP Helm chart. we need not only 2 but also 1
4. I understand the whole `storage` object, without caring what fields are inside, is turned into YAML and embedded into the config files. however, the current config interfaces in values.yaml is already a bit awkward and might need some more work (see https://github.com/y-scope/spider/pull/384#discussion_r3597422655 )
5. if the values listed here are consistent with the default values used in the code, i think we should leave them uncommented. (unless having them specified will interfere the normal code execution?) as in, we shouldn't comment out the configuration interfaces

junhaoliao (changes requested):
the rest lgtm

tools/deployment/spider-helm/values.yaml
LinZhihao-723: I'm ok to move it under DB config.
However, we need to notice that this max connection is for Rust client side, not the actual DB side config.

tools/deployment/spider-helm/templates/database-statefulset.yaml
20001020ycx: All resources related features are tracked in https://github.com/y-scope/spider/issues/394, btw, CLP also seems to be using the same parameter. 

But thanks for the catch!

tools/deployment/spider-helm/templates/storage-deployment.yaml
20001020ycx: Right, forgot to tracked this when you mentioned to me offline, @coderabbitai, can you track this issue for me, reviewer did a good catch.

tools/deployment/spider-helm/values.yaml
20001020ycx: As discussed offline, `max_connections: 64` will be moved to spiderConfig.database. `spider.yaml` provided by spider will also follow the similar change

tools/deployment/spider-helm/templates/storage-deployment.yaml
20001020ycx: done

tools/deployment/spider-helm/values.yaml
20001020ycx: port is also moved as well into the storage struct.

tools/deployment/spider-helm/values.yaml
20001020ycx: As discussed offline, all commented config are deleted, user shall consult with Spider's doc when configuration optional parameter

junhaoliao (approved):
lgtm

LinZhihao-723 (approved):

PR #385, opened 2026-07-10
test(e2e): Add end-to-end tests simulating neural-network workloads on a Spider cluster.

sitaowang1998:
# Description

This PR adds neural-network integration test, with the following change:
- `tests/huntsman/e2e/src/nn`: New mod contains neural network helpers.
  - `neuron.rs`: `Neuron` activation enums and mapping to task name and evaluate function.
  - `wiring.rs`: Wiring outputs to inputs.
  - `network.rs`: `NeuralNetwork` struct that holds all neural network information and conversion to `TaskGraph`.
- `tests/huntsman/e2e/payload_serde.rs`: Helpers that serialize/deserialize task `msgpack` input/output payload.
- `taskfiles/test.yaml`: Excludes `e2e` tests from `test:rust-unit-tests` task.

# Checklist

  breaking change.

# Validation performed

- [x] Run neural network test on a Spider stack. The test codes are in [another branch](https://github.com/sitaowang1998/spider/tree/integration-test-stack).
- [x] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Added an end-to-end neural-network workflow with layered neuron graphs, deterministic wiring, and ReLU, sigmoid, and identity activations.
  * Added MessagePack payload encoding and decoding for task inputs and outputs.
  * Added validation for neural-network layer wiring and test payload handling.
* **Tests**
  * Added an asynchronous end-to-end test that submits generated networks and verifies results against in-process simulation.
* **Bug Fixes**
  * Updated standard test execution to exclude end-to-end tests, preventing duplicate runs.


tests/huntsman/e2e/src/nn/wiring.rs
```diff
@@ -0,0 +1,154 @@
...
+/// * The layer list is non-empty.
+/// * Each layer's size is at least the neuron fan-in [`NUM_INPUTS`].
+/// * Each consecutive pair of layers fully covers the previous layer's outputs (`next_size *
+///   NUM_INPUTS >= prev_size`).
+///
+/// # Errors
+///
+/// Returns an error if:
+///
+/// * [`anyhow::Error`] if an invariant is voilated.
```
LinZhihao-723:
```suggestion
/// * [`anyhow::Error`] if an invariant is violated.
```

tests/huntsman/e2e/src/nn/wiring.rs
```diff
@@ -0,0 +1,154 @@
...
+/// Builds the per-neuron fan-in wiring for every layer.
+///
+/// # Returns
+///
+/// The per-neuron fan-in wiring, indexed `[layer][neuron][fan_in]`. `wiring[0]` is empty since
+/// layer 0 reads graph inputs directly.
+///
+/// # Panics
+///
+/// Panics if the layer invariants do not hold.
```
LinZhihao-723:
```suggestion
```
This function doesn't panic directly. Don't think we need to explicitly document it.

tests/huntsman/e2e/src/nn/wiring.rs
```diff
@@ -0,0 +1,154 @@
...
+    wiring
+}
+
+/// Deals numbers from a shuffled deck, reshuffling a fresh permutation whenever the current deck
+/// runs out.
+struct Dealer<'a> {
+    rng: &'a mut StdRng,
+    range_size: usize,
+    deck: Vec<usize>,
+}
```
LinZhihao-723: - `'a` is not a valid lifetime parameter name even as a short-term one according to the coding guideline: https://github.com/LinZhihao-723/claude-instruction/blob/main/rust/coding-style.md#generic-type-parameter-names
- I feel like we can use the following so that it doesn't need to store a rng.

```rust
  struct Dealer {                                                                                      
      range_size: usize,                                                                               
      deck: Vec<usize>,                                                                                
  }                                                                                                    
                                                                                                       
  impl Dealer {                                                                                        
      fn new(range_size: usize) -> Self {                                                              
          Self { range_size, deck: Vec::new() }                                                        
      }                                                                                                
                                                                                                       
      fn draw(&mut self, rng: &mut StdRng) -> usize {                                                  
          if self.deck.is_empty() {                                                                    
              self.deck = shuffled_range(rng, self.range_size);                                        
          }                                                                                            
          self.deck.pop().expect("reshuffled deck must be non-empty")                                  
      }                                                                                                
  }
```

tests/huntsman/e2e/src/nn/network.rs
```diff
@@ -0,0 +1,179 @@
...
+                    let neuron_inputs: [f64; NUM_INPUTS] =
+                        std::array::from_fn(|j| layer_outputs[layer.fan_in[i][j]]);
+                    layer.activation.evaluate_func()(&neuron_inputs)
+                })
+                .collect();
+        }
+
+        Ok(layer_outputs)
+    }
+}
```
LinZhihao-723, line 165: Nit: should rename `first` to `first_layer` for clarity.

tests/huntsman/e2e/src/payload_serde.rs
```diff
@@ -0,0 +1,154 @@
...
+            );
+        }
+    }
+
+    #[derive(Debug, PartialEq, Serialize, Deserialize)]
+    struct Sample {
+        flag: bool,
+        count: i64,
+        label: String,
+    }
```
LinZhihao-723: Group with other type definitions.

tests/huntsman/e2e/src/payload_serde.rs
```diff
@@ -0,0 +1,154 @@
...
+/// # Errors
+///
+/// Returns an error if:
+///
+/// * Forwards [`rmp_serde::to_vec`]'s return values on failure.
+pub fn encode_input<T>(value: &T) -> anyhow::Result<TaskInput>
+where
+    T: Serialize, {
+    Ok(TaskInput::ValuePayload(rmp_serde::to_vec(value)?))
+}
```
LinZhihao-723: We should prefer not to use `where` if possible:
```suggestion
pub fn encode_input<T: Serialize>(value: &T) -> anyhow::Result<TaskInput> {
    Ok(TaskInput::ValuePayload(rmp_serde::to_vec(value)?))
}
```
Do it for the rest.

tests/huntsman/e2e/tests/nn.rs
```diff
@@ -0,0 +1,99 @@
...
+const NUM_LAYERS: usize = 10;
+
+/// Neurons per layer in the test network.
+const LAYER_SIZE: usize = 1000;
+
+#[tokio::test]
+async fn test_nn() -> anyhow::Result<()> {
+    if std::env::var("SPIDER_ENDPOINT").is_err() {
+        bail!("SPIDER_ENDPOINT is not set");
+    }
```
LinZhihao-723: I don't think we should assert this endpoint env var here. It should be abstracted by the test driver.

tests/huntsman/e2e/tests/nn.rs
```diff
@@ -0,0 +1,99 @@
...
+const REL_TOL: f64 = 1.0e-12;
+
+/// Number of layers in the test network.
+const NUM_LAYERS: usize = 10;
+
+/// Neurons per layer in the test network.
+const LAYER_SIZE: usize = 1000;
+
+#[tokio::test]
+async fn test_nn() -> anyhow::Result<()> {
```
LinZhihao-723, line 21: We should enrich this test case: it only submits one job iiuc, which runs once. We should probably submit X jobs together in a batch (in parallel with a tokio join), and in total Y batches. For now, we should only assert the results, but in a future PR we should also assert that, for every batch, the e2e time across a batch doesn't vary too much, ensuring fairness.

taskfiles/test.yaml
```diff
@@ -248,7 +248,8 @@ tasks:
            "{{.G_TDL_PACKAGES_DIR}}/complex/libcomplex.so"
         cp "{{.G_RUST_RELEASE_DIR}}/libintegration_test_tasks.so" \
            "{{.G_TDL_PACKAGES_DIR}}/integration_test_tasks/libintegration_test_tasks.so"
-        cargo nextest run --all --all-features --run-ignored all --release
+        cargo nextest run --all --all-features --run-ignored all --release \
```
LinZhihao-723: I will do some research on whether we can make this better to understand.

LinZhihao-723 (changes requested):
Overall I'm good with the shape. Shouldn't take too much effort to merge it.

tests/huntsman/e2e/src/payload_serde.rs
sitaowang1998: This is a local, unit test only type definition. There is no other type definition to group to. I move it to the end of the tests to follow the style guide.

LinZhihao-723: I think the convention is to order non-test-case definitions before test cases, right? They share the same visibility.

tests/huntsman/e2e/src/payload_serde.rs
```diff
@@ -0,0 +1,152 @@
...
+/// The decoded `T` on success.
+///
+/// # Errors
+///
+/// Returns an error if:
+///
+/// * Forwards [`rmp_serde::from_slice`]'s return values on failure.
+pub fn decode_output<T>(output: &TaskOutput) -> anyhow::Result<T>
+where
+    T: DeserializeOwned, {
```
LinZhihao-723: Make sure you read the comment before you merge it: https://github.com/y-scope/spider/pull/385#discussion_r3825663157

tests/huntsman/e2e/src/payload_serde.rs
```diff
@@ -0,0 +1,152 @@
...
+    struct Sample {
+        flag: bool,
+        count: i64,
+        label: String,
+    }
+
+    /// Round-trips `value` through [`encode_input`] then [`decode_output`].
+    fn round_trip<T>(value: &T) -> T
+    where
+        T: Serialize + DeserializeOwned, {
```
LinZhihao-723: Same.

LinZhihao-723 (changes requested):

tests/huntsman/e2e/tests/nn.rs
```diff
@@ -0,0 +1,136 @@
...
+const LAYER_SIZE: usize = 1000;
+
+/// Maximum duration of one neural-network job.
+const JOB_TIMEOUT: Duration = Duration::from_secs(300);
+
+/// Number of neural-network job batches.
+const NUM_BATCHES: usize = 3;
+
+/// Number of concurrent neural-network jobs in each batch.
+const NUM_JOBS_PER_BATCH: usize = 8;
```
LinZhihao-723: Not all of these constants are used globally, right? We should norrow the scope by moving them into the function body if it's only used once in that function.

tests/huntsman/e2e/Cargo.toml
```diff
@@ -4,8 +4,16 @@ version = { workspace = true }
 edition = { workspace = true }
 publish = false
 
+[[test]]
+name = "nn"
+test = false
```
LinZhihao-723, line 9: Ok, this will disable the test by default. There's no need to explicitly disable it in the current task.
To run the nn test case:
```bash
cargo nextest run --package e2e --all-features --release --test nn
```

LinZhihao-723 (changes requested):

LinZhihao-723 (approved):
For the PR title, how about:
```
test(e2e): Add end-to-end tests simulating neural-network workloads on a Spider cluster.
```

PR #390, opened 2026-07-12
feat(huntsman): Add support for accessing task graph outputs in a commit task.

LinZhihao-723:
# Description

This PR lets a commit task read the outputs of its job's task graph from its `TaskContext`, mirroring the job-level output-retrieval APIs but delivering the data inline to the commit task instead of to an external client. The execution context that storage ships to the execution manager already carries a serialized payload per task instance; this PR generalizes that payload so it carries task *inputs* for regular tasks (unchanged behavior) and the task-graph *outputs* for a commit task, then routes it into the commit task's `TaskContext` where a new accessor deserializes it on demand. The wire field is renamed once to reflect the dual role, and the rename preserves the protobuf tag so it stays wire-compatible.

## Wire payload: one field for inputs and outputs (`spider-core`, `spider-proto`)

- Renames `ExecutionContext::serialized_inputs` to `serialized_task_io` in the core type (`spider-core/src/types/io.rs`) and in the protobuf message (`spider-proto/storage/storage.proto`, regenerated `spider-proto-rust/src/generated/storage.rs`, and the `TryFrom` conversion in `spider-proto-rust/src/io.rs`). The protobuf field keeps tag `4`, so the change is wire-compatible.
- The field now means "task inputs for a regular task, task-graph outputs for a commit task"; the producer (storage) decides which based on the task kind.

## Storage: build the commit task's execution context from task-graph outputs (`spider-storage`)

- Adds `TaskGraph::read_output_payloads()` (`cache/task.rs`), which reads the payloads of every task-graph output, and collapses the three previously-duplicated read loops (`get_outputs`, the `succeed_task_instance` completion path, and `create_commit_task_instance`) onto it.
- `create_commit_task_instance` (`cache/job.rs`) now serializes the task-graph outputs via `SerializedTaskOutputs::serialize_with_size_hint` and stores the raw bytes in the execution context's `serialized_task_io`. Regular-task registration continues to place the task inputs there, and the cleanup path leaves it empty.
- Adds `InternalError::TaskOutputs(#[from] spider_core::types::io::TaskOutputsError)` (`cache/error.rs`) so the output serialization error propagates through the cache error type.

## Execution manager: route the payload into the `TaskContext` (`spider-execution-manager`)

- `build_request` (`process_pool.rs`) inspects the task kind: for a commit task it moves `serialized_task_io` into the `TaskContext`'s task-graph outputs and leaves the executor's `raw_inputs` empty; for any other task it passes `serialized_task_io` through as `raw_inputs` and leaves the context without task-graph outputs.
- Adds `InternalError::BuildTaskContext(#[from] spider_tdl::TdlError)` so a failed `TaskContext` construction surfaces as a pool error.

## Executor API: `TaskContext` (`spider-tdl`)

- `TaskContext` gains a private `serialized_task_graph_outputs: Option<Vec<u8>>` field holding the raw serialized outputs (kept raw so they pass through the execution manager to the executor without an intermediate deserialize/re-encode).
- `TaskContext::new` is now fallible (`Result<Self, TdlError>`) and validates the invariant that task-graph outputs are present for a commit task and absent otherwise, returning `TdlError::InvalidTaskContext` on a violation.
- `get_task_graph_outputs(&self) -> Result<Option<Vec<TaskOutput>>, TdlError>` deserializes the raw bytes on demand: `Ok(Some(..))` for a commit task, `Ok(None)` for a non-commit task, and a deserialization error if the stored bytes are corrupt.
- Adds the `TdlError::InvalidTaskContext(String)` variant (`error.rs`). Call sites that construct a `TaskContext` (the `#[task]` handler test helpers and integration harnesses) are updated for the fallible constructor.

## Tests

- **Storage** (`spider-storage/tests/scheduling_infra.rs`): the mock execution manager's commit handler now decodes the commit context's `serialized_task_io` as `SerializedTaskOutputs` and asserts the decode succeeds, so every commit-bearing workload (`test_flat_success` and its variants) exercises the new payload end-to-end.
- **Execution manager** (`spider-execution-manager/src/process_pool.rs`): unit tests for `build_request` covering the commit routing (populated and empty outputs) and the regular-task pass-through.
- **Executor unit** (`spider-tdl/src/task_context.rs`, `error.rs`): construction-invariant tests (`new` rejects outputs on a non-commit task and missing outputs on a commit task), accessor tests (populated, empty, non-commit `None`, corrupt bytes, msgpack round-trip), and the error round-trip covers the new variant.
- **Executor end-to-end** (`integration-test-tasks`, `test-utils`, `task-executor` tests): adds a commit task `assert_outputs_sum_zero` that reads its task-graph outputs and asserts the `i64` values sum to zero; adds `build_commit_ctx` / `commit_execute_request` harness helpers to construct a commit context carrying serialized outputs; and adds four `test_executor.rs` cases (sums to zero, doesn't sum to zero, empty outputs, and a non-commit context) exercised against a real `spider-task-executor` subprocess.

## Notes

- The raw serialized outputs are carried verbatim from storage through the execution manager into the `TaskContext` and are only deserialized when the commit task calls `get_task_graph_outputs`, so the execution manager never pays for a deserialize/re-encode of a payload it only forwards.
- The commit-vs-outputs invariant is enforced once, at `TaskContext::new`; `get_task_graph_outputs` trusts it and returns `None` purely from the absence of stored bytes rather than re-checking the task id.

# Checklist

  breaking change.

# Validation performed

* [x] Ensure new test cases all pass.

## Summary by CodeRabbit

* **New Features**
  * Commit tasks can now receive and process serialized task-graph outputs.
  * Added validation for task contexts to ensure outputs are supplied only for commit tasks.
  * Added support for reading stored task outputs and reporting unavailable outputs clearly.
  * Added a task for verifying that committed output values sum to zero.

* **Bug Fixes**
  * Improved routing of task inputs and outputs during task execution.
  * Added clearer errors for invalid task contexts and output-processing failures.
  * Updated execution-context payload naming from `serialized_inputs` to `serialized_task_io`.


sitaowang1998:
Before I go into details, I have a question: should we reuse `TaskContext` and make it hold different things for different kinds of tasks or should we just create another struct for it?

LinZhihao-723:
> Before I go into details, I have a question: should we reuse `TaskContext` and make it hold different things for different kinds of tasks or should we just create another struct for it?

I think we should use a single type for it, even the underlying data it holds differ depending on the task types. This makes the user-side more friendly since you don't need to decide what the type of the first parameter for a task function is.

components/spider-tdl/src/task_context.rs
```diff
@@ -22,27 +26,215 @@ pub struct TaskContext {
...
+    /// # Returns
+    ///
+    /// The job's task-graph outputs (if any) or `None` on success.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * Forwards [`SerializedTaskOutputs::deserialize_from_raw`]'s return values on failure.
+    pub fn get_task_graph_outputs(&self) -> Result<Option<Vec<TaskOutput>>, TdlError> {
```
sitaowang1998, line 90: Maybe we should return an error instead of `None` if called by non-commit task?

LinZhihao-723: That's actually my initial implementation in https://github.com/y-scope/spider/pull/390/commits/f6b22279304c4563a0a2f78d60a8c00145eb94d4. But I think from a general perspective, `get_task_graph_outputs` is more like a method to query the task graph outputs; it doesn't necessarily need to fail if the outputs are not ready. This should give us more flexibility to reuse this method, for example, in cleanup tasks.

sitaowang1998 (dismissed):

sitaowang1998 (approved):

PR #392, opened 2026-07-13
fix(huntsman): Use `tokio::net::TcpListener` to enable `SO_REUSEADDR` on gRPC servers.

sitaowang1998:
# Description

## Background
Tonic uses [`std::net::TcpListener`](https://doc.rust-lang.org/std/net/struct.TcpListener.html) by default, which does not set [`SO_REUSEADDR`](https://man7.org/linux/man-pages/man7/socket.7.html) when opening the socket.

TCP has a [complicated state transition](https://en.wikipedia.org/wiki/File:Tcp_state_diagram_fixed.svg) on close down. After a server actively closes its socket listening on a port, the port will enter a `TIME_WAIT` state and timeout by default 60 seconds for the port to go back to `CLOSED` state. This is to ensure that stray/duplicate packet still on transmission will timeout on arrival and won't be delivered to newly connected listener.

However, this 60-second wait is too long for a fast crash-recovered server. It tries to bind to a port in `TIME_WAIT`, and fails if `SO_REUSEADDR` is not set. It crashes again, and docker compose will restart it after exponential backoff. Thus, it is likely that the crashed server will only become healthy after 60 second `TIME_WAIT` timeout + 40 second docker compose backoff, which is way too long for critical services like storage server and scheduler.

## Fix
This PR fixes the problem by using the `tokio::net::TcpListener` instead of `std::net::TcpListener`, the default TCP listener used by tonic. `std::net::TcpListener` does not set `SO_REUSEADDR`, while the `tokio::net::TcpListener` has it on.

This PR:
- In two `grpc_server.rs` in scheduler and storage, creates a `tokio::net::TcpListener` and pass it to tonic.
- In two `Cargo.toml` in aforementioned components, adds `net` feature for tokio.

# Checklist

  breaking change.

# Validation performed

- [x] The e2e failure recovery test succeeds on [another branch](https://github.com/sitaowang1998/spider/tree/integration-test-dev-compose), with an unrelated fix that will be put up in another PR.
- [ ] GitHub workflow pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gRPC server startup by explicitly binding to the configured network address.
  * Added clearer error logging when a server cannot bind to its listening address.
  * Updated shutdown handling to work reliably with incoming network connections.


LinZhihao-723 (approved):
Directly modified the PR title. Notice that the scope should not be `spider-grpc-server` since we don't have any crates or namespaces of that name.

PR #401, opened 2026-07-16
feat(huntsman)!: Replace endpoint's IP address with general hostname.

### LINKED ISSUE #400: Allow scheduler config to accept hostnames instead of raw IP addresses
## Summary
The scheduler's config addresses are typed as `IpAddr`, so they can only hold a numeric IP, never a DNS name:

- `storage_endpoint.host` — `spider_utils::config::EndpointConfig.host: IpAddr` — the storage endpoint the scheduler connects to.
- `runtime.host` — `spider-scheduler` `RuntimeConfig.host: IpAddr` — the address the scheduler advertises to storage during registration (and also its gRPC bind host).

In Kubernetes, services are reached by stable **Service DNS names** (pod and ClusterIP addresses are dynamic). Because these fields reject hostnames, the Helm chart must pre-resolve names to IPs at pod startup — an initContainer running `getent` + `sed` over a config template + a scratch volume. Accepting a hostname would let the chart pass Service names directly and let DNS resolve at connect time, removing that machinery.

## Proposed change
1. **`EndpointConfig.host` (connect side, e.g. storage endpoint):** accept a hostname/URL (e.g. `spider-storage:50051`). gRPC/tonic builds channels from a URI, so DNS resolves at connect time — no pre-resolution needed.
2. **`RuntimeConfig.host` (scheduler self):** this field does double duty — the gRPC **bind** host *and* the address **advertised** to storage. Separate the two:
   - bind stays an IP (`0.0.0.0`),
   - advertised host accepts a hostname (so it can be a Service name or stable per-pod name).

## Acceptance criteria
- Scheduler config accepts a hostname for the storage endpoint; resolution happens at connect time.
- Scheduler bind vs advertised address are separable, so `0.0.0.0` bind + hostname/IP advertise both work.
- The Helm chart can drop the scheduler initContainer / `getent` / `sed` config rendering.

## Context
- Surfaced by the Helm chart work in #384; related chart follow-up: #394.
- `EndpointConfig` is a shared type in `spider-utils`, so the change may affect other components that use it (e.g. storage's own endpoint config).

sitaowang1998:
# Description

> [!Note]
> This PR is a breaking change because it changes configuration.

## Background

Spider's storage and scheduler use the same IP address for them to listen on and for other components to connect to them. However, for deployment in docker compose and helm, the network are split and each service's network typically expose a hostname for other service to connect to, which is different from the IP address inside the network where each server bind on.

## Solution
This PR resolves #400 by separating IP address to bind on with the connecting hostname.
- Adds a shared `Host` alias backed by `NonEmptyString`, representing a DNS name or an IP address.
- Updates `EndpointConfig` to accept the `Host`, so all components use `Host` to connect to storage and scheduler.
- Separates the scheduler's bind address from its advertised endpoint:
  - `ServerConfig.host` and `port` remain the socket bind address.
  - `RuntimeConfig.advertised_endpoint` identifies the scheduler endpoint registered with storage.
- Propagates scheduler hostnames through the registration protobuf, core types, storage, and database.

## Scheduler Configuration
This PR changes the shape of the scheduler's configuration. Below is an example of the configuration after change. `host` can be a DNS name, an IPv4 address of an IPv6 address surrounded by bracket (e.g. `[::1]`).
```
host: "0.0.0.0"
port: 50052

storage_endpoint:
  host: "spider-storage"
  port: 50051

connection_pool_size: 4

runtime:
  advertised_endpoint:
    host: "spider-scheduler"
    port: 50052

  scheduler: !round_robin
    active_job_queue_capacity: 64
    cleanup_ready_task_capacity: 256
    commit_ready_task_capacity: 256
    dispatch_queue_capacity: 64
    finalizing_job_expiration_timeout_sec: 300
    ready_task_capacity: 65536
    storage_poll_timeout_ms: 10
    tick_interval_ms: 5

  stop_timeout_sec: 30

  em_registry:
    dead_em_cutoff_sec: 60
    liveness_tracking_interval_ms: 1000
```

# Checklist

  breaking change.

# Validation performed

- [ ] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Scheduler endpoints now support validated hostnames as well as IP addresses.
  * Scheduler configuration separates the advertised endpoint from the server’s listening address.
  * Scheduler registration and storage now preserve host and port information consistently.

* **Bug Fixes**
  * Improved validation prevents empty scheduler hosts from being registered.
  * Scheduler replacement and persistence now work with DNS hostnames.

### DIFF
--- a/components/spider-proto-rust/src/unpack/storage.rs
+++ b/components/spider-proto-rust/src/unpack/storage.rs
@@ -9,6 +9,7 @@ use spider_core::types::id::ResourceGroupId;
 use spider_core::types::id::SessionId;
 use spider_core::types::id::TaskId;
 use spider_core::types::id::TaskInstanceId;
+use spider_utils::config::Host;
 use tonic::Code;
 
 use crate::storage::AddResourceGroupRequest;
@@ -185,21 +186,16 @@ impl RequestUnpack for ExecutionManagerIdRequest {
     }
 }
 
-/// Unpacks [`RegisterSchedulerRequest`] into a tuple containing:
-///
-/// * The scheduler IP address.
-/// * The scheduler port.
+/// Unpacks [`RegisterSchedulerRequest`] into the scheduler host and port.
 impl RequestUnpack for RegisterSchedulerRequest {
-    type Unpacked = (IpAddr, u16);
+    type Unpacked = (Host, u16);
 
     fn unpack(self) -> Result<Self::Unpacked, UnpackError> {
-        let ip_address = self
-            .ip_address
-            .parse::<IpAddr>()
-            .map_err(|error| invalid_argument(format!("invalid IP address: {error}")))?;
+        let host = Host::new(self.host)
+            .map_err(|_| invalid_argument("host must not be empty".to_owned()))?;
         let port = u16::try_from(self.port)
             .map_err(|_| invalid_argument(format!("port does not fit in `u16`: {}", self.port)))?;
-        Ok((ip_address, port))
+        Ok((host, port))
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2010,6 +2010,7 @@ version = "0.1.0"
 dependencies = [
  "prost",
  "spider-core",
+ "spider-utils",
  "thiserror",
  "tonic",
  "tonic-prost",
@@ -2114,6 +2115,7 @@ dependencies = [
 name = "spider-utils"
 version = "0.1.0"
 dependencies = [
+ "non-empty-string",
  "rand 0.9.4",
  "rmp-serde",
  "serde",
--- a/components/spider-core/src/types/scheduler.rs
+++ b/components/spider-core/src/types/scheduler.rs
@@ -1,4 +1,4 @@
-use std::net::IpAddr;
+use spider_utils::config::Host;
 
 use crate::types::id::JobId;
 use crate::types::id::ResourceGroupId;
@@ -8,10 +8,10 @@ use crate::types::id::TaskAssignmentId;
 use crate::types::id::TaskId;
 
 /// The currently registered scheduler endpoint.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct RegisteredScheduler {
     pub id: SchedulerId,
-    pub ip_address: IpAddr,
+    pub host: Host,
     pub port: u16,
 }
 
--- a/components/spider-proto-rust/Cargo.toml
+++ b/components/spider-proto-rust/Cargo.toml
@@ -10,6 +10,7 @@ path = "src/lib.rs"
 [dependencies]
 prost = "0.14.4"
 spider-core = { path = "../spider-core" }
+spider-utils = { path = "../spider-utils" }
 thiserror = "2.0.18"
 tonic = "0.14.6"
 tonic-prost = "0.14.6"
--- a/components/spider-proto-rust/src/generated/storage.rs
+++ b/components/spider-proto-rust/src/generated/storage.rs
@@ -181,7 +181,7 @@ pub struct UpdateExecutionManagerHeartbeatResponse {
 #[derive(Clone, PartialEq, Eq, Hash, ::prost::Message)]
 pub struct RegisterSchedulerRequest {
     #[prost(string, tag = "1")]
-    pub ip_address: ::prost::alloc::string::String,
+    pub host: ::prost::alloc::string::String,
     #[prost(uint32, tag = "2")]
     pub port: u32,
 }
@@ -202,7 +202,7 @@ pub struct Scheduler {
     #[prost(uint64, tag = "1")]
     pub scheduler_id: u64,
     #[prost(string, tag = "2")]
-    pub ip_address: ::prost::alloc::string::String,
+    pub host: ::prost::alloc::string::String,
     #[prost(uint32, tag = "3")]
     pub port: u32,
 }
--- a/components/spider-proto-rust/src/scheduler_registration.rs
+++ b/components/spider-proto-rust/src/scheduler_registration.rs
@@ -8,7 +8,7 @@ impl From<RegisteredScheduler> for storage::Scheduler {
     fn from(scheduler: RegisteredScheduler) -> Self {
         Self {
             scheduler_id: scheduler.id.get(),
-            ip_address: scheduler.ip_address.to_string(),
+            host: scheduler.host.into_inner(),
             port: u32::from(scheduler.port),
         }
     }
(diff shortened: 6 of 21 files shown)


20001020ycx (commented):
Before diving into the code details, I wonder what is the new configuration looks like (do we need http attached to url? is this backward compatible with ipv4 or ipv6)? Can you show how this breaking change would affect user's configuration e2e?

sitaowang1998:
> Before diving into the code details, I wonder what is the new configuration looks like (do we need http attached to url? is this backward compatible with ipv4 or ipv6)? Can you show how this breaking change would affect user's configuration e2e?

Below is a configuration example. All `host` can be either a hostname, an IPv4 or an IPv6 with bracket around it (e.g. `[::1]`). The top level port and the `advertised_endpoint` port don't need to match, depend on how helm routes the traffic.
```
host: "0.0.0.0"
port: 50052

storage_endpoint:
  host: "spider-storage"
  port: 50051

connection_pool_size: 4

runtime:
  advertised_endpoint:
    host: "spider-scheduler"
    port: 50052

  scheduler: !round_robin
    active_job_queue_capacity: 64
    cleanup_ready_task_capacity: 256
    commit_ready_task_capacity: 256
    dispatch_queue_capacity: 64
    finalizing_job_expiration_timeout_sec: 300
    ready_task_capacity: 65536
    storage_poll_timeout_ms: 10
    tick_interval_ms: 5

  stop_timeout_sec: 30

  em_registry:
    dead_em_cutoff_sec: 60
    liveness_tracking_interval_ms: 1000
```

20001020ycx:
> > Before diving into the code details, I wonder what is the new configuration looks like (do we need http attached to url? is this backward compatible with ipv4 or ipv6)? Can you show how this breaking change would affect user's configuration e2e?
> 
> Below is a configuration example. All `host` can be either a hostname, an IPv4 or an IPv6 with bracket around it (e.g. `[::1]`). The top level port and the `advertised_endpoint` port don't need to match, depend on how helm routes the traffic.
> 
> ```
> host: "0.0.0.0"
> port: 50052
> 
> storage_endpoint:
>   host: "spider-storage"
>   port: 50051
> 
> connection_pool_size: 4
> 
> runtime:
>   advertised_endpoint:
>     host: "spider-scheduler"
>     port: 50052
> 
>   scheduler: !round_robin
>     active_job_queue_capacity: 64
>     cleanup_ready_task_capacity: 256
>     commit_ready_task_capacity: 256
>     dispatch_queue_capacity: 64
>     finalizing_job_expiration_timeout_sec: 300
>     ready_task_capacity: 65536
>     storage_poll_timeout_ms: 10
>     tick_interval_ms: 5
> 
>   stop_timeout_sec: 30
> 
>   em_registry:
>     dead_em_cutoff_sec: 60
>     liveness_tracking_interval_ms: 1000
> ```

Thanks for the explanation, can we also include this in the PR so the reference is more clear (unless this violate any policies you guys have in Spider).

Also a few questions:
1. Why do we need host in the configuration? Seems conflicting with the config `advertised_endpoint`
2. Why quotation mark around "spider-scheduler"? I assume `spider-scheduler` is the URL name, and we are omitting the `http://` right? @LinZhihao-723 are you okay with this?

components/spider-proto-rust/src/unpack/storage.rs
```diff
@@ -185,21 +186,16 @@ impl RequestUnpack for ExecutionManagerIdRequest {
...
 impl RequestUnpack for RegisterSchedulerRequest {
-    type Unpacked = (IpAddr, u16);
+    type Unpacked = (Host, u16);
 
     fn unpack(self) -> Result<Self::Unpacked, UnpackError> {
-        let ip_address = self
-            .ip_address
-            .parse::<IpAddr>()
-            .map_err(|error| invalid_argument(format!("invalid IP address: {error}")))?;
+        let host = Host::new(self.host)
```
20001020ycx, line 194: What if DNS failed to resolve the URL. This failure is often seen in the networking, and we should be able to deal with the gracefully

components/spider-proto-rust/src/unpack/storage.rs
```diff
@@ -185,21 +186,16 @@ impl RequestUnpack for ExecutionManagerIdRequest {
...
+/// Unpacks [`RegisterSchedulerRequest`] into the scheduler host and port.
 impl RequestUnpack for RegisterSchedulerRequest {
-    type Unpacked = (IpAddr, u16);
+    type Unpacked = (Host, u16);
 
     fn unpack(self) -> Result<Self::Unpacked, UnpackError> {
-        let ip_address = self
-            .ip_address
-            .parse::<IpAddr>()
-            .map_err(|error| invalid_argument(format!("invalid IP address: {error}")))?;
```
20001020ycx, line 199: So what if the input is an IP, and the IP is invalid, would deleting the check cause any regression

20001020ycx (commented):
Thanks for the work and get the PR out so promptly. I don't have too much comment on the actual code itself as most of them are straight update from ip to host, but I do have a few question regarding the failure model in networking:

While reviewing https://github.com/y-scope/spider/pull/401/changes#r3600548223, I am actually wondering about the failure cases. I think it would be best if we have unit test to cover the failure cases unless we can already clearly define the behavior themselves: 
1. Invalid ipv4/ipv6 IP
2. DNS fails to resolve a domain name

I think what I am looking for is the e2e example of how an IP/host is passed through the configs, and our handle when the input are invalid (as I often config them incorrectly at the first few attempts).

sitaowang1998:
> Thanks for the explanation, can we also include this in the PR so the reference is more clear (unless this violate any policies you guys have in Spider).
> 
> Also a few questions:
> 
> 1. Why do we need host in the configuration? Seems conflicting with the config `advertised_endpoint`
> 
> 2. Why quotation mark around "spider-scheduler"? I assume `spider-scheduler` is the URL name, and we are omitting the `http://` right? @LinZhihao-723 are you okay with this?

Thanks for the suggestion. I have added the example in the PR description.

For your questions:
1. The top level `host` is the address that scheduler's gRPC server binds on. It is likely to always be `0.0.0.0`, but leaving it to be configurable could be useful later. The `advertised_endpoint` is the address that is going to be saved in database for execution managers to connect to (or passed to the em by config).
2. This is a good question. For now, `host` only refers to the host name, not the full URL, which will contains port in itself. The `http://` will be added in rust code when a client is trying to make connection. If there is possibility that we will have other schema than `http://`, then we should replace the whole `EndpointConfig` with a URL string.

20001020ycx:
> It is likely to always be 0.0.0.0, but leaving it to be configurable could be useful later. 

Agreed with all other points except this one, wouldn't this confuse user?

components/spider-proto-rust/src/unpack/storage.rs
sitaowang1998: If the DNS name is syntactically wrong, then in `EndpointConfig::endpoint` function, the `Endpoint::from_shared` will return an error, and an error will be logged in the respecting `grpc_server.rs` and the binary exits with an error.

If the DNS name is syntactically correct, and not an existing address that is connectable, then the gRPC client's `connect` will return an error, an error will be logged and the binary exits with an error.

I am not sure if these are graceful enough in your opinion.

components/spider-proto-rust/src/unpack/storage.rs
sitaowang1998: There is no good way to tell if it is an invalid IP or just not an IP address. I think the best way is to leave it to the other functions mentioned in https://github.com/y-scope/spider/pull/401#discussion_r3600670627 to handle the failure.

sitaowang1998:
> > It is likely to always be 0.0.0.0, but leaving it to be configurable could be useful later.
> 
> Agreed with all other points except this one, wouldn't this confuse user?

We can provide `0.0.0.0` as the default value, and it should be clear what this `host` is expecting. We can even not expose it as a configurable value in helm.

The final solution should be a user doc that explains all the configuration, but we don't have time for it now.

components/spider-proto-rust/src/unpack/storage.rs
20001020ycx: really? isnt that you are doing this in the deleted line? 

Say user inputs 256.256.256.256, this is not a valid IP right? We can easily parse it (of course though some lib) and let the user knows

components/spider-proto-rust/src/unpack/storage.rs
20001020ycx: Its good that you are thinking about this, but maybe its my problem that I dont really understand what exactly is this "error" you are talking about.

can you tell me say if I input an invalid ip or user name, what would the user see? 

I am okay if you think what you presented to user can clearly show what is the root cause.

sitaowang1998: For example, if user set an invalid storage hostname in scheduler config, scheduler will fail with "Failed to parse storage endpoint." or "Failed to connect to storage gRPC service.". It also includes the message from underlying library.

components/spider-proto-rust/src/unpack/storage.rs
sitaowang1998: Previously, it can only be an IPv4, Thus, we know for sure if it is not a valid IPv4 address, it is invalid. The [`AddrParseError`](https://doc.rust-lang.org/std/net/struct.AddrParseError.html) is not specific enough to inform us if it is a invalid IP address out of range, or just something totally does not look like an IP address.

LinZhihao-723 (approved):

PR #403, opened 2026-07-16
feat(helm): Add scheduler service to the Helm chart.

20001020ycx:
# Description

This PR adds the **scheduler** service so `helm install` brings up storage + scheduler, with the scheduler registering and polling storage for submitted jobs.

> [!NOTE]  
Part of the ongoing Spider Huntsman Kubernetes integration. Up to this PR, Helm chart has storage + database and scheduler, leaving execution manager, and task executor to be integrated in the follow-up PRs.

> [!NOTE]
> This PR needs review from two teams:
>
> - Cloud Team: Please review the Helm chart itself
> - Spider Team: Please review the user-facing Spider's scheduler configuration. You can leave the rest of the chart to the Cloud reviewer.

# Checklist

  breaking change.

# Validation performed

- All CI jobs passed
- We run the exact steps as #384's validation section for regression verification, then perform below steps to verify the deployed Helm chart in this PR works end-to-end in a kind cluster:

1. Confirm the scheduler registered itself with storage: one row whose ip_address is the scheduler Pod IP and port is the scheduler gRPC port (50052):
```
  kubectl exec spider-database-0 -- mysql -uspider-user -pspider-password spider-db -e "SELECT * FROM schedulers;"
```
  Expected table:
 | id | ip_address | port | created_at |
  |----|------------|------|------------|
  | 3 | 10.244.0.12 | 50052 | 2026-07-15 23:37:36 |

## Summary by CodeRabbit

- **New Features**
  - Added a Helm-managed Kubernetes Deployment for the scheduler component.
  - Exposed the scheduler via a gRPC Service, including readiness and liveness checks.
  - Introduced configurable scheduler settings (image, port, connection pool) and round-robin runtime options.
  - Generated scheduler configuration now includes storage connectivity settings.

- **Chores**
  - Bumped the Helm chart version to **0.1.2**.

### DIFF
--- a/tools/deployment/spider-helm/templates/configmap.yaml
+++ b/tools/deployment/spider-helm/templates/configmap.yaml
@@ -1,0 +1,19 @@
+apiVersion: "v1"
+kind: "ConfigMap"
+metadata:
+  name: {{ include "spider.componentFullname" (dict "root" . "component" "config") }}
+  labels:
+    {{- include "spider.labels" . | nindent 4 }}
+data:
+  storage.yaml: |
+    host: "0.0.0.0"
+    port: {{ .Values.spiderConfig.storage.port }}
+    runtime:
+      db_config:
+        host: {{ include "spider.databaseHost" . | quote }}
+        max_connections: {{ .Values.spiderConfig.database.max_connections }}
+        name: {{ .Values.spiderConfig.database.name | quote }}
+        port: {{ include "spider.databasePort" . }}
+      {{- with .Values.spiderConfig.storage.runtime }}
+      {{- toYaml . | nindent 6 }}
+      {{- end }}
--- a/tools/deployment/spider-helm/templates/configmap.yaml
+++ b/tools/deployment/spider-helm/templates/configmap.yaml
@@ -7,15 +7,18 @@ metadata:
 data:
   scheduler.yaml: |
     connection_pool_size: {{ .Values.spiderConfig.scheduler.connection_pool_size }}
+    host: "0.0.0.0"
+    port: {{ .Values.spiderConfig.scheduler.port }}
     runtime:
-      host: "__SCHEDULER_HOST__"
-      port: {{ .Values.spiderConfig.scheduler.port }}
+      advertised_endpoint:
+        host: {{ include "spider.componentFullname" (dict "root" . "component" "scheduler") | quote }}
+        port: {{ .Values.spiderConfig.scheduler.port }}
       {{- $scheduler := .Values.spiderConfig.scheduler.runtime.scheduler }}
       {{- $algorithm := keys $scheduler | first }}
       scheduler: !{{ $algorithm }}
         {{- toYaml (index $scheduler $algorithm) | nindent 8 }}
     storage_endpoint:
-      host: "__STORAGE_HOST__"
+      host: {{ include "spider.componentFullname" (dict "root" . "component" "storage") | quote }}
       port: {{ .Values.spiderConfig.storage.port }}
   storage.yaml: |
     host: "0.0.0.0"
--- a/tools/deployment/spider-helm/templates/configmap.yaml
+++ b/tools/deployment/spider-helm/templates/configmap.yaml
@@ -5,6 +5,18 @@ metadata:
   labels:
     {{- include "spider.labels" . | nindent 4 }}
 data:
+  scheduler.yaml: |
+    connection_pool_size: {{ .Values.spiderConfig.scheduler.connection_pool_size }}
+    runtime:
+      host: "__SCHEDULER_HOST__"
+      port: {{ .Values.spiderConfig.scheduler.port }}
+      {{- $scheduler := .Values.spiderConfig.scheduler.runtime.scheduler }}
+      {{- $algorithm := keys $scheduler | first }}
+      scheduler: !{{ $algorithm }}
+        {{- toYaml (index $scheduler $algorithm) | nindent 8 }}
+    storage_endpoint:
+      host: "__STORAGE_HOST__"
+      port: {{ .Values.spiderConfig.storage.port }}
   storage.yaml: |
     host: "0.0.0.0"
     port: {{ .Values.spiderConfig.storage.port }}
--- a/tools/deployment/spider-helm/templates/configmap.yaml
+++ b/tools/deployment/spider-helm/templates/configmap.yaml
@@ -13,10 +13,10 @@ data:
       advertised_endpoint:
         host: {{ include "spider.componentFullname" (dict "root" . "component" "scheduler") | quote }}
         port: {{ .Values.spiderConfig.scheduler.port }}
-      {{- $scheduler := .Values.spiderConfig.scheduler.runtime.scheduler }}
-      {{- $algorithm := keys $scheduler | first }}
-      scheduler: !{{ $algorithm }}
-        {{- toYaml (index $scheduler $algorithm) | nindent 8 }}
+      scheduler:
+        policy: {{ .Values.spiderConfig.scheduler.runtime.scheduler.policy | quote }}
+        config:
+          {{- toYaml .Values.spiderConfig.scheduler.runtime.scheduler.config | nindent 10 }}
     storage_endpoint:
       host: {{ include "spider.componentFullname" (dict "root" . "component" "storage") | quote }}
       port: {{ .Values.spiderConfig.storage.port }}
--- a/tools/deployment/spider-helm/templates/scheduler-deployment.yaml
+++ b/tools/deployment/spider-helm/templates/scheduler-deployment.yaml
@@ -18,7 +18,7 @@ spec:
         app.kubernetes.io/component: "scheduler"
     spec:
       initContainers:
-        # Resolve IP addresses of the scheduler and storage services, and render the scheduler
+        # Resolves IP addresses of the scheduler and storage services, and renders the scheduler
         # configuration file with the resolved addresses.
         - name: "render-config"
           image: {{ include "spider.imageRef" (dict "root" . "component" "scheduler") | quote }}
--- a/tools/deployment/spider-helm/templates/scheduler-deployment.yaml
+++ b/tools/deployment/spider-helm/templates/scheduler-deployment.yaml
@@ -1,0 +1,79 @@
+apiVersion: "apps/v1"
+kind: "Deployment"
+metadata:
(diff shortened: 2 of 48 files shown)


sitaowang1998 (commented):
The scheduler configuration conforms to the current Spider's structure. Though I strongly suggest we merge #401 first and get rid of the hostname resolution part.

hoophalab:
Let's merge hostname resolution PR first. I guess the current solution might not work (need a full restart) when a pod is rescheduled to another node and got assigned another IP.

tools/deployment/spider-helm/values.yaml
```diff
@@ -30,6 +34,21 @@ spiderConfig:
     root_password: "spider-root-password"
     username: "spider-user"
 
+  scheduler:
+    connection_pool_size: 4
+    port: 50052
```
hoophalab, line 39: How about just 50051?

hoophalab (dismissed):
LGTM. Validated that the charts start correctly.

tools/deployment/spider-helm/values.yaml
sitaowang1998: I have tested that setting it to 50051 works. However, I don't see a reason to make it the same with storage's port.

tools/deployment/spider-helm/templates/configmap.yaml
```diff
@@ -5,6 +5,21 @@ metadata:
   labels:
     {{- include "spider.labels" . | nindent 4 }}
 data:
+  scheduler.yaml: |
+    connection_pool_size: {{ .Values.spiderConfig.scheduler.connection_pool_size }}
+    host: "0.0.0.0"
+    port: {{ .Values.spiderConfig.scheduler.port }}
+    runtime:
+      advertised_endpoint:
```
junhaoliao, line 13: (note to self: would `public_endpoint` be more idiomatic?

junhaoliao: though `advertised_endpoint` clearly reflects the purpose. i guess it's fine to keep `advertised_endpoint` then

tools/deployment/spider-helm/values.yaml
```diff
@@ -30,6 +34,21 @@ spiderConfig:
...
+      scheduler:
+        round_robin:
+          active_job_queue_capacity: 64
+          cleanup_ready_task_capacity: 256
+          commit_ready_task_capacity: 256
+          dispatch_queue_capacity: 64
+          finalizing_job_expiration_timeout_sec: 300
+          ready_task_capacity: 65536
+          storage_poll_timeout_ms: 10
+          tick_interval_ms: 5
```
junhaoliao: current interface

```yaml
runtime:
  scheduler:
    round_robin:
      tick_interval_ms: 5
      dispatch_queue_capacity: 64
```

assuming we will support FIFO - users might want to `helm install` with:
```sh
--set spiderConfig.scheduler.runtime.scheduler.fifo.queue_capacity=128
```

Helm produces:
```yaml
runtime:
  scheduler:
    round_robin:
      tick_interval_ms: 5
      dispatch_queue_capacity: 64
    fifo:
      queue_capacity: 128
```

now two algorithms are specified but only the first one is selected

---

->

can we make the scheduling algorithm an explicit value instead of encoding it as the only key in a map?

```yaml
    runtime:
      scheduler:
        algorithm: "round_robin"
        config:
          active_job_queue_capacity: 64
          cleanup_ready_task_capacity: 256
          commit_ready_task_capacity: 256
          dispatch_queue_capacity: 64
          finalizing_job_expiration_timeout_sec: 300
          ready_task_capacity: 65536
          storage_poll_timeout_ms: 10
          tick_interval_ms: 5
```

the explicit structure is easier to read and allows focused Helm overrides (in helm install):

```shell
 --set spiderConfig.scheduler.runtime.scheduling.algorithm=round_robin
 --set spiderConfig.scheduler.runtime.scheduling.config.tick_interval_ms=10
```

it also gives us a clear place to validate the algorithm name and produces more useful errors for unsupported algorithms. adding another implementation no longer requires interpreting a map key as both the selector and part of the configuration structure

junhaoliao: or this could be even better

```
runtime:
  scheduling:
    algorithm: "round_robin"
    algorithms:
      round_robin:
        tick_interval_ms: 5
        dispatch_queue_capacity: 64
      fifo:
        queue_capacity: 128
```

this way, we keep all interfaces deterministic

junhaoliao: @sitaowang1998 @LinZhihao-723

tools/deployment/spider-helm/templates/scheduler-deployment.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: "apps/v1"
+kind: "Deployment"
+metadata:
+  name: {{ include "spider.componentFullname" (dict "root" . "component" "scheduler") }}
+  labels:
+    {{- include "spider.labels" . | nindent 4 }}
+    app.kubernetes.io/component: "scheduler"
+spec:
+  replicas: 1
```
junhaoliao, line 9: i suppose we don't plan to support rolling update with `helm update` yet, but when we do, we should consider adding

```yaml
  strategy:
    type: "Recreate"
```

because we expect exactly one instance of the scheduler at a time?

the default strategy is "RollingUpdate" which might cause multiple schedulers to exist

---

i don't think we need to add "Recreate" today, but can we declare that we are lacking such proper `helm update` support, create an issue to track, and note this singleton requirement should be one of the considerations?

tools/deployment/spider-helm/templates/configmap.yaml
```diff
@@ -5,6 +5,21 @@ metadata:
...
+        host: {{ include "spider.componentFullname" (dict "root" . "component" "scheduler") | quote }}
+        port: {{ .Values.spiderConfig.scheduler.port }}
+      {{- $scheduler := .Values.spiderConfig.scheduler.runtime.scheduler }}
+      {{- $algorithm := keys $scheduler | first }}
+      scheduler: !{{ $algorithm }}
+        {{- toYaml (index $scheduler $algorithm) | nindent 8 }}
```
junhaoliao: i could be understanding the end user usages wrong but this could create a problem for end users to supply alternative algorithms when we do support such. see below values.yaml for discussion

junhaoliao (changes requested):
the rest lgtm

tools/deployment/spider-helm/templates/scheduler-deployment.yaml
sitaowang1998: I've created an issue #410 to track this.

tools/deployment/spider-helm/values.yaml
LinZhihao-723: I think the concern you brought up makes sense, but let me first walk through why the config is shaped this way: the current schema matches serde's default *externally-tagged* enum representation, where the variant name is the single key of a map. However, this behavior will be deprecated by the serde library we use.

That said, this shape doesn't actually work with the latest YAML parser. In `yaml_serde` 0.10.4 (the version we pin), the externally-tagged representation isn't encoded as a nested map at all — it requires a YAML tag, i.e. `scheduler: !round_robin` followed by the payload (check the source [here](https://github.com/yaml/yaml-serde/blob/0.10.4/src/de.rs#L1747-L1758)). The nested-map form is rejected outright, and not only when multiple keys are present; even the single-key form fails:

  ```
  failed to parse config file: runtime.scheduler: invalid type: map, expected a YAML tag starting with '!'
  ```
  
So the `values.yaml` in this PR won't parse as written, which is worth fixing regardless of which direction we take here.

But tags lead to another problem: as far as I know, they're unmanageable in Helm. Therefore, I think your first proposed solution makes sense: breaking the config into `algorithm` and `config` should have the following native Rust serde support:

  ```rust
  #[derive(Clone, Debug, Deserialize)]
  #[serde(tag = "algorithm", content = "config", rename_all = "snake_case", deny_unknown_fields)]
  pub enum SchedulerConfig {
      RoundRobin(RoundRobinConfig),
  }
  ```

 I vote for this one, for the following reasons:
  
- It's native to serde, as shown above: we don't need any extra code to interpret the algorithm name and dispatch to the corresponding config type.
- I strongly prefer this over enumerating every supported algorithm in the values file. Considering how many algorithms we may need to support in the future, and how many options each config carries, that list would quickly become bloated. I don't think that would be elegant, lol.
- I don't think setting a config without specifying the algorithm explicitly is a behavior we need to support. Instead, what we can do:
    - Ship `config` empty and default the algorithm to `round_robin` (this needs `#[serde(default)]` on the field plus a `Default` impl selecting `RoundRobin`).
    - Any subsequent `--set spiderConfig.scheduler.runtime.scheduler.config.<field>=<value>` overrides the corresponding default.
    - Any such override assumes the user has explicitly selected the algorithm via `--set spiderConfig.scheduler.runtime.scheduler.algorithm=<algorithm>`: we won't infer it from which config keys were set.

Let me know if this makes sense to you.

junhaoliao: > Considering how many algorithms we may need to support in the future, and how many options each config carries, that list would quickly become bloated. I don't think that would be elegant

it's probably not that bad if we treat the config file as documentation and list all available fields explicitly. keeping the configuration interface deterministic also helps reduce the likelihood of user errors. that said, if you see the list of supported algorithms might grow indefinitely or we might add some of those as extensions (rather than bundled), i think it's fine to keep a single config object that only lists the relevant config of the selected algorithm.

---

then i guess we are settling on

```yaml
    runtime:
      scheduler:
        algorithm: "round_robin"
        config: {}
```

@sitaowang1998 please proceed to make the changes. thanks

LinZhihao-723: > if we treat the config file as documentation

I'm not fully convinced this is a good practice, lol. I prefer the `algorithm + config` solution + better documentation: I think it should be sufficient for AI agent to follow assuming we properly document what's the expected way to override the config.

LinZhihao-723: Btw, maybe `algorithm` -> `core` to match our internal concept for the scheduling algorithm implementation?

junhaoliao: #412 has been merged. we can proceed to update this PR now

junhaoliao (approved):

PR #408, opened 2026-07-18
feat(spider-storage): Release task control blocks when a job leaves the `Running` state.

LinZhihao-723:
# Description

A job's `TaskGraph` holds one TCB per task, and each TCB transitively holds the shared data-dependency payloads that flow between tasks. Once a job leaves `JobState::Running`, none of these regular TCBs are needed anymore: the remaining work is driven by the commit or cleanup task, and any outputs required downstream have already been captured separately. Until now, however, the full TCB vector (and everything it kept alive) was retained for the lifetime of the JCB, so a completed or cancelled job kept holding onto its entire task graph in memory.

This PR adds `TaskGraph::finalize`, which clears the regular TCBs at each transition out of `Running`, releasing the TCBs and — because the data dependencies are only reachable through them — the associated dataflow payloads as well. This aligns the in-memory representation of a finalized job with what the recovery paths already produce, where `tasks` is initialized empty.

## Add `TaskGraph::finalize` (`cache/task.rs`)

- Clears the `tasks` vector, dropping every regular TCB. The `outputs`, commit TCB, and cleanup TCB are left untouched, since the post-`Running` states still depend on them.

## Finalize on transitions out of `Running` (`cache/job.rs`)

- After the success path commits job outputs and transitions to `CommitReady`/`Succeeded`, the task graph is finalized.
- After the cancellation path transitions to `CleanupReady`/`Cancelled`, the task graph is finalized. This runs after `cancel_non_terminal`, so any non-terminal TCBs are still marked cancelled before being released.
- The failure path (transition to `Failed`) deliberately does **not** finalize. A failed job's task graph is retained so that its internal data remains available for future investigation of what went wrong.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved job completion and cancellation handling by finalizing task state transitions consistently.
  * Cleared cached task data after jobs reach a final state, helping prevent stale task information from persisting.

### DIFF
--- a/components/spider-storage/src/cache/job.rs
+++ b/components/spider-storage/src/cache/job.rs
@@ -424,6 +424,7 @@ impl<
                 .send_commit_ready(jcb.owner_id, jcb.id)
                 .await?;
         }
+        job.task_graph.finalize();
         Ok(job.state)
     }
 
@@ -635,6 +636,7 @@ impl<
                 .send_cleanup_ready(jcb.owner_id, jcb.id)
                 .await?;
         }
+        job.task_graph.finalize();
         Ok(job.state)
     }
 
--- a/components/spider-storage/src/cache/task.rs
+++ b/components/spider-storage/src/cache/task.rs
@@ -169,6 +169,15 @@ impl TaskGraph {
         }
     }
 
+    /// Finalizes the task graph, indicating the job state has transitioned from
+    /// [`JobState::Running`] to the next state.
+    ///
+    /// At this stage, all TCBs are explicitly removed. This implies the internal data dependencies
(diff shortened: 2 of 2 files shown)


sitaowang1998 (approved):
LGTM. However, I am wondering if there is a better way to show that only failed path has the full task graph through type system, or just by a good documentation?

LinZhihao-723:
> LGTM. However, I am wondering if there is a better way to show that only failed path has the full task graph through type system, or just by a good documentation?

I think for now only by documentation. The task graph is already internally visible only. We don't need to make this behavior visible to users yet.

PR #411, opened 2026-07-21
feat(helm): Add execution manager service to the Helm chart.

sitaowang1998:
# Description

This PR adds the **execution manager** service so `helm install` brings up all Spider services: database, storage, scheduler, and execution manager.

The chart now:

- Deploys a configurable number of execution manager replicas.
- Generates the execution manager configuration with scheduler, storage, liveness, and task executor settings.
- Adds configurable execution manager image settings.
- Bumps the Helm chart version to **0.1.3**.

> [!NOTE]
> Part of the ongoing Spider Huntsman Kubernetes integration.
>
> This PR is the final PR in this series.

# Checklist

  breaking change.

# Validation performed

- [x] Tested in https://github.com/sitaowang1998/spider/tree/helm-dev-e2e by running Spider e2e integration tests on a cluster in kind by running `helm upgrade --install`.
- [ ] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Added a Helm-managed worker Deployment to run the execution-manager component.
  * Extended Helm values to configure worker replicas and the execution-manager runtime (scheduler/storage connectivity, polling interval, and liveness/heartbeat settings).
  * Automatically generates and mounts the execution-manager configuration via a chart ConfigMap.

* **Chores**
  * Bumped the Helm chart version from `0.1.2` to `0.1.3`.

### DIFF
--- a/tools/deployment/spider-helm/templates/configmap.yaml
+++ b/tools/deployment/spider-helm/templates/configmap.yaml
@@ -1,0 +1,19 @@
+apiVersion: "v1"
+kind: "ConfigMap"
+metadata:
+  name: {{ include "spider.componentFullname" (dict "root" . "component" "config") }}
+  labels:
+    {{- include "spider.labels" . | nindent 4 }}
+data:
+  storage.yaml: |
+    host: "0.0.0.0"
+    port: {{ .Values.spiderConfig.storage.port }}
+    runtime:
+      db_config:
+        host: {{ include "spider.databaseHost" . | quote }}
+        max_connections: {{ .Values.spiderConfig.database.max_connections }}
+        name: {{ .Values.spiderConfig.database.name | quote }}
+        port: {{ include "spider.databasePort" . }}
+      {{- with .Values.spiderConfig.storage.runtime }}
+      {{- toYaml . | nindent 6 }}
+      {{- end }}
--- a/tools/deployment/spider-helm/templates/configmap.yaml
+++ b/tools/deployment/spider-helm/templates/configmap.yaml
@@ -5,6 +5,20 @@ metadata:
   labels:
     {{- include "spider.labels" . | nindent 4 }}
 data:
+  execution-manager.yaml: |
+    connection_pool_size: {{ .Values.spiderConfig.execution_manager.connection_pool_size }}
+    host: "__EXECUTION_MANAGER_HOST__"
+    liveness:
+      {{- toYaml .Values.spiderConfig.execution_manager.liveness | nindent 6 }}
+    scheduler:
+      host: {{ include "spider.componentFullname" (dict "root" . "component" "scheduler") | quote }}
+      port: {{ .Values.spiderConfig.scheduler.port }}
+    scheduler_poll_wait_ms: {{ .Values.spiderConfig.execution_manager.scheduler_poll_wait_ms }}
+    storage:
+      host: {{ include "spider.componentFullname" (dict "root" . "component" "storage") | quote }}
+      port: {{ .Values.spiderConfig.storage.port }}
+    task_executor:
+      {{- toYaml .Values.spiderConfig.execution_manager.task_executor | nindent 6 }}
   scheduler.yaml: |
     connection_pool_size: {{ .Values.spiderConfig.scheduler.connection_pool_size }}
     host: "0.0.0.0"
--- a/tools/deployment/spider-helm/templates/configmap.yaml
+++ b/tools/deployment/spider-helm/templates/configmap.yaml
@@ -7,15 +7,18 @@ metadata:
 data:
   scheduler.yaml: |
     connection_pool_size: {{ .Values.spiderConfig.scheduler.connection_pool_size }}
+    host: "0.0.0.0"
+    port: {{ .Values.spiderConfig.scheduler.port }}
     runtime:
-      host: "__SCHEDULER_HOST__"
-      port: {{ .Values.spiderConfig.scheduler.port }}
+      advertised_endpoint:
+        host: {{ include "spider.componentFullname" (dict "root" . "component" "scheduler") | quote }}
+        port: {{ .Values.spiderConfig.scheduler.port }}
       {{- $scheduler := .Values.spiderConfig.scheduler.runtime.scheduler }}
       {{- $algorithm := keys $scheduler | first }}
       scheduler: !{{ $algorithm }}
         {{- toYaml (index $scheduler $algorithm) | nindent 8 }}
     storage_endpoint:
-      host: "__STORAGE_HOST__"
+      host: {{ include "spider.componentFullname" (dict "root" . "component" "storage") | quote }}
       port: {{ .Values.spiderConfig.storage.port }}
   storage.yaml: |
     host: "0.0.0.0"
--- a/tools/deployment/spider-helm/templates/configmap.yaml
+++ b/tools/deployment/spider-helm/templates/configmap.yaml
@@ -26,10 +26,10 @@ data:
       advertised_endpoint:
         host: {{ include "spider.componentFullname" (dict "root" . "component" "scheduler") | quote }}
         port: {{ .Values.spiderConfig.scheduler.port }}
-      {{- $scheduler := .Values.spiderConfig.scheduler.runtime.scheduler }}
-      {{- $algorithm := keys $scheduler | first }}
-      scheduler: !{{ $algorithm }}
-        {{- toYaml (index $scheduler $algorithm) | nindent 8 }}
+      scheduler:
+        policy: {{ .Values.spiderConfig.scheduler.runtime.scheduler.policy | quote }}
+        config:
+          {{- toYaml .Values.spiderConfig.scheduler.runtime.scheduler.config | nindent 10 }}
     storage_endpoint:
       host: {{ include "spider.componentFullname" (dict "root" . "component" "storage") | quote }}
       port: {{ .Values.spiderConfig.storage.port }}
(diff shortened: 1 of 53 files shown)


20001020ycx (commented):
I appreciate the effort for the e2e testing in your branch. I think it covers what you have implemented pretty well. However, can you first summarize what you have done in the PR description briefly, I had to ask AI to explain it for me. 

Also, a note to myself, we might want to add similar testing as what Sitao did, at least the testing scope, as part of the helm's testing infra. Currently, none is provided.

tools/deployment/spider-helm/values.yaml
```diff
@@ -50,6 +54,19 @@ spiderConfig:
           storage_poll_timeout_ms: 10
           tick_interval_ms: 5
 
+  execution_manager:
```
20001020ycx: Let's make these configuration alphabetically ordered

tools/deployment/spider-helm/templates/configmap.yaml
```diff
@@ -5,6 +5,19 @@ metadata:
   labels:
     {{- include "spider.labels" . | nindent 4 }}
 data:
+  execution-manager.yaml: |
```
20001020ycx, line 8: Let's make these configuration alphabetically ordered

20001020ycx (dismissed):
LGTM, just some nits for you to address.

LinZhihao-723 (commented):
In general, the use of "worker" makes sense to me.
Shall we update the PR title as well?
Still need review from someone with more k8s background.

tools/deployment/spider-helm/templates/configmap.yaml
sitaowang1998: They are already alphabetically ordered.

20001020ycx:
> In general, the use of "worker" makes sense to me. Shall we update the PR title as well? Still need review from someone with more k8s background.

I second this proposal, at least worker is the terminology that we use extensively in CLP package integration as well.

tools/deployment/spider-helm/values.yaml
```diff
@@ -34,6 +38,19 @@ spiderConfig:
     root_password: "spider-root-password"
     username: "spider-user"
 
+  execution_manager:
+    connection_pool_size: 4
+    liveness:
+      scheduler_heartbeat_interval_sec: 10
+      storage_heartbeat_interval_sec: 10
+    replicas: 1
```
hoophalab: Probably move `replicas` outside the config? This configures the worker, rather than the execution manager. We could also pass everything under execution_manager directly to the yaml file in configmap `{{- toYaml .Values.spiderConfig.execution_manager}}`

tools/deployment/spider-helm/values.yaml
```diff
@@ -34,6 +38,19 @@ spiderConfig:
...
+    scheduler_poll_wait_ms: 1000
+    task_executor:
+      bin_path: "/usr/local/bin/spider-task-executor"
+      inherited_env: []
+      log_dir: "/tmp/spider/task-executor"
```
hoophalab, line 50: I guess `/tmp` is meant for disposable files. `/var/log/spider/task-executor` might be more suitable.

tools/deployment/spider-helm/templates/worker-deployment.yaml
```diff
@@ -0,0 +1,33 @@
...
+      {{- include "spider.selectorLabels" . | nindent 6 }}
+      app.kubernetes.io/component: "worker"
+  template:
+    metadata:
+      labels:
+        {{- include "spider.labels" . | nindent 8 }}
+        app.kubernetes.io/component: "worker"
+    spec:
+      containers:
+        - name: "execution-manager"
```
hoophalab, line 21: Is there any way to inject a liveness probe? I'm not sure how spider is designed: I guess the scheduler will reschedule the job if one worker doesn't respond in X minutes, but will the worker restart automatically if there is an unexpected deadlock in user.so, or the execution manager itself?

hoophalab (dismissed):
LTGM, some comments.

tools/deployment/spider-helm/templates/worker-deployment.yaml
sitaowang1998: User can set a hard timeout on a task, and task executor will be killed and retried if it exceeds the timeout.

hoophalab: I guess the execution manager could still deadlock and this pod just hangs forever?

tools/deployment/spider-helm/templates/worker-deployment.yaml
```diff
@@ -0,0 +1,33 @@
...
+          image: {{ include "spider.imageRef" (dict "root" . "component" "worker") | quote }}
+          imagePullPolicy: {{ .Values.image.worker.pullPolicy | quote }}
+          command: ["spider-execution-manager", "--config", "/etc/spider/execution-manager.yaml"]
```
junhaoliao, line 24: i believe pod termination skips the graceful path here. Kubernetes uses `SIGTERM` because the worker image has no `STOPSIGNAL`, while the binary only awaits `ctrl_c()` / `SIGINT`; the scheduler shutdown RPC is therefore bypassed and assignments wait for cutoff recovery.

can we handle `SIGTERM` in the binary?

tools/deployment/spider-helm/values.yaml
```diff
@@ -34,6 +38,19 @@ spiderConfig:
...
+  execution_manager:
+    connection_pool_size: 4
+    liveness:
+      scheduler_heartbeat_interval_sec: 10
+      storage_heartbeat_interval_sec: 10
+    replicas: 1
+    scheduler_poll_wait_ms: 1000
+    task_executor:
+      bin_path: "/usr/local/bin/spider-task-executor"
+      inherited_env: []
```
junhaoliao, line 49: `inherited_env` names variables to forward into the task executor, but the Deployment exposes no `env` / `envFrom` hook, so operators cannot inject Secret-backed values through the chart. can we expose worker environment configuration alongside this option?

junhaoliao (changes requested):
the rest lgtm. deferring to @hoophalab 's review

tools/deployment/spider-helm/templates/worker-deployment.yaml
hoophalab: i guess this is already tracked in https://github.com/y-scope/spider/issues/393?

tools/deployment/spider-helm/templates/worker-deployment.yaml
sitaowang1998: Good question. For this PR we just assume that it will not hang. We will design and implement a proper liveness hook later.

tools/deployment/spider-helm/values.yaml
sitaowang1998: The logs are written to stderr, and our goal is to route it into container's stdout. But I'll change the `log_dir` in case we change the behavior latter.

sitaowang1998: The problem is `/var/log` is not writtable by a non-root user, such as `spider-user`. Should we use a user-owned directory?

tools/deployment/spider-helm/templates/worker-deployment.yaml
LinZhihao-723: Yes, we're aware of this issue. We (the Spider team) should try to address it next week.

tools/deployment/spider-helm/values.yaml
LinZhihao-723: Yeah, I think this is missed.

tools/deployment/spider-helm/values.yaml
hoophalab: hmm right. How about `~/.cache/spider/task-executor`? But I haven't seen any pod use user home folder for log.

tools/deployment/spider-helm/values.yaml
LinZhihao-723: I think the key is to allow users to define environment variables in the execution manager container. For now, we can leave `inherited_env` empty: the current configuration will ensure all the forked processes to automatically inherit the execution manager's environment. We only need to make sure env variables can be passed into the execution manager. @sitaowang1998 @20001020ycx Can you check this?

tools/deployment/spider-helm/values.yaml
hoophalab: Or maybe we just mount an emptydir to /var/log/spider and set uid so that spider-user is writable?

sitaowang1998: My bad. The logs are written to stderr, and currently redirected to `{log_dir}/{em_id}-{executor_id}.log`.

sitaowang1998: According to an offline discussion with Zhihao, we decided to use `/tmp/spider/task-executor/` for now. Next week we will fix the log routing and remove the `log_dir` entirely.

tools/deployment/spider-helm/values.yaml
20001020ycx: > I think the key is to allow users to define environment variables in the execution manager container. 

This is exactly what I am doing in #2408

20001020ycx: But I do wonder should we plan out adding some test .so within spider's worker container, such that the spider k8s can be tested as a standalone entity? 

Of course, not in this PR, I am okay with leaving it empty in this PR. In fact, this should be left as empty even we have the test.so as the CLP_HOME shall be defined in the container image rather than here.

sitaowang1998: I don't think defining it in image works. Some envs are only known at deployment time. My plan is to add a `extra_envs` map in values, default to empty, and fill it in worker deployment's `env`. User can define `extra_envs` and pass in the values they need.

junhaoliao (approved):

PR #421, opened 2026-07-30
build(huntsman): Manage crate dependency versions centrally through the Cargo workspace; Set all crates' version to `0.1.0-dev`.

LinZhihao-723:
# Description

## Summary

This PR introduces workspace-level dependency and package inheritance so that every external crate is declared exactly once, in the root `Cargo.toml`, instead of being pinned independently in each of the 21 member manifests. Before this change a version bump was an N-file edit, and five crates had already drifted apart across members (`tokio` was simultaneously `1.50.0` and `1.52.3`, `anyhow` `1.0.98` and `1.0.102`, `tokio-util` `0.7` and `0.7.18`, `tracing` `0.1.41` and `0.1.44`, `tracing-subscriber` `0.3.19` and `0.3.23`). All drift is resolved and every dependency is aligned to the newest semver-compatible release.

Every bump stays within its crate's current major, so **no Rust source file is modified by this PR** — it is confined to manifests and `Cargo.lock`. Where a crate has a breaking major available, it is pinned to the last release before that break rather than upgraded; those upgrades are deferred to individual PRs (see below).

The change is intended to be behavior-neutral. It was verified as such by diffing the fully resolved feature graph (`cargo metadata`) before and after: of roughly 318 packages common to both graphs, exactly one differs, and that difference comes from an upstream manifest change rather than from this migration (see Notes).

All member crates additionally move from `version = "0.1.0"` to a shared `version = "0.1.0-dev"`, and the Helm chart's `appVersion` is updated to match.

## Workspace inheritance (root `Cargo.toml`)

- Adds `[workspace.package]` holding the two fields every member repeated identically: `version = "0.1.0-dev"` and `edition = "2024"`.
- Adds `[workspace.dependencies]` with 39 external crates and the 13 internal path dependencies. Internal crates are now referenced by name rather than by a relative path, which removes path spellings like `../../../../components/spider-tdl` from the example crates.
- `resolver`, `members`, and `default-members` are unchanged.

## Member manifests (21 crates)

- `[package]` now uses `version = { workspace = true }` and `edition = { workspace = true }`.
- Every entry in `[dependencies]`, `[dev-dependencies]`, and `[build-dependencies]` becomes `{ workspace = true }`. Members retain only the features they need *in addition* to what the workspace entry already provides, so the common case collapses to a single bare inheritance.
- The `{ workspace = true }` inline-table form is used throughout rather than the dotted `dep.workspace = true` form, for consistency and to avoid tripping the `dotted-keys-out-of-order` lint rule configured in `tombi.toml`.
- Features that remain member-local: `sqlx` adds `runtime-tokio` in `spider-storage`; `syn` adds `full` in `spider-tdl-derive`; `tracing` adds `attributes` in `spider-storage`; `spider-tdl` adds `derive` in `huntsman-complex`, `huntsman-nn-tasks`, and `integration-test-tasks`. The `tokio` and `tokio-util` workspace entries deliberately carry no features, because the per-member feature sets differ substantially, so each member keeps its full list.
- `spider-tdl-derive` remains an optional dependency of `spider-tdl` (`{ workspace = true, optional = true }`).

## `default-features = false` moves up to the workspace

`futures-util`, `tracing`, and `tracing-subscriber` were each declared with `default-features = false` in individual members. Cargo ignores a member's `default-features = false` when the workspace entry leaves defaults enabled, so this key now lives on the workspace entry instead; leaving it in the members would have silently re-enabled default features.

For `tracing` this meant restating the intent slightly: the workspace entry is `default-features = false, features = ["std"]`, and `spider-storage` — which previously took defaults plus `attributes` — now declares `attributes` explicitly. The resolved feature set is unchanged (verified below).

### Aligned to the newest compatible release

Where a crate had drifted, the drifted requirements are listed together in "Before".

| Crate | Before | After | Note |
| --- | --- | --- | --- |
| `anyhow` | `1.0.98`, `1.0.102` | `1.0.104` | Drift resolved |
| `async-channel` | `2.3.1` | `2.5.0` | |
| `async-trait` | `0.1.89` | `0.1.91` | |
| `bytes` | `1.10` | `1.12.1` | Requirement was not patch-exact |
| `clap` | `4.6.1` | `4.6.4` | |
| `const_format` | `0.2.35` | `0.2.36` | |
| `dashmap` | `6.1.0` | `6.2.1` | |
| `futures-util` | `0.3.31` | `0.3.33` | |
| `libloading` | `0.8.5` | `0.8.9` | Major deferred |
| `proc-macro2` | `1.0.106` | `1.0.107` | |
| `quote` | `1.0.45` | `1.0.47` | |
| `rand` | `0.9.1` | `0.9.5` | Major deferred |
| `semver` | `1.0.27` | `1.0.28` | |
| `serde` | `1.0.228` | `1.0.229` | |
| `serde_json` | `1.0.149` | `1.0.151` | |
| `serial_test` | `3.2.0` | `3.5.0` | Major deferred |
| `syn` | `2.0.117` | `2.0.119` | Major deferred |
| `thiserror` | `2.0.18` | `2.0.19` | |
| `tokio` | `1.50.0`, `1.52.3` | `1.53.1` | Drift resolved |
| `tokio-util` | `0.7`, `0.7.18` | `0.7.19` | Drift resolved; requirement was not patch-exact |
| `tracing` | `0.1.41`, `0.1.44` | `0.1.44` | Drift resolved |
| `tracing-subscriber` | `0.3.19`, `0.3.23` | `0.3.23` | Drift resolved |

Every entry in `[workspace.dependencies]` is now patch-exact, per the repository convention. `bytes` and `tokio-util` previously named no patch version, and those two pre-existing deviations are corrected here.

### Breaking major available, deliberately not taken

Each of these is pinned to the newest release within its current major — the last version before the API break — so that this PR needs no source changes. Each upgrade is deferred to its own PR so it can be reviewed against its own migration guide.

| Crate | Pinned at | Major available | Why deferred |
| --- | --- | --- | --- |
| `bincode` | `1.3.3` | `3.0.0` | `3.0.0` is a tombstone, not a usable release; crate is unmaintained |
| `libloading` | `0.8.9` | `0.9.0` | `dlerror` text moves out of `Display`, degrading load-failure diagnostics |
| `rand` | `0.9.5` | `0.10.2` | `Rng` renamed to `RngExt`; `SeedableRng::from_os_rng` removed |
| `serial_test` | `3.5.0` | `4.0.1` | MSRV bump and internal `syn` 3 only; no user-facing API change |
| `sqlx` | `0.8.6` | `0.9.0` | `ArgumentBuffer` loses its lifetime; `query*()` narrowed; MySQL behavior changes |
| `syn` | `2.0.119` | `3.0.3` | Workspace stays on `syn` 2.x deliberately; see Notes |
| `tabled` | `0.20.0` | `0.21.0` | `Charset`, `TabSize`, and `ColumnNames` reworked; none used here |

`sqlx` and `libloading` are worth calling out, because their majors would each require code changes that this pinning avoids:

- `sqlx` `0.9.0` removes the lifetime parameter from `ArgumentBuffer` (which the `Encode` impls in `spider-core/src/types/id.rs` and the `MySqlEnum` derive in `spider-derive/src/mysql.rs` both name) and narrows `query*()` to `impl SqlSafeStr` (which the dynamically built statements in `spider-storage/src/db/mariadb.rs` would need to opt out of). It also changes MySQL connection behavior — `SET NAMES utf8mb4 COLLATE utf8_general_ci` is no longer sent by default — and touches how database errors are surfaced, which matters here because duplicate-key and foreign-key handling in `mariadb.rs` classifies errors by downcasting to `MySqlDatabaseError` and matching on the error number. Staying on `0.8.6` keeps all of that as-is.
- `libloading` `0.9.0` moves the underlying `dlerror` text out of `Error`'s `Display` implementation and behind `Error::source()`. Because `ExecutorError`'s `From<libloading::Error>` impl formats via `to_string()`, taking that major without also rewriting the conversion would silently reduce every TDL package load failure to `"dlopen failed"`. Staying on `0.8.9` preserves the current diagnostics.

`bincode` is a special case in that table: it is not merely deferred. The crate is unmaintained, and its `3.0.0` release is a deliberate tombstone whose `src/lib.rs` is a bare `compile_error!`, so it cannot be depended on at all — `2.0.1` is the last functional release. Migrating off `bincode` entirely is tracked separately; this repo already uses `rmp-serde` for every other serialization path.

## Crate version scheme

All 21 members now report `0.1.0-dev` through workspace inheritance. `0.1.0-dev` is a valid SemVer 2.0.0 pre-release (`dev` is a well-formed alphanumeric identifier) and orders below `0.1.0`. No member declares a version requirement on another member — internal dependencies are path-only — so the suffix has no effect on resolution.

## Helm chart (`tools/deployment/spider-helm/Chart.yaml`)

- `appVersion` moves from `"0.1.0"` to `"0.1.0-dev"` so the deployed application version tracks the workspace crate version. `appVersion` is surfaced only as the `app.kubernetes.io/version` label via `spider.labels` in `_helpers.tpl`; container image tags come from `values.yaml` through `spider.imageRef` and are unaffected.
- The chart `version` is bumped `0.1.5` → `0.1.6`, matching the convention that every change to the chart carries a patch bump.

## Notes

- Resolved-feature verification: comparing `cargo metadata` output between a pristine checkout and this branch, exactly one package's feature set differs. `tokio-util` gains `libc`, because `tokio-util` `0.7.19` declares `codec = ["libc"]` where `0.7.18` did not; the workspace already enabled `codec`, so this follows mechanically from the upstream bump. No package loses a feature. `futures-util`, `tracing`, `tracing-subscriber`, `tokio`, `sqlx`, `serde`, `clap`, `syn`, and `spider-tdl` all resolve to byte-identical feature sets.
- `Cargo.lock` gains exactly one package, `syn 3.0.3`, pulled in by the bumped `async-trait`, `clap_derive`, `serde_derive`, and `thiserror-impl`. It resolves only as a host dependency, so it adds one proc-macro-side compilation unit and contributes nothing to the target artifact. `syn 2.x` is unaffected and still required by 23 crates in the graph, including this repo's two proc-macro crates, and `syn 1.0.109` remains reachable through `non-empty-string` → `delegate`, so a single syn version is not achievable at present and the workspace intentionally stays on `syn 2.x`.
- `default-features = false` on `tracing` is effectively inert in this workspace, since `sqlx-core` enables tracing's default features regardless. It is declared for correctness of intent rather than for effect. The same is true of `futures-util`, whose defaults are enabled transitively by `serial_test`, `hyper-util`, `sqlx`, and `tonic`.
- `test-utils` previously declared `serde` with no features and now inherits `derive` from the workspace entry. This is inherent to Cargo's feature unioning and has no resolved effect, as `serde/derive` is already enabled by 12 other members. Avoiding it would mean removing `derive` from the workspace entry and repeating it in those 12 members.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

- **Chores**
  - Centralized package versions, editions, and dependency management across the project.
  - Updated components, examples, and test packages to use shared workspace configuration.
  - Preserved existing dependency features while simplifying project configuration.

- **Deployment**
  - Bumped the Helm chart version to 0.1.6.
  - Updated the application version label to 0.1.0-dev.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,9 +69,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.102"
+version = "1.0.104"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
+checksum = "330a5ed07fa54e4702c9d6c4174f74427fc0ef6e214bbd677ae50a5099946470"
 
 [[package]]
 name = "async-channel"
@@ -87,13 +87,13 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.89"
+version = "0.1.91"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
+checksum = "ae36dc4177970ef04fde5178d3e2429882def40e57a451f919c098f72baa6cec"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.117",
+ "syn 3.0.3",
 ]
 
 [[package]]
@@ -213,9 +213,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.11.1"
+version = "1.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
+checksum = "fc652a48c352aef3ea3aed32080501cf3ef6ed5da78602a020c991775b0aff04"
 
 [[package]]
 name = "cc"
@@ -237,9 +237,9 @@ checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
 name = "clap"
-version = "4.6.1"
+version = "4.6.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1ddb117e43bbf7dacf0a4190fef4d345b9bad68dfc649cb349e7d17d28428e51"
+checksum = "d91e0c145792ef73a6ad36d27c75ac09f1832222a3c209689d90f534685ee5b7"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -247,9 +247,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.6.0"
+version = "4.6.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "714a53001bf66416adb0e2ef5ac857140e7dc3a0c48fb28b2f10762fc4b5069f"
+checksum = "f09628afdcc538b57f3c6341e9c8e9970f18e4a481690a64974d7023bd33548b"
 dependencies = [
  "anstream",
  "anstyle",
@@ -259,14 +259,14 @@ dependencies = [
 
 [[package]]
 name = "clap_derive"
-version = "4.6.1"
+version = "4.6.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f2ce8604710f6733aa641a2b3731eaa1e8b3d9973d5e3565da11800813f997a9"
+checksum = "d012d2b9d65aca7f18f4d9878a045bc17899bba951561ba5ec3c2ba1eed9a061"
 dependencies = [
  "heck",
  "proc-macro2",
  "quote",
- "syn 2.0.117",
+ "syn 3.0.3",
 ]
 
 [[package]]
@@ -443,7 +443,7 @@ checksum = "1ac70aa55017e108007fbaf5aa0f54b021c98f92ff8af59d42eda9da96e3dd4f"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.117",
+ "syn 2.0.119",
 ]
 
 [[package]]
@@ -454,7 +454,7 @@ checksum = "1aaf95b3e5c8f23aa320147307562d361db0ae0d51242340f558153b4eb2439b"
 
 [[package]]
 name = "e2e"
-version = "0.1.0"
+version = "0.1.0-dev"
 dependencies = [
  "anyhow",
  "spider-client",
@@ -474,7 +474,7 @@ dependencies = [
 
 [[package]]
 name = "em-runtime-tests"
-version = "0.1.0"
+version = "0.1.0-dev"
 dependencies = [
  "anyhow",
  "spider-core",
@@ -604,9 +604,9 @@ dependencies = [
 
 [[package]]
 name = "futures-core"
-version = "0.3.32"
+version = "0.3.33"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7e3450815272ef58cec6d564423f6e755e25379b217b0bc688e295ba24df6b1d"
+checksum = "2cd50c473c80f6d7c3670a752354b8e569b1a7cbfdc0419ec88e5edad85e0dc7"
 
 [[package]]
 name = "futures-executor"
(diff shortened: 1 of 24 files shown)


sitaowang1998 (approved):
Shall we create issues to track the pending upgrade and the removal of `bincode`?

LinZhihao-723:
@coderabbitai Based on the PR description, create issues to track the pending upgrades for the following dependencies:
- bincode: Since this library is no longer maintained, we should probably switch to an alternative.
- sqlx: API changed.
- libloading: API changed.
- syn: Other dependencies used by this project depend on different versions of syn: some on 2.x and others on 3.x.

PR #422, opened 2026-07-30
feat(helm): Adds `log_level` support for storage, scheduler and worker.

sitaowang1998:
# Description

This PR adds support for setting log level in storage, scheduler and worker services by:
* Adding `log_level` field in `values.yaml` for all three services, default to `INFO`.
* Adding `RUST_LOG` in env for all three services using the corresponding `log_level`.
* Bumping chart version to satisfy chart linting workflow.

# Checklist

  breaking change.

# Validation performed

* [x] Run e2e test on [test branch](https://github.com/sitaowang1998/spider/tree/helm-dev-e2e) and observer that:
  * `RUST_LOG` set to `INFO` for all three pods by default.
  * Info level logs are available using `kubectl logs`.
* [x] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Added configurable log levels for scheduler, storage, and execution manager services.
  * Added default `INFO` log-level settings to the deployment configuration.
  * Preserved support for additional worker environment variables.

* **Chores**
  * Updated the deployment chart version.

### DIFF
--- a/tools/deployment/spider-helm/templates/worker-deployment.yaml
+++ b/tools/deployment/spider-helm/templates/worker-deployment.yaml
@@ -22,10 +22,12 @@ spec:
           image: {{ include "spider.imageRef" (dict "root" . "component" "worker") | quote }}
           imagePullPolicy: {{ .Values.image.worker.pullPolicy | quote }}
           command: ["spider-execution-manager", "--config", "/etc/spider/execution-manager.yaml"]
-          {{- with .Values.spiderConfig.worker.extra_envs }}
           env:
+            - name: "RUST_LOG"
+              value: {{ .Values.spiderConfig.execution_manager.log_level | quote }}
+            {{- with .Values.spiderConfig.worker.extra_envs }}
             {{- tpl (toYaml .) $ | nindent 12 }}
-          {{- end }}
+            {{- end }}
           volumeMounts:
             - name: "config"
               mountPath: "/etc/spider/execution-manager.yaml"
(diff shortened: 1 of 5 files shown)


tools/deployment/spider-helm/templates/worker-deployment.yaml
```diff
@@ -22,10 +22,12 @@ spec:
           image: {{ include "spider.imageRef" (dict "root" . "component" "worker") | quote }}
           imagePullPolicy: {{ .Values.image.worker.pullPolicy | quote }}
           command: ["spider-execution-manager", "--config", "/etc/spider/execution-manager.yaml"]
-          {{- with .Values.spiderConfig.worker.extra_envs }}
           env:
+            - name: "RUST_LOG"
+              value: {{ .Values.spiderConfig.execution_manager.log_level | quote }}
+            {{- with .Values.spiderConfig.worker.extra_envs }}
```
20001020ycx, line 28: env and extra_evns? 

Let's discuss, I think I am okay with the change in this PR, but might worth some refactoring for the extra_envs, extra_volumn_mounts and extra_volumn

20001020ycx (approved):

tools/deployment/spider-helm/templates/worker-deployment.yaml
sitaowang1998: We reached the following conclusion during offline discussion:
* Storage and scheduler should not accept extra environment variables for safety.
* Worker should have a standalone `log_level` instead of using `extra_envs` to:
  * Keep the same form as with storage and scheduler.
  * Avoid user accidentally overwriting the log level when adding new environment variables.

LinZhihao-723: I agree with the decision to split the log level to the default env.

LinZhihao-723 (approved):
Before merge: I think you should make the scope consistent with other PRs. Looks like we were always using `feat(helm)`, not `feat(spider-helm)`.

PR #427, opened 2026-08-06
feat(docker-compose): Add Docker Compose deployment.

sitaowang1998:
# Description

This PR adds Docker Compose deployment:
- Adds a `compose.yaml` file:
  - Defines MariaDB, storage, scheduler, and worker services.
  - Keeps the database and scheduler on an internal network while publishing storage at a configurable IP and port.
  - Adds storage, scheduler and worker configs that are controlled by `.env` below and will be mounted to corresponding container.
  - Supports configurable worker replicas, log levels, image revision, restart policy.
  - Supports a complete worker image reference for images containing user-provided TDL packages.
- Centralizes all user-facing deployment and Spider settings in `.env` and provides `.env.example` with defaults matching the Helm deployment.
- Adds tasks to validate, start, stop, and clean the deployment.
- Includes the Compose deployment in YAML linting.

# Checklist

  breaking change.

# Validation performed

Verified using Docker Compose v5.4.0.

### Worker image selection

Verified all supported worker image-selection paths.

With neither override set:

``` bash
unset SPIDER_IMAGE_TAG
unset SPIDER_WORKER_CONTAINER_IMAGE_REF

docker compose \
  --file tools/deployment/spider-compose/compose.yaml \
  config --images
```

The worker resolves to:

```
ghcr.io/y-scope/spider/worker:main
```

With only a Spider image tag:

``` bash
SPIDER_IMAGE_TAG="sha-test" \
docker compose \
  --file tools/deployment/spider-compose/compose.yaml \
  config --images
```

The worker resolves to:

```
ghcr.io/y-scope/spider/worker:sha-test
```

With a complete worker image reference:

``` bash
SPIDER_IMAGE_TAG="sha-test" \
SPIDER_WORKER_CONTAINER_IMAGE_REF="example.com/custom/worker@sha256:0123456789abcdef" \
docker compose \
  --file tools/deployment/spider-compose/compose.yaml \
  config --images
```

The complete reference takes precedence:

```
example.com/custom/worker@sha256:0123456789abcdef
```

### Runtime startup

With the required database variables still exported, started the cluster using:

``` bash
task docker:compose:up
```

The resulting deployment contained:
- One healthy MariaDB container.
- One healthy storage container.
- One healthy scheduler container.
- Four running worker containers.

 All seven containers had a restart count of zero.

### Port exposure

Verified the container port bindings:

```
database:  no host bindings
storage:   127.0.0.1:50051 → 50051/tcp
scheduler: no host bindings
```
Therefore, storage is the only service published to the host.

### Network isolation

Started a temporary container attached only to the consumer-facing spider_default network:

``` bash
docker run \
  --rm \
  --network spider_default \
  --entrypoint sh \
  mariadb:10.11.16 \
  -c '
    getent hosts spider-storage >/dev/null &&
    ! getent hosts spider-scheduler >/dev/null &&
    ! getent hosts spider-database >/dev/null
  '
```

The check passed:
- spider-storage was resolvable from the consumer network.
- spider-scheduler was not resolvable.
- spider-database was not resolvable.

Also verified that a worker retained outbound connectivity:

``` bash
docker exec spider-spider-worker-1 \
  timeout 5 bash -c '</dev/tcp/github.com/443'
```
  
### Cleanup

Removed the deployment and persistent database data using:

``` bash
task docker:compose:down
task docker:compose:clean
```

After cleanup, no Compose resources with project name spider remained:
- No containers.
- No networks.
- No database volume.

### End-to-end tests

End-to-end tests is conducted on [a separate test branch](https://github.com/sitaowang1998/spider/tree/docker-compose-test-mount). The test includes:
- Adds wrapper Docker Compose files that mount the e2e testing library.
- Starts the cluster and runs e2e testing client binary, and observes that all tests pass.
- Tears down and cleans up the cluster.

## Summary by CodeRabbit

* **New Features**
  * Added a Docker Compose deployment for Spider with database, storage, scheduler, and worker services.
  * Added commands to validate, start, stop, and clean the Compose environment.
  * Added configurable health checks, persistent storage, ports, logging, and worker scaling.
  * Added a documented environment configuration template with recommended defaults.

* **Chores**
  * Added deployment configuration files to YAML linting.
  * Excluded local deployment credentials from version control.


jackluo923:
Nice work — I ran this end-to-end and it genuinely works, including executing real jobs. Details on the validation below. A couple of small things to fix, one question, and one follow-up suggestion.

## Blocking

### `pull_policy: "always"` makes locally-built images unusable

The repo ships `task docker:build`, which produces `spider-{service}:dev-$USER-$id`. There's no supported path from that to this deployment: `SPIDER_IMAGE_TAG` controls only the tag, not the repository, and even after retagging into `ghcr.io/y-scope/spider/*`, `pull_policy: always` forces a registry round-trip and hard-fails:

```
SPIDER_IMAGE_TAG=local-test task docker:compose:up
→ failed to resolve reference "ghcr.io/y-scope/spider/worker:local-test": not found
```

…with the image present in the local daemon. I had to carry a `pull_policy: never` override through every test to work around it.

Suggested one-liner, keeping the safe default:

```yaml
pull_policy: "${SPIDER_PULL_POLICY:-always}"
```

## Should fix

### Undocumented environment variables

The header comment documents the 3 required vars; these 7 are undocumented:
`SPIDER_IMAGE_TAG`, `SPIDER_WORKER_CONTAINER_IMAGE_REF`, `SPIDER_RESTART_POLICY`, `SPIDER_STORAGE_HOST`, `SPIDER_STORAGE_HOST_PORT`, `SPIDER_WORKER_REPLICAS`, `SPIDER_{STORAGE,SCHEDULER,WORKER}_RUST_LOG`.

Since both tasks `cd` into the compose directory, Compose auto-loads a `.env` from there — an `.env.example` would document all ten at once, and would also make `task docker:compose:validate` runnable without hand-exporting credentials (today it can't run at all without them, and `--no-interpolate` doesn't help since it then trips on `${SPIDER_STORAGE_HOST:-127.0.0.1}` as an IP).

## Question

### Was dropping `--wait-timeout` intentional?

`bb36fec` ("Add clean task") also removed the `--wait-timeout` flag and the `SPIDER_COMPOSE_WAIT_TIMEOUT` knob from `compose:up`, which isn't mentioned in that commit's message.

For what it's worth the behaviour is effectively unchanged — I measured `up --wait` self-terminating after **124s** with this PR's healthcheck timings (`start_period 60s`, `interval 30s`, `retries 3`), versus the 120s that was removed. So no functional concern, but the PR description still advertises "a configurable startup timeout defaulting to 120 seconds", which is now stale either way.

## Nits

- `SPIDER_STORAGE_HOST` is the host IP the published port binds to, not the storage host — confusing next to `SPIDER_STORAGE_DB_*` and `storage.yaml`'s own `host:` (a bind address). `SPIDER_STORAGE_PUBLISH_IP` reads more clearly.
- `SPIDER_WORKER_CONTAINER_IMAGE_REF` carries a `CONTAINER` that no other variable has.
- Worth confirming: is a custom image via `SPIDER_WORKER_CONTAINER_IMAGE_REF` the intended way to supply TDL packages? Helm has `worker.extra_volumes`/`extra_envs` and compose has no equivalent. I verified the custom-image path works end-to-end, so this seems like a deliberate design — just want to confirm it's documented as the mechanism.

## Follow-up (not for this PR)

30 config values are duplicated between `config/*.yaml` and the Helm chart's `values.yaml`. I diffed all 30 mechanically and **found zero drift today** — but nothing enforces it, and `RoundRobinConfig` is `deny_unknown_fields`, so a missed rename breaks the deployment at runtime rather than at validation.

It's worse inside the compose deployment itself, since `config/*.yaml` can't be templated by `compose.yaml`: storage port `50051` appears in **6** places, scheduler port `50052` in **4**, and service hostnames in **5**. Renaming a service silently breaks DNS resolution in the config files, and `compose config --quiet` won't catch it (it validates schema, not cross-references).

This is intrinsic to Compose having no templating — you already de-duplicated everything YAML anchors can reach with `x-service-defaults`/`x-healthcheck-defaults`, which is the right call. Suggestion for later: a `spider-compose.yaml` CI workflow (mirroring `spider-helm.yaml`) that runs `task docker:compose:validate` and asserts the compose configs and `values.yaml` still agree. Right now nothing runs `compose:validate` at all — the new files only get yamllint, which checks style, not compose schema.

## Separate issue, not caused by this PR

Running jobs surfaces a scheduler bug that also affects the Helm deployment — filed separately as #429. Summary: every completed assignment leaves the scheduler retrying forever at ~3 warnings/sec, growing with lifetime task count. Job correctness is unaffected.

Worth calling out that `logging: driver: "local"` contains the blast radius here: I measured it capping at ~69MB retained out of 151MB written. Docker's default `json-file` is unbounded, so that choice is doing real work.

---

## Validation performed

Host is arm64 and the published images are amd64-only, so I built all images natively via `task docker:build` and ran against those.

**Everything in the PR description reproduces:** config validates, missing-var errors are clear, all 3 image-resolution paths resolve as documented, 7 containers come up healthy in ~19s with restart count 0, only storage is published (`127.0.0.1:50051`), and `down`/`clean` leave zero containers/networks/volumes.

**Beyond the description:**

- **Real jobs execute correctly.** I built a worker image with the `integration_test_tasks` TDL package staged at `package_dir`, pointed the deployment at it via `SPIDER_WORKER_CONTAINER_IMAGE_REF`, and submitted `fibonacci(20)` from a container attached only to `spider_default` (the documented consumer topology). Result: `6765`, correct. **9/9 jobs succeeded**, distributed across 3 of 4 workers, all containers still at restart count 0 afterward. Confirmed at each hop: scheduler `num_task_assignments_enqueued: 1` → worker `Task completed successfully, elapsed_us: 1489` → `SELECT state, COUNT(*) FROM jobs` → `Succeeded 9`.
- `depends_on: restart: true` cascade works — restarting the DB correctly cascaded to storage → scheduler → all 4 workers, all recovering healthy.
- Volume persists across `down` and is removed by `clean` (verified with a marker row).
- `down` completes in 2.6s — all three binaries handle SIGTERM, so `stop_grace_period: 60s` is a ceiling, not a cost.
- Config files match the Rust structs exactly, including the 8 `RoundRobinConfig` keys that `deny_unknown_fields` would reject.
- `yamllint --strict` passes on the new files.

sitaowang1998:
I have fixed:
* The pull policy is configurable, with default set as `always`.
* All environment variables are grouped in `.env.example` file.
* Rename `SPIDER_STORAGE_HOST` to `SPIDER_STORAGE_PUBLISH_IP`.

To answer the question:
* `wait-time` is never the goal of the PR. It was added and removed during the composing of the PR, and CodeRabbit mistakenly include it in the PR description.
* `SPIDER_WORKER_CONTAINER_IMAGE_REF` is intended so the users can override and provide their own worker images.

jackluo923 (dismissed):
Changes look good to me, thanks for the quick turnaround.

Nit: I'd rename `SPIDER_WORKER_CONTAINER_IMAGE_REF` to `SPIDER_WORKER_IMAGE_REF` — it's the only variable carrying a `CONTAINER`, and the rest already follow `SPIDER_<COMPONENT>_<SETTING>`. Worth doing before the CLP integration hardcodes the name on its side.

20001020ycx (commented):
Could you perform the e2e testing explained in the notion page, i.e., say you manually compile a TDL package, then mount it with `compose.override.yaml`, finally retrieve result using a client through rust SDK.

I want to make sure this works as this is the agreed upon workflow that we have discussed offline. 

Nit, I really appreciate for the PR description in this PR, its very clear and precise, thanks for the work. But can you remove some redundant details like port check? I am pretty confident that you already handle it well.

tools/deployment/spider-compose/.env.example
```diff
@@ -0,0 +1,20 @@
+# Copy this file to `.env` and replace the development credentials before starting Spider.
+
+# Required database credentials.
+SPIDER_DATABASE_ROOT_PASSWORD=spider-root-password
+SPIDER_STORAGE_DB_PASSWORD=spider-password
+SPIDER_STORAGE_DB_USERNAME=spider-user
+
+# Optional deployment settings and their defaults.
+SPIDER_IMAGE_TAG=main
+SPIDER_PULL_POLICY=always
```
20001020ycx, line 10: I wonder this could be the workflow of a spider developer who wishes to spin up a docker compose against their local chnage:

1. Build; writes build/spider-storage-image.id (and the other two).
task docker:build

2. Point compose at the exact image just built (the env script does this on every `up`):
export SPIDER_STORAGE_IMAGE_REF="$(cat build/spider-storage-image.id)"

3. Bring the stack up.
docker compose up

CLP does this automatically, particularly the 2nd step, like a single `task package` would do the step 1 and 2. Then we do ./sbin/start.sh

Shall we provide wiring and script for the same user experience.

sitaowang1998:
> Could you perform the e2e testing explained in the notion page, i.e., say you manually compile a TDL package, then mount it with `compose.override.yaml`, finally retrieve result using a client through rust SDK.

It is available at https://github.com/sitaowang1998/spider/tree/docker-compose-test-mount. https://github.com/sitaowang1998/spider/blob/docker-compose-test-mount/tests/huntsman/e2e/README.md contains the instructions to run the e2e test.

As for starting the docker compose cluster for local testing, I think that providing the full image ref is better than the current tag approach, and that we can add a new set of tasks for this purpose.
My plan is:
1. Adds `IMAGE_REF` for storage and scheduler, which behave the same as the worker one.
2. Adds a task that:
  - Depends on the existing build task to build the images.
  - Sets up the environment variables and starts the docker compose cluster.
3. Add tasks to tear down and clean up the docker compose cluster started in step 2.
    The existing tear down and clean up tasks should suffice, so probably these will be alias tasks just to avoid confusion.
I am wondering if we need docs for these.

sitaowang1998:
As discussed offline, the `IMAGE_REF` and local testing support will be done in a future PR. @coderabbit Please open an issue to track that.

tools/deployment/spider-compose/.env.example
```diff
@@ -0,0 +1,20 @@
+# Copy this file to `.env` and replace the development credentials before starting Spider.
```
20001020ycx, line 1: > replace the development credentials before starting Spider.

Is this a MUST?

I feel like developer sonly need to copy this to `.env`, and docker compose shall just work right?

tools/deployment/spider-compose/.env.example
```diff
@@ -0,0 +1,20 @@
...
+# Optional deployment settings and their defaults.
+SPIDER_IMAGE_TAG=main
+SPIDER_PULL_POLICY=always
+SPIDER_RESTART_POLICY=on-failure:3
+SPIDER_SCHEDULER_LOG_LEVEL=INFO
+SPIDER_STORAGE_HOST_PORT=50051
+SPIDER_STORAGE_LOG_LEVEL=INFO
+SPIDER_STORAGE_PUBLISH_IP=127.0.0.1
+SPIDER_WORKER_LOG_LEVEL=INFO
+SPIDER_WORKER_REPLICAS=4
```
20001020ycx, line 17: Have you tried increase this to test if this works? I dont see this part in the validation performed section

tools/deployment/spider-compose/.env.example
```diff
@@ -0,0 +1,20 @@
...
+
+# A complete worker image reference takes precedence over `SPIDER_IMAGE_TAG` when set.
+# SPIDER_WORKER_IMAGE_REF=ghcr.io/y-scope/spider/worker:main
```
20001020ycx, line 20: I don't understand the purpose of this comment? is there anything a user would add under this comment

Or do you mean this is how user override the IMAGE_REF if they are using a locally build one

20001020ycx (commented):
Changes LGTM, but I do have a few comment regarding how a user would use the .env.example. Also, can we update the validation performed section as discussed? Thanks

tools/deployment/spider-compose/.env.example
sitaowang1998: It is not a must if security is not important in the use case.

tools/deployment/spider-compose/.env.example
sitaowang1998: This means if both the image tag and image ref is set, image ref is used and the tag is ignored.

tools/deployment/spider-compose/.env.example
sitaowang1998: No. Anything larger than 1 should be sufficient for tests.

sitaowang1998:
I have updated the validation section with e2e testing. However, I don't want to remove the port and network part, as they are important.

tools/deployment/spider-compose/.env.example
20001020ycx: This comment sounds like a must. 

I would imagine the comment be something like: this is an example of the .env, when needing configuration adjustment, please change it and update it in .env

Also add a TODO for configuration doc once it is available, I know you guys are planning to have one

tools/deployment/spider-compose/.env.example
20001020ycx: Then shouldn't this be placed at
```
# Deployment settings.
SPIDER_IMAGE_TAG=main
SPIDER_PULL_POLICY=always
SPIDER_RESTART_POLICY=on-failure:3
```

tools/deployment/spider-compose/compose.yaml
```diff
@@ -0,0 +1,196 @@
...
+      spider-storage:
+        condition: "service_healthy"
+        restart: true
+    deploy:
+      replicas: "${SPIDER_WORKER_REPLICAS:?}"
+
+configs:
+  spider-scheduler-config:
+    content: |
+      connection_pool_size: ${SPIDER_SCHEDULER_CONNECTION_POOL_SIZE:?}
```
20001020ycx: Just for my own understanding: 

why we have no default in here? Are you saying .env.example is the default. What if I forgot copy the .env.example to .env. 

I could be wrong but I feel like leaving the default in here may be a better practice because .env.example is a hidden file that user may not see it at once. Unless what I state here is wrong

tools/deployment/spider-compose/.env.example
```diff
@@ -0,0 +1,52 @@
...
+SPIDER_STORAGE_READY_QUEUE_CLEANUP_CAPACITY=256
+SPIDER_STORAGE_READY_QUEUE_COMMIT_CAPACITY=256
+SPIDER_STORAGE_READY_QUEUE_TASK_CAPACITY=1048576
+
+# Scheduler settings.
+SPIDER_SCHEDULER_LOG_LEVEL=INFO
+SPIDER_SCHEDULER_PORT=50052
+SPIDER_SCHEDULER_CONNECTION_POOL_SIZE=4
+SPIDER_SCHEDULER_POLICY=round_robin
+SPIDER_SCHEDULER_ACTIVE_JOB_QUEUE_CAPACITY=16
```
20001020ycx: I think using the same naming convention as the actual config may be better:

For example:
`SPIDER_SCHEDULER_ACTIVE_JOB_QUEUE_CAPACITY` shall be `SPIDER_SCHEDULER_RUNTIME_SCHEDULER_CONFIG_ACTIVE_JOB_QUEUE_CAPACITY`

tools/deployment/spider-compose/compose.yaml
sitaowang1998: This is a good point. My reason for not providing default is to avoid user, for example me when doing some tests and benchmarks, and especially CLP, to forget a config but still has everything working. It is better to fail fast than hiding the problem.

As for the `.env.example` is a hidden file, this is a valid concern because we currently don't have any doc. Once we have a doc, the user can easily avoid the problem. If they don't read the doc, I'd rather they fail to start everything. Since for now you are going to be the only user in CLP, I don't think you will be trapped.

tools/deployment/spider-compose/.env.example
sitaowang1998: I don't think `RUNTIME` is actually providing any meaningful information. It is there because Spider internally has a `runtime` class that handles everything except for network connections. It is actually a bad idea to expose it in the first place because this is more like an implementation details. Given that the deployment going through Helm and Docker Compose won't expose this, I don't see the point of removing them either.
@LinZhihao-723 What's your opinion on this?

tools/deployment/spider-compose/compose.yaml
20001020ycx: @LinZhihao-723 What is your user experience, I suggest we should have default embedded because I could be only testing a specific set of configs while leaving others to default.

LinZhihao-723: iirc this connection pool size is for gRCP client to parallelize concurrent requests. I think in practice 4 to 8 should be sufficient according to Jack's metalog testing. This should not be a user-oriented option tbh, so my suggestion is:
- Provide a default value of 8, which should be sufficient for most cases.
- We should document what this value is doing in our formal documentation.

tools/deployment/spider-compose/.env.example
LinZhihao-723: If I have to pick one, then I feel like `SPIDER_SCHEDULER_RUNTIME_SCHEDULER_CONFIG_ACTIVE_JOB_QUEUE_CAPACITY` is cleaner since it mirrors the config hierarchy explicitly.
However, I think for the scheduler this is a bit not future-proof: if we need to support different types of scheduling policies, the config schema may change according to the actual policy. It's too verbose to mirror all of that and hard to validate. Is it possible to:
- Expose `SPIDER_SCHEDULER_POLICY` for the policy.
- Expose `SPIDER_SCHEDULER_POLICY_CONFIG` as a single env var that takes a JSON string of the input config, which allows a dynamic schema?

sitaowang1998: The JSON string does not really work since the Docker Compose file cannot parse it. Unless we provide Spider support, this won't work.
Actually we have discussed the problem of multiple policy support with Junhao. He suggested that for Docker Compose side we could provide different config setup for different scheduler policy, and enable one of them depending on the environment variable. As for the actual setting, we can include all possible settings in `.env.example`, and let user comment out the ones they use.

LinZhihao-723: So they are all provided through env vars?

sitaowang1998: Yes. That make me thinking instead of mirroring the hierarchy, how about mentioning the policy in name in case there are two configs with same name for different policies. For example, `SPIDER_SCHEDULER_ROUND_ROBIN_ACTIVE_JOB_QUEUE_CAPACITY`?

LinZhihao-723: - Since we're under time pressure, I'm ok with this solution for now.
  - You may need to explain how `compose.yaml` handles `spider-scheduler-config`: the schema under `config` depends on the actual policy.
- However, I still think we should figure out a way (if possible) to compact the policy config into inline JSON/YAML. This solves the dynamic schema issue and also makes configs from different policies mutually exclusive. Might be worth creating an issue to track this.

20001020ycx: this looks good for me: SPIDER_SCHEDULER_ROUND_ROBIN_ACTIVE_JOB_QUEUE_CAPACITY, can you apply it? Also, can you make sure all env var are respecting this convention?

tools/deployment/spider-compose/configs/worker.yaml
```diff
@@ -0,0 +1,19 @@
+configs:
+  spider-worker-config:
+    content: |
+      connection_pool_size: ${SPIDER_WORKER_CONNECTION_POOL_SIZE:-4}
+      liveness:
+        scheduler_heartbeat_interval_sec: ${SPIDER_WORKER_SCHEDULER_HEARTBEAT_INTERVAL_SEC:-1}
```
20001020ycx: For example, this is another good candidate that falls in the discussion of https://github.com/y-scope/spider/pull/427/changes#r3770118608

SPIDER_WORKER_SCHEDULER_HEARTBEAT_INTERVAL_SEC should be SPIDER_WORKER_LIVENESS_SCHEDULER_HEARTBEAT_INTERVAL_SEC

20001020ycx: Pretty sure there's more, can you do a pass for all env var?

taskfiles/docker.yaml
```diff
@@ -21,3 +22,19 @@ tasks:
...
     deps: [":init"]
     cmd: "'{{.G_BUILD_SCRIPT}}' worker"
+
+  compose:validate:
+    dir: "{{.G_SPIDER_COMPOSE_DIR}}"
+    cmd: "docker compose --file compose.yaml config --quiet"
+
+  compose:up:
+    dir: "{{.G_SPIDER_COMPOSE_DIR}}"
+    cmd: "docker compose --file compose.yaml up --detach --wait"
```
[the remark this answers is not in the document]
20001020ycx: @sitaowang1998 sounds reasonable, can you confirm

tools/deployment/spider-compose/compose.yaml
```diff
@@ -0,0 +1,143 @@
...
+        "bash", "-c", "</dev/tcp/127.0.0.1/${SPIDER_SCHEDULER_PORT:-50052}",
+      ]
+
+  spider-worker:
+    <<: *service_defaults
+    image: >-
+      ${SPIDER_WORKER_IMAGE_REF:-ghcr.io/y-scope/spider/worker:${SPIDER_IMAGE_TAG:-main}}
+    networks:
+      - "default"
+      - "spider-internal"
```
20001020ycx, line 117: why we add default to spider-worker and spider-storage, is it for communication with clp package or any other consumer we may have? Just for my own curiosity.

tools/deployment/spider-compose/compose.yaml
```diff
@@ -0,0 +1,143 @@
+name: "spider"
+
+include:
```
20001020ycx, line 3: Neat! Just to confirm, oci can publish these and we can download these at the clp package side right? I think if this is true, it makes the integration quite easy as it naturally supports interpolation. 

Good job in making the interpolation possible, thanks!

20001020ycx (commented):
code changes LGTM, just some nits about environmental variable renaming and some high-level question I would like confirm.

tools/deployment/spider-compose/compose.yaml
sitaowang1998: Yes. This is for external communication.

tools/deployment/spider-compose/compose.yaml
sitaowang1998: Yes. I am working on that.

20001020ycx (dismissed):
LGTM, but https://github.com/y-scope/spider/pull/427#discussion_r3739057462 and https://github.com/y-scope/spider/pull/427#discussion_r3730466447 are still unresolved, can you take a look to see if they make sense?

tools/deployment/spider-compose/.env.example
20001020ycx: Nvm, as discussed offline, this is the right place for the WORKER image ref

taskfiles/docker.yaml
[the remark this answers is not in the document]
sitaowang1998: As discussed offline, add a 5 minutes timeout.

20001020ycx (approved):

tools/deployment/spider-compose/.env.example
sitaowang1998: As discussed offline, we should preserve the current layout.

As for whether we need extra comment on why we need this override, I don't think it necessary. Currently, we only have internal CLP user. For external user, we should have a user doc on how to deploy Spider, which would explain everything clearly.

LinZhihao-723 (approved):
Approved on behave of code owner.

PR #430, opened 2026-08-06
fix(spider-execution-manager): Send `prev_assignment` at most once per `next_task` call (fixes #429).

### LINKED ISSUE #429: spider-execution-manager: `prev_assignment` is re-sent on every `next_task` long-poll iteration.
### Bug

`GrpcSchedulerClient::next_task` long-polls in an internal loop, re-sending the same `prev_assignment` on every iteration (`components/spider-execution-manager/src/client/grpc/scheduler.rs`):

```rust
loop {
    ...next_task(scheduler::NextTaskRequest {
        prev_assignment: prev_assignment.map(Into::into),   // same value every iteration
        ...
```

The scheduler completes an assignment by removing it from its in-memory registry, so only the first iteration succeeds. Every later iteration asks the scheduler to complete a record that it has already dropped, which logs a warning at `components/spider-scheduler/src/service.rs:194`:

```
WARN Failed to complete a previously consumed assignment. Skipping.
     assignment_id=0 from_em_id=2 error="task assignment 0 not found for execution manager 2"
```

An idle execution manager stays in that loop, so the warning repeats once per poll until the execution manager receives work.

**Expected.** Send `prev_assignment` at most once per `next_task` call, and send `None` on every later iteration.

**Impact.** Log noise only, and jobs still succeed. The polls are sent whether or not the field is stale, so the defect adds no RPC. The registry is an in-memory `HashMap`, so there is no storage or database load either.

Measured 3 warnings/sec with 4 idle execution managers, steady across the sampling window. Each execution manager polls independently, so the rate should track execution manager count rather than tasks run; that follows from the loop above and was not measured at a second execution manager count.

Docker's default `json-file` driver does not rotate, and one session wrote 151 MB of these warnings. They also mask genuine ones.

### Spider version

`2ce711a` (main), built from source.

### Environment

1 storage / 1 scheduler / 4 execution managers, via the Docker Compose deployment in #427. MariaDB 10.11.16, Docker 29.6.1, Linux aarch64 (WSL2).

### Reproduction steps

1. Deploy a scheduler, a storage service, and at least one execution manager.
2. Submit any job and let it run to completion.
3. `docker logs <scheduler> 2>&1 | grep 'previously consumed assignment'` — the warnings continue at a steady rate for as long as the execution managers sit idle.

jackluo923:
# Description

Fixes #429.

`GrpcSchedulerClient::next_task` re-sent `prev_assignment` on every iteration of its long-poll loop, so the scheduler logged a warning for each iteration after the first. This calls `Option::take` on the field, which sends it on the first iteration and `None` on every later one. #429 carries the mechanism and the log output.

## Not addressed here

When the first RPC fails at the transport layer, the completion is lost and the assignment stays in the scheduler's registry until the execution manager's liveness expires. That predates this change and is worth handling separately.

# Checklist

# Validation performed

**Cluster reproduction.** A cluster of 1 storage, 1 scheduler, and 4 execution managers, backed by MariaDB 10.11.16, emitted a steady 3 warnings/sec once every job had reached `Succeeded`. Details in #429.

**Suites.** `cargo test -p spider-execution-manager` and `task lint:check-rust` pass on this branch.


components/spider-execution-manager/src/client/grpc/scheduler.rs
```diff
@@ -147,8 +149,144 @@ fn to_protocol_error(error: impl std::fmt::Display) -> SchedulerError {
...
+            observed[0],
+            Some(scheduler::TaskAssignmentRecord { id: 7, from: 4 })
+        );
+        assert!(
+            observed[1..].iter().all(Option::is_none),
+            "later poll iterations resent the assignment: {observed:?}"
+        );
+        Ok(())
+    }
+
```
LinZhihao-723: Prefer not to add this unit test since it adds too much unnecessary scaffolding to test this simple behavior. This increases the lines of code changes needed when we need to make any change to the scheduler gRPC trait.

LinZhihao-723 (commented):
The fix is correct. Let's remove the unit tests as explained in my comment.

LinZhihao-723 (approved):

PR #433, opened 2026-08-07
refactor(spider-storage)!: Remove the redundant `_config` suffix from `RuntimeConfig`'s fields.

LinZhihao-723:
# Description

Every field in spider-storage's `RuntimeConfig` carries a `_config` suffix. The suffix is redundant: the struct is already named `RuntimeConfig`, each field's type is already a `*Config`, and the fields are reached as `config.db_config`, `config.ready_queue_config`, and so on. The suffix also leaks into the YAML configuration schema, making the keys under `runtime:` needlessly verbose.

This PR drops the suffix from all four fields:

| Before | After |
| --- | --- |
| `db_config` | `db` |
| `ready_queue_config` | `ready_queue` |
| `task_instance_pool_config` | `task_instance_pool` |
| `job_cache_gc_config` | `job_cache_gc` |

Since these fields are deserialized from `storage.yaml`, the configuration keys change accordingly:

```yaml
runtime:
  db:
    host: "spider-database"
    max_connections: 64
    name: "spider-db"
    port: 3306
  ready_queue:
    cleanup_capacity: 256
    commit_capacity: 256
    task_capacity: 1048576
```

The Helm chart is updated in lockstep: `templates/configmap.yaml` renders `db:` instead of `db_config:`, and `values.yaml` exposes `spiderConfig.storage.runtime.ready_queue` instead of `ready_queue_config`. The chart version is bumped to satisfy the chart linting workflow.

Existing storage configuration files must rename these four keys. This applies to custom Helm values as well: the chart passes `spiderConfig.storage.runtime` through with `toYaml`, so a stale `ready_queue_config` (or any other old key) in a user's overrides lands in the rendered config and is rejected by Serde when the storage server starts.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

- **Breaking Changes**
  - Updated storage runtime configuration names for database, ready queue, task instance pool, and job cache settings.
  - Deployment configurations must use the updated key names to ensure runtime settings continue to load correctly.

- **Deployment**
  - Updated the Helm chart version to 0.1.10.
  - Aligned Helm defaults and generated configuration with the revised runtime setting names.

- **Bug Fixes**
  - Updated runtime recovery configuration to remain compatible with the renamed settings.

### DIFF
--- a/components/spider-storage/src/state/runtime.rs
+++ b/components/spider-storage/src/state/runtime.rs
@@ -29,13 +29,13 @@ use crate::task_instance_pool::create_task_instance_pool;
 /// Runtime configuration for the storage service.
 #[derive(Clone, Debug, Deserialize)]
 pub struct RuntimeConfig {
-    pub db_config: DatabaseConfig,
+    pub db: DatabaseConfig,
     #[serde(default)]
-    pub ready_queue_config: ReadyQueueConfig,
+    pub ready_queue: ReadyQueueConfig,
     #[serde(default)]
-    pub task_instance_pool_config: TaskInstancePoolConfig,
+    pub task_instance_pool: TaskInstancePoolConfig,
     #[serde(default)]
-    pub job_cache_gc_config: JobCacheGcConfig,
+    pub job_cache_gc: JobCacheGcConfig,
 }
 
 /// Runtime state for the storage service.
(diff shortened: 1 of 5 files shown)


sitaowang1998 (commented):
Instead of changing the variable name, how about changing the serialization name?

LinZhihao-723:
> Instead of changing the variable name, how about changing the serialization name?

By looking at the diff I don't really see a benefit of doing that.

sitaowang1998 (approved):

PR #440, opened 2026-08-14
feat(docker-compose): Add Docker Compose stack that mounts a local directory.

### LINKED ISSUE #431: Add local Docker Compose testing support with full image references
## Summary

Add full image-reference configuration and local Docker Compose testing support for Spider.

## Rationale

The current Compose deployment supports `SPIDER_WORKER_CONTAINER_IMAGE_REF` for a custom worker image. Local testing also requires storage and scheduler images that are built in the local Docker daemon. Full image references provide a direct and consistent configuration path for these images.

## Required changes

- Add storage and scheduler image-reference variables. They must behave consistently with `SPIDER_WORKER_CONTAINER_IMAGE_REF`.
- Add a task that depends on the existing Docker image build task, configures the local image references, and starts the Compose deployment.
- Add clearly named local-testing teardown and cleanup tasks. These can be aliases for the existing Compose teardown and cleanup tasks when the behaviour is identical.
- Document the supported local-testing workflow, including how to use manually compiled TDL packages with `compose.override.yaml` and how to retrieve job results through the Rust SDK.

## Affected areas

- `tools/deployment/spider-compose/compose.yaml`
- `tools/deployment/spider-compose/.env.example`
- `taskfiles/docker.yaml`
- Docker Compose local-testing documentation and the Huntsman E2E instructions

## Acceptance criteria

- Users can configure complete image references for worker, storage, and scheduler services.
- A documented task builds the local images and starts the Compose deployment without registry pulls.
- Users can tear down and clean up a locally started deployment through documented tasks.
- The documentation describes the TDL package mount workflow through `compose.override.yaml` and result retrieval through the Rust SDK.

## Backlinks

- PR: https://github.com/y-scope/spider/pull/427
- Request comment: https://github.com/y-scope/spider/pull/427#issuecomment-5220867156
- Requested by: @sitaowang1998

sitaowang1998:
# Description

This PR adds support to start, stop and clean up Docker Compose cluster that mounts a local directory so the locally built libraries are available to the task executors. This PR:
- Adds a `compose.local.yaml` that adds a volume mount for `spider-worker` services.
- Adds tasks that starts a Docker Compose cluster with a special project name and with directory mount.
- Adds tasks that stop and cleans up the Docker Compose cluster started by previously described task.
- Adds image ref override for storage and scheduler, fixes #431.
- Adds a task that builds rust, builds images and starts the Docker Compose cluster. This is a single command that a developer can use to test the change in a Docker Compose cluster.

# Checklist

  breaking change.

# Validation performed

* [x] New tasks start, stop and cleans up a cluster, and the cluster has access to the expected libraries.
* [ ] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Added local Docker Compose commands to start, stop, and clean up Spider services.
  * Local deployments now wait for services to become ready before completing startup.
  * Added support for mounting a local worker library into the Spider worker service.
  * Added optional image overrides for storage and scheduler services, while retaining existing defaults.
  * Added configuration examples for local worker library paths and service image references.

### DIFF
--- a/taskfiles/docker.yaml
+++ b/taskfiles/docker.yaml
@@ -39,6 +39,23 @@ tasks:
     dir: "{{.G_SPIDER_COMPOSE_DIR}}"
     cmd: "docker compose --file compose.yaml down --volumes --remove-orphans"
 
+  # Builds Spider and its images, then starts a local deployment.
+  #
+  # @param {string} SPIDER_WORKER_LIBRARY_DIR The host directory containing the libraries.
+  compose:local:build-and-up:
+    cmds:
+      - task: ":build:rust"
+      - task: "build"
+      - task: "compose:local:up"
+        vars:
+          SPIDER_STORAGE_IMAGE_REF:
+            sh: "sed -n '1p' '{{.G_BUILD_DIR}}/spider-storage-image.id'"
+          SPIDER_SCHEDULER_IMAGE_REF:
+            sh: "sed -n '1p' '{{.G_BUILD_DIR}}/spider-scheduler-image.id'"
+          SPIDER_WORKER_IMAGE_REF:
+            sh: "sed -n '1p' '{{.G_BUILD_DIR}}/spider-worker-image.id'"
+          SPIDER_WORKER_LIBRARY_DIR: "{{.SPIDER_WORKER_LIBRARY_DIR}}"
+
   # Starts a local Spider deployment with libraries mounted into every worker.
   #
   # @param {string} SPIDER_STORAGE_IMAGE_REF The storage image reference.
--- a/taskfiles/docker.yaml
+++ b/taskfiles/docker.yaml
@@ -52,7 +52,8 @@ tasks:
     dir: "{{.G_SPIDER_COMPOSE_DIR}}"
     cmd: "docker compose --file compose.yaml down --volumes --remove-orphans"
 
-  # Builds Spider and its images, then starts a local deployment.
+  # Builds Spider's Rust binaries and container images, then starts a local Docker Compose
+  # deployment that uses the freshly built images.
   #
   # @param {string} SPIDER_WORKER_LIBRARY_DIR The host directory containing the libraries.
   compose:local:build-and-up:
@@ -69,7 +70,11 @@ tasks:
             sh: "sed -n '1p' '{{.G_BUILD_DIR}}/spider-worker-image.id'"
           SPIDER_WORKER_LIBRARY_DIR: "{{.SPIDER_WORKER_LIBRARY_DIR}}"
 
-  # Starts a local Spider deployment with libraries mounted into every worker.
+  # Starts a local Docker Compose deployment with the library directory mounted
+  # into every worker.
+  #
+  # NOTE: Users can override image refs for each service; when unset, each defaults to
+  # `ghcr.io/y-scope/spider/<service>:${SPIDER_IMAGE_TAG:-main}`.
   #
   # @param {string} SPIDER_STORAGE_IMAGE_REF The storage image reference.
   # @param {string} SPIDER_SCHEDULER_IMAGE_REF The scheduler image reference.
--- a/taskfiles/docker.yaml
+++ b/taskfiles/docker.yaml
@@ -41,9 +41,15 @@ tasks:
 
   # Starts a local Spider deployment with libraries mounted into every worker.
   #
+  # @param {string} SPIDER_STORAGE_IMAGE_REF The storage image reference.
+  # @param {string} SPIDER_SCHEDULER_IMAGE_REF The scheduler image reference.
+  # @param {string} SPIDER_WORKER_IMAGE_REF The worker image reference.
   # @param {string} SPIDER_WORKER_LIBRARY_DIR The host directory containing the libraries.
   compose:local:up:
     env:
+      SPIDER_STORAGE_IMAGE_REF: "{{.SPIDER_STORAGE_IMAGE_REF}}"
+      SPIDER_SCHEDULER_IMAGE_REF: "{{.SPIDER_SCHEDULER_IMAGE_REF}}"
+      SPIDER_WORKER_IMAGE_REF: "{{.SPIDER_WORKER_IMAGE_REF}}"
       SPIDER_WORKER_LIBRARY_DIR: >-
         {{- if .SPIDER_WORKER_LIBRARY_DIR}}
           {{- if osIsAbs .SPIDER_WORKER_LIBRARY_DIR}}
--- a/taskfiles/docker.yaml
+++ b/taskfiles/docker.yaml
@@ -38,3 +38,26 @@ tasks:
   compose:clean:
     dir: "{{.G_SPIDER_COMPOSE_DIR}}"
     cmd: "docker compose --file compose.yaml down --volumes --remove-orphans"
+
+  # Starts a local Spider deployment with libraries mounted into every worker.
+  #
+  # @param {string} SPIDER_WORKER_LIBRARY_DIR The host directory containing the libraries.
+  compose:local:up:
+    env:
+      SPIDER_WORKER_LIBRARY_DIR: "{{.SPIDER_WORKER_LIBRARY_DIR}}"
+    requires:
+      vars: ["SPIDER_WORKER_LIBRARY_DIR"]
+    dir: "{{.G_SPIDER_COMPOSE_DIR}}"
+    cmd: >-
+      docker compose --project-name spider-local --file compose.yaml
+      --file compose.local.yaml up --detach --wait --wait-timeout 300
+
+  compose:local:down:
+    dir: "{{.G_SPIDER_COMPOSE_DIR}}"
+    cmd: "docker compose --project-name spider-local --file compose.yaml down"
+
+  compose:local:clean:
+    dir: "{{.G_SPIDER_COMPOSE_DIR}}"
+    cmd: >-
+      docker compose --project-name spider-local --file compose.yaml down
+      --volumes --remove-orphans
--- a/taskfiles/docker.yaml
+++ b/taskfiles/docker.yaml
@@ -52,33 +52,13 @@ tasks:
     dir: "{{.G_SPIDER_COMPOSE_DIR}}"
     cmd: "docker compose --file compose.yaml down --volumes --remove-orphans"
 
-  # Builds Spider's Rust binaries and container images, then starts a local Docker Compose
-  # deployment that uses the freshly built images.
-  #
-  # @param {string} SPIDER_WORKER_LIBRARY_DIR The host directory containing the libraries.
-  compose:local:build-and-up:
-    env:
-    cmds:
-      - task: ":build:rust"
-      - task: ":docker:build"
-      - task: "compose:local:up"
-        vars:
-          SPIDER_PULL_POLICY: "never"
-          SPIDER_STORAGE_IMAGE_REF:
-            sh: "sed -n '1p' '{{.G_BUILD_DIR}}/spider-storage-image.id'"
-          SPIDER_SCHEDULER_IMAGE_REF:
-            sh: "sed -n '1p' '{{.G_BUILD_DIR}}/spider-scheduler-image.id'"
-          SPIDER_WORKER_IMAGE_REF:
-            sh: "sed -n '1p' '{{.G_BUILD_DIR}}/spider-worker-image.id'"
-          SPIDER_WORKER_LIBRARY_DIR: "{{.SPIDER_WORKER_LIBRARY_DIR}}"
-
   # Starts a local Docker Compose deployment with the library directory mounted
   # into every worker.
   #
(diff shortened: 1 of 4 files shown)


taskfiles/docker.yaml
```diff
@@ -38,3 +38,26 @@ tasks:
   compose:clean:
     dir: "{{.G_SPIDER_COMPOSE_DIR}}"
     cmd: "docker compose --file compose.yaml down --volumes --remove-orphans"
+
+  # Starts a local Spider deployment with libraries mounted into every worker.
+  #
+  # @param {string} SPIDER_WORKER_LIBRARY_DIR The host directory containing the libraries.
```
20001020ycx:
```suggestion
```

20001020ycx: Looks like SPIDER_WORKER_LIBRARY_DIR is an env var required for running this command, does task support argument check? if not, what is the default? if no default, then what is the failed bahaviour?

20001020ycx (commented):
Could you explain to me how to reproduce this as if you are writing a user guide?

For example, say I have a TDL package that I would like to package into the worker, what is that directory should look like, just a .so file?

Then do I just run `task docker:local:up`? 

Where do I set my env? I am assuming in .env, so you should update the .env.example. And what is the path I shall use? Absolute path or relative path, if relative path, what is the reference?

I dont have further comment regarding the code, thanks!

taskfiles/docker.yaml
20001020ycx: I understand you have `requires`, but that is holding against `SPIDER_WORKER_LIBRARY_DIR` rather than the env var of `.SPIDER_WORKER_LIBRARY_DIR` right?

taskfiles/docker.yaml
```diff
@@ -38,3 +38,26 @@ tasks:
...
+  compose:local:up:
+    env:
+      SPIDER_WORKER_LIBRARY_DIR: "{{.SPIDER_WORKER_LIBRARY_DIR}}"
+    requires:
+      vars: ["SPIDER_WORKER_LIBRARY_DIR"]
+    dir: "{{.G_SPIDER_COMPOSE_DIR}}"
+    cmd: >-
+      docker compose --project-name spider-local --file compose.yaml
+      --file compose.local.yaml up --detach --wait --wait-timeout 300
```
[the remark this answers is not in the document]
20001020ycx: @sitaowang1998 looks interesting, relate to my concern, can you take a look?

taskfiles/docker.yaml
sitaowang1998: There is no default provided because this task is supposed to run with a user-specified directory to mount. If running the task without the environment variable, it fails immediately with exit code 206 and with an error message:
```
task: Task "docker:compose:local:up" cancelled because it is missing required variables: SPIDER_WORKER_LIBRARY_DIR
```

taskfiles/docker.yaml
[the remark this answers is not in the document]
sitaowang1998: Relative path does pose a problem. However, the cleanest way to solve this is to use `absPath` template introduced in `go-task` [3.5.11](https://taskfile.dev/docs/changelog#v3-51-1-2026-05-16). However, since we still use `3.48.0` in all workflows, we have two options:
1. Leave the code as it is, and in command specify that we only accept an absolute path. Maybe open an issue to track the problem and remove the absolute path requirement once we pump the `go-task` version.
2. Use a more complex template to solve the problem:
```
SPIDER_WORKER_LIBRARY_DIR: >-
  {{- if .SPIDER_WORKER_LIBRARY_DIR}}
    {{- if osIsAbs .SPIDER_WORKER_LIBRARY_DIR}}
      {{- .SPIDER_WORKER_LIBRARY_DIR}}
    {{- else}}
      {{- .ROOT_DIR}}/{{.SPIDER_WORKER_LIBRARY_DIR}}
    {{- end}}
  {{- end}}
```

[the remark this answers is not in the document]
LinZhihao-723: Looks like this is docker compose side issue, then is it a problem for users as well? If so, it's probably worth taking option 2, as users may provide a relative path as well.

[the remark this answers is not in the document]
sitaowang1998: Most likely the only user is ourselves. The e2e test will use this task to start the cluster. However, if we don't resolve the relative path to absolute path here, we will probably need to do it in the e2e test task later, so I think option 2 is better. I'll still open an issue in case we bump `go-task` version later.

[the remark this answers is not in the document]
sitaowang1998: @coderabbit Please open an issue to track this.

taskfiles/docker.yaml
sitaowang1998: With the update to the task, this changes a bit. The task still fails, but during `docker compose up` executes, and with fail with error code 201 and different error messages:
```
error while interpolating services.spider-worker.volumes.[].source: required variable SPIDER_WORKER_LIBRARY_DIR is missing a value
task: Failed to run task "docker:compose:local:up": exit status 1
```

taskfiles/docker.yaml
```diff
@@ -51,3 +51,54 @@ tasks:
...
+        vars:
+          SPIDER_STORAGE_IMAGE_REF:
+            sh: "sed -n '1p' '{{.G_BUILD_DIR}}/spider-storage-image.id'"
+          SPIDER_SCHEDULER_IMAGE_REF:
+            sh: "sed -n '1p' '{{.G_BUILD_DIR}}/spider-scheduler-image.id'"
+          SPIDER_WORKER_IMAGE_REF:
+            sh: "sed -n '1p' '{{.G_BUILD_DIR}}/spider-worker-image.id'"
+          SPIDER_WORKER_LIBRARY_DIR: "{{.SPIDER_WORKER_LIBRARY_DIR}}"
+
+  # Starts a local Spider deployment with libraries mounted into every worker.
```
20001020ycx: Add image override to the docstring

taskfiles/docker.yaml
20001020ycx: clean failure result is good

taskfiles/docker.yaml
```diff
@@ -51,3 +51,54 @@ tasks:
...
     dir: "{{.G_SPIDER_COMPOSE_DIR}}"
     cmd: "docker compose --file compose.yaml down --volumes --remove-orphans"
+
+  # Builds Spider and its images, then starts a local deployment.
+  #
+  # @param {string} SPIDER_WORKER_LIBRARY_DIR The host directory containing the libraries.
+  compose:local:build-and-up:
+    cmds:
+      - task: ":build:rust"
+      - task: "build"
```
20001020ycx: how about naming it to docker:build for more explicit use?

taskfiles/docker.yaml
```diff
@@ -51,3 +51,54 @@ tasks:
   compose:clean:
     dir: "{{.G_SPIDER_COMPOSE_DIR}}"
     cmd: "docker compose --file compose.yaml down --volumes --remove-orphans"
+
+  # Builds Spider and its images, then starts a local deployment.
+  #
+  # @param {string} SPIDER_WORKER_LIBRARY_DIR The host directory containing the libraries.
+  compose:local:build-and-up:
```
20001020ycx: since this is in docker.yaml, user would use it as `task docker:compose:local:build-and-up"

this naming is a bit weird, how about `task docker:compose:up". This distinguish the below `task docker:compose:local:up`.

But I am not good at naming, so feel free to suggest any other style.

20001020ycx (commented):
I dont have other comment regarding the code. However, I would appreciate if there's some documentation to be done in how would a user use the new task implemented in this PR. I think put it into the PR description makes sense, but feel free to use other means if you find it to be more usable.

taskfiles/docker.yaml
sitaowang1998: `docker:compose:up` is already used.

taskfiles/docker.yaml
sitaowang1998: They are already listed as parameters.

taskfiles/docker.yaml
```diff
@@ -51,3 +51,54 @@ tasks:
   compose:clean:
     dir: "{{.G_SPIDER_COMPOSE_DIR}}"
     cmd: "docker compose --file compose.yaml down --volumes --remove-orphans"
+
+  # Builds Spider and its images, then starts a local deployment.
```
20001020ycx:
```suggestion
  # Builds Spider's Rust binaries and container images, then starts a local Docker Compose
  # deployment that uses the freshly built images.
```

taskfiles/docker.yaml
```diff
@@ -51,3 +51,54 @@ tasks:
...
+        vars:
+          SPIDER_STORAGE_IMAGE_REF:
+            sh: "sed -n '1p' '{{.G_BUILD_DIR}}/spider-storage-image.id'"
+          SPIDER_SCHEDULER_IMAGE_REF:
+            sh: "sed -n '1p' '{{.G_BUILD_DIR}}/spider-scheduler-image.id'"
+          SPIDER_WORKER_IMAGE_REF:
+            sh: "sed -n '1p' '{{.G_BUILD_DIR}}/spider-worker-image.id'"
+          SPIDER_WORKER_LIBRARY_DIR: "{{.SPIDER_WORKER_LIBRARY_DIR}}"
+
+  # Starts a local Spider deployment with libraries mounted into every worker.
```
20001020ycx:
```suggestion
  # Starts a local Docker Compose deployment with the library directory mounted
  # into every worker.
  #
  # NOTE: Users can override image refs for each service; when unset, each defaults to
  # `ghcr.io/y-scope/spider/<service>:${SPIDER_IMAGE_TAG:-main}`.
```

taskfiles/docker.yaml
20001020ycx: see my latest review comment, that's what I mean by this

taskfiles/docker.yaml
20001020ycx: you are right, how about `compose:dev`? I imagine this task would sit on the hot path for Spider developer.

20001020ycx (changes requested):
I also found that `SPIDER_WORKER_LIBRARY_DIR=/tmp task docker:compose:local:build-and-up` does not work, this is the error message:

> task: [docker:compose:local:up] docker compose --project-name spider-local --file compose.yaml --file compose.local.yaml up --detach --wait --wait-timeout 300
[+] up 4/4
 ! Image sha256:f3af79a7622e1ece09c611880f470f91e518daa3a88965771145483456e199c1 Interrupted                                                                                                                                                                              0.3s
 ! Image sha256:1ad3e5e0563f4f29f843563e8a1da3f18003fa057d8573fbfe0eac4e82054724 Interrupted                                                                                                                                                                              0.3s
 ! Image mariadb:10.11.16                                                        Interrupted                                                                                                                                                                              0.3s
 ✘ Image sha256:91cb7cd3e0318a267c7723770e11b9bc224d989310431cf4d41e2386278d54e2 Error pull access denied for sha256, repository does not exist or may require 'docker login'                                                                                             0.3s
Error response from daemon: pull access denied for sha256, repository does not exist or may require 'docker login'
task: Failed to run task "docker:compose:local:build-and-up": task: Failed to run task "docker:compose:local:up": exit status 1

taskfiles/docker.yaml
sitaowang1998: This is worse. It describes its target use case, but not what it does. In fact, I don't see the problem of the original name other than a bit long.

sitaowang1998:
> I also found that SPIDER_WORKER_LIBRARY_DIR=/tmp task docker:compose:local:build-and-up does not work

My bad. The problem is the image poll policy is by default `always`. I have a local `.env` that sets it to `never`, which creates an illusion that the task does not pull from remote. I have added the image pull policy setting in task and tested with `.env` removed.

taskfiles/docker.yaml
20001020ycx: @LinZhihao-723 what's your take? 

For context, this task is to build Spider's Rust binaries and container images, then starts a local Docker Compose deployment that uses the freshly built images.

This is the last point we need to address before merging this PR.

LinZhihao-723: Hmmm, why do we need a single task command to build and start?

sitaowang1998: Easier for developers to test the code changes. Otherwise they have to run these manually.

LinZhihao-723: I prefer to keep them separate: running build and up doesn't make a fundamental difference compared to run build+up in one command. This also leads to better error reporting in a CI flow: it's now clear which step failed, either build or up; no need to read the task log.

sitaowang1998: As discussed offline, remove the task.

20001020ycx (approved):
We have discussed offline that such docker compose wont sit on the hot path for developers dev use but for CI who runs e2e test. Therefore, we delegate the env passing to the CI jobs after our discussion.

The change now LGTM, thanks.

PR #443, opened 2026-08-17
refactor(spider-scheduler): Move dispatch queue ownership into the scheduler core.

LinZhihao-723:
# Description

The scheduler runtime currently creates the dispatch queue, splits it into a `(writer, reader)` pair, hands the writer to the core as a parameter of `SchedulerCore::run`, and keeps the reader for the execution-manager-facing service. That arrangement assumes the dispatch queue is **one channel**, and two consequences follow from the assumption:

* **The runtime must know the sink's concrete type at compile time.** `create_runtime` calls `make_core::<SchedulerStorageClientType, DispatchQueueWriter>()`, so every core implementation is forced to share one dispatch queue type.
* **A core cannot own a dispatch structure that is not a single channel.** A resource-group-aware core's dispatch state is a set of per-group queues created on demand by either side, a hint channel, and a table co-owned with the service. There is no way to build that outside the core and pass it in as a sink.

This PR inverts the ownership: the core constructs and owns whatever dispatch structure its algorithm needs, and exposes only a read handle to the service. The runtime builds the core first, takes the handle from it, and then builds the service.

```rust
let core = scheduler_config.make_core::<SchedulerStorageClientType>();
let dispatch_queue_handle = core.get_dispatch_queue_handle();
let service = SchedulerServiceState::new(dispatch_queue_handle, registry, scheduler_id);
```

## API changes

| Before | After |
| --- | --- |
| `SchedulerCore::Sink` (associated type) | removed |
| `SchedulerCore::run(self, storage_client, sink, ..)` | `SchedulerCore::run(self, storage_client, ..)` |
| — | `SchedulerCore::get_dispatch_queue_handle(&self) -> SharedDispatchQueueHandle` |
| `trait DispatchQueueSource: Send + Sync + Clone` | `trait DispatchQueueHandle: Send + Sync` |
| — | `pub type SharedDispatchQueueHandle = Arc<dyn DispatchQueueHandle>` |
| `trait DispatchQueueSink` | removed; `enqueue`, `bump_session_id` and `size` become inherent methods on `DispatchQueueWriter` |
| `SchedulerServiceState<DispatchQueueSourceType>` | `SchedulerServiceState` |
| `GrpcSchedulerService<DispatchQueueSourceType>` | `GrpcSchedulerService` |
| `SchedulerConfig::make_core<SchedulerStorageClientType, DispatchQueueSinkType>()` | `SchedulerConfig::make_core<SchedulerStorageClientType>()` |
| `SchedulerConfig::dispatch_queue_capacity()` | removed |
| `create_runtime(..) -> (Runtime, SchedulerServiceState<DispatchQueueReader>, CancellationToken)` | `create_runtime(..) -> (Runtime, SchedulerServiceState, CancellationToken)` |

Behaviour is unchanged. The queue is still created at `SessionId::default()` with the capacity from the core's own config, which is exactly what the runtime did before.

## Why `Arc` and not `Box`

`Clone` implies `Sized`, so the trait could not be made into a trait object while `Clone` was a supertrait. Dropping it is what makes `dyn DispatchQueueHandle` possible; `#[async_trait]` boxes the returned future, so `dequeue` remains callable through the trait object.

The handle is an `Arc` with exactly **one** level of indirection. `DispatchQueueReader` previously held an `Arc<DispatchQueueReaderInner>`, so returning `Arc<dyn DispatchQueueHandle>` unchanged would have wrapped an `Arc` in an `Arc` — three pointer hops from `SchedulerServiceState` to the channel where there had been two. `DispatchQueueReaderInner` is therefore flattened into `DispatchQueueReader`, whose two remaining fields (`session_id` and `assignment_receiver`) are already cheap-clone handles.

`Box` was considered and rejected: it removes the double indirection, but hands the service a source with no guarantee of cheap cloning. Nothing clones it today, but that is a property of the current service rather than of the contract, and the contract is what a second core inherits.

## Why `DispatchQueueSink` is deleted rather than kept

Before this change the trait was load-bearing: `RoundRobinCore` was generic over `DispatchQueueSinkType: DispatchQueueSink`, so the bound did real work. Once the core owns a concrete `DispatchQueueWriter`, the only holder is that concrete type, every call resolves statically, and the trait's sole remaining effect is forcing an import to bring the methods into scope.

The surviving trait stays a trait, and the asymmetry is the point: it is the seam the service depends on, and a second core will implement it differently. `DispatchQueueSink` had no second implementor coming.

With `DispatchQueueSink` gone, `DispatchQueueSource` named one half of a pair that no longer existed, so it is renamed to `DispatchQueueHandle` — the sole external access point to a core's dispatch queue, which is expected to grow further caller-to-core methods beyond `dequeue`. Its `Arc` alias follows the `Shared<Thing>` convention already used by `spider-storage`'s `SharedRw` and `SharedJobControlBlock`.

One consequence worth flagging for review: `size` gained `#[must_use]`. As a trait method it was exempt from `clippy::must_use_candidate`; as a public inherent method it is not, and the pedantic lint group is deny-level in this workspace.

These were considered and left alone, so please read them as decisions rather than as misses:

* **`DispatchQueueHandle::dequeue` keeps its signature.** A resource-group-aware core needs to distinguish a pinned execution manager from a general one, which `dequeue(wait_time)` cannot express. The minimal shape is `dequeue(&self, resource_group_id: Option<ResourceGroupId>, wait_time: Duration)`, with the round-robin implementation ignoring the argument. Deferred to a follow-up PR to keep this diff reviewable.
* **`SchedulerCore: Send` is retained**, and `create_runtime` still spawns `core.run(..)` with `tokio::spawn`. A core whose scheduling state is not `Send` cannot be spawned that way; addressing it is a separate change and does not affect anything here.
* **The round-robin scheduling policy.** No admission, ordering or retirement behaviour is touched. The only change to `RoundRobin` is that it holds a concrete `DispatchQueueWriter` field instead of a generic sink, renamed to `dispatch_queue_writer` to match.
* **`create_dispatch_queue` remains public.** The round-robin core calls it from `make_core`, and the white-box tests call it directly.

## Notes for reviewers

The round-robin tests already passed the real `DispatchQueueWriter` rather than a mock, so the test changes are mechanical: the helpers stop threading a `sink` parameter and take the core's reader for assertions instead.

`RoundRobin::run` destructures `*self` with `..`, which drops the core's dispatch queue reader. That is safe because `DispatchQueueWriterInner` holds its own clone of the `async_channel::Receiver`, so the channel does not close even if `get_dispatch_queue_handle` is never called.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **Refactor**
  * Consolidated dispatch-queue ownership within the scheduler for simpler integration.
  * Introduced a single dispatch-queue handle for monitoring and retrieving scheduled work.
  * Simplified scheduler, runtime, and service interfaces.
  * Preserved queue behaviour, including back-pressure, session validation, task assignment, and rescheduling.
* **Tests**
  * Updated coverage for queue draining, rescheduling, deduplication, session updates, and job finalization.

### DIFF
--- a/components/spider-scheduler/src/dispatch_queue.rs
+++ b/components/spider-scheduler/src/dispatch_queue.rs
@@ -11,46 +11,9 @@ use tokio::sync::RwLock;
 use crate::error::SchedulerError;
 use crate::types::TaskAssignment;
 
-/// The writer side of the dispatching queue used by the scheduler core.
-#[async_trait]
-pub trait DispatchQueueSink: Send + Sync + Clone {
-    /// Enqueues a task assignment for execution managers to consume.
-    ///
-    /// # Parameters
-    ///
-    /// * `assignment` - The task assignment to enqueue.
-    ///
-    /// # Errors
-    ///
-    /// Returns an error if:
-    ///
-    /// * [`SchedulerError::DispatchQueueClosed`] if the dispatching queue is closed.
-    async fn enqueue(&self, assignment: TaskAssignment) -> Result<(), SchedulerError>;
-
-    /// Bumps the session ID and invalidates all queued task assignments.
-    ///
-    /// # Parameters
-    ///
-    /// * `new_session_id` - The new session ID. Must be greater than the current session ID.
-    ///
-    /// # Errors
-    ///
-    /// Returns an error if:
-    ///
-    /// * [`SchedulerError::DispatchQueueClosed`] if the dispatching queue is closed.
-    /// * [`SchedulerError::InvalidSessionId`] if the new session ID is not greater than the current
-    ///   session ID.
-    async fn bump_session_id(&self, new_session_id: SessionId) -> Result<(), SchedulerError>;
-
-    /// # Returns
-    ///
-    /// The current size of the dispatch queue.
-    fn size(&self) -> usize;
-}
-
 /// The reader side of the dispatching queue, drained by the execution-manager-facing service.
 #[async_trait]
-pub trait DispatchQueueSource: Send + Sync + Clone {
+pub trait DispatchQueueSource: Send + Sync {
     /// Dequeues the next task assignment for an execution manager to execute.
     ///
     /// # Parameters
@@ -70,8 +33,7 @@ pub trait DispatchQueueSource: Send + Sync + Clone {
     async fn dequeue(&self, wait_time: Duration) -> Result<Option<TaskAssignment>, SchedulerError>;
 }
 
-/// A cloneable writer handle for the dispatching queue, implementing [`DispatchQueueSink`] using
-/// an async channel.
+/// A cloneable writer handle for the dispatching queue, backed by an async channel.
 ///
 /// # NOTE
 ///
@@ -83,9 +45,15 @@ pub struct DispatchQueueWriter {
     inner: Arc<DispatchQueueWriterInner>,
 }
 
-#[async_trait]
-impl DispatchQueueSink for DispatchQueueWriter {
-    async fn enqueue(&self, assignment: TaskAssignment) -> Result<(), SchedulerError> {
+impl DispatchQueueWriter {
+    /// Enqueues a task assignment for execution managers to consume.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * [`SchedulerError::DispatchQueueClosed`] if the dispatching queue is closed.
+    pub async fn enqueue(&self, assignment: TaskAssignment) -> Result<(), SchedulerError> {
         self.inner
             .assignment_sender
             .send(assignment)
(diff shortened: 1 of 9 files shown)


components/spider-scheduler/src/dispatch_queue.rs
```diff
@@ -108,34 +90,39 @@ impl DispatchQueueSink for DispatchQueueWriter {
         Ok(())
     }
 
-    fn size(&self) -> usize {
+    /// # Returns
+    ///
+    /// The current size of the dispatch queue.
```
sitaowang1998: Do we really need this docstring?

sitaowang1998 (commented):
LGTM.

components/spider-scheduler/src/dispatch_queue.rs
LinZhihao-723: Removed. I thought the docstring is necessary since we now make the API public. But looks like clippy doesn't complain if I remove it.

sitaowang1998 (approved):

sitaowang1998:
Verified that e2e tests from #385 pass after merge.

PR #444, opened 2026-08-18
feat(spider-scheduler): Add the job registry for the resource-group-round-robin scheduler core.

LinZhihao-723:
# Description

This is the first piece of a second scheduler core, `resource_group_round_robin`, which schedules with two levels of round-robin: the outer level interleaves resource groups, the inner level interleaves active jobs within a group. This PR lands only its **job registry** — the structure holding each running job's queue of ready tasks. The core that consumes it follows in later PRs, so nothing is wired up and no behaviour changes.

## What the registry is

`JobRegistry` owns every registered job's `JobEntry` and supports two access paths:

* **By job ID**, through a `HashMap<JobId, JobKey>`, used when tasks arrive from the inbound queue and have to be routed to the job that owns them.
* **By key**, a direct arena lookup, used by the scheduling loop, which holds keys rather than looking a job up by ID on every decision.

A `JobEntry` holds the job's FIFO queue of ready tasks and a *downgrade counter* — the number of further chances the job gets to be refilled before the core retires it.

## Why a generational arena

The entries live in a `slotmap::SlotMap` keyed by a generational `JobKey`, rather than in a `HashMap` or behind reference-counted handles. The reason is specific, and it is the design decision most worth reviewing:

**A scheduling position is expected to hold a key to a job that has already been removed, and to dereference it afterwards.** That is not a bug to be defended against — it is the mechanism by which the scheduling loop *discovers* the job is gone when it is finalized.

With a plain index or a slot reused without versioning, the arena would be free to hand that slot to the next job created, and the scan would then read a live but unrelated entry: it would schedule tasks under another job's position and swap the wrong entry out of the active list — silently, with nothing to observe. A generation makes the same lookup return `None`, which is the answer the scan wanted.

Two tests pin this down: `a_removed_jobs_key_never_resolves_to_a_later_job` covers exactly the reuse hazard, and `upsert_keeps_the_scheduling_position_of_an_existing_job` covers the converse — that re-upserting a live job appends to it rather than displacing it.

## Visibility and the dead-code expectation

Every item is `pub(super)`: visible throughout `resource_group_round_robin` so the forthcoming core can use it, and no wider. This mirrors the sibling, where `round_robin/implementation.rs` uses `pub(super)` for internals and reserves `pub` for the config and core types. Nothing is re-exported from `core_impl`, so the registry does not reach `spider_scheduler`'s public API.

Because the consumer has not landed, every item reads as dead to the compiler. The module declaration therefore carries:

```rust
#[expect(dead_code, reason = "the core that consumes the registry has not landed yet")]
mod job_registry;
```

`expect` rather than `allow` deliberately: when the core lands and uses the registry, the lint stops firing, the expectation becomes unfulfilled, and the compiler emits `this lint expectation is unfulfilled` with that reason attached — which is the instruction to delete the line. An `allow` would simply outlive its purpose in silence.

These were considered and left alone, so please read them as decisions rather than as misses:

* **No configuration, no wiring, no behaviour.** `SchedulerConfig` gains no variant, `scheduler.yaml` gains no key, and no code path reaches the registry outside its own tests. The core selects a policy through `SchedulerConfig`, and this PR adds nothing there.
* **`round_robin` is untouched.** The two cores will sit side by side under `core_impl`; nothing about the existing one moves or changes.

# Checklist

  breaking change.

# Validation performed

* [x] Add unit tests to assert expected behaviors.

## Summary by CodeRabbit

* **New Features**
  * Added foundational support for resource-group-aware round-robin scheduling.
  * Added job registration and tracking, including task queues, scheduling order, removal, clearing, and stale-entry handling.
  * Added downgrade tracking and reset behaviour to support fair scheduling decisions.

* **Tests**
  * Added coverage for job registration, task ordering, queue management, removal, clearing, stale entries, and scheduling state transitions.

### DIFF
--- a/components/spider-scheduler/src/core_impl/resource_group_round_robin/job_registry.rs
+++ b/components/spider-scheduler/src/core_impl/resource_group_round_robin/job_registry.rs
@@ -1,0 +1,395 @@
+//! Registry of currently running jobs, each holding a queue of ready tasks to schedule from.
+
+use std::collections::HashMap;
+use std::collections::VecDeque;
+
+use slotmap::SlotMap;
+use spider_core::task::TaskIndex;
+use spider_core::types::id::JobId;
+
+slotmap::new_key_type! {
+    /// A generational handle to a job entry owned by a [`JobRegistry`].
+    pub(super) struct JobKey;
+}
+
+/// The result of inserting a batch of ready tasks for a job.
+pub(super) enum UpsertOutcome {
+    /// The job was already registered.
+    Exist,
+
+    /// The job was registered by this call, along with a key to access its entry.
+    New(JobKey),
+}
+
+/// The scheduling state of a single job, maintaining a FIFO queue for its ready tasks.
+pub(super) struct JobEntry {
+    job_id: JobId,
+    ready_task_queue: VecDeque<TaskIndex>,
+    downgrade_counter: u32,
+}
+
+impl JobEntry {
+    /// Factory function.
+    ///
+    /// # Returns
+    ///
+    /// A newly created entry for `job_id` holding `task_indices`.
+    fn new(job_id: JobId, task_indices: Vec<TaskIndex>) -> Self {
+        Self {
+            job_id,
+            ready_task_queue: task_indices.into(),
+            downgrade_counter: DOWNGRADE_LIVES,
+        }
+    }
+
+    /// # Returns
+    ///
+    /// The job ID of this entry.
+    pub(super) const fn job_id(&self) -> JobId {
+        self.job_id
+    }
+
+    /// Appends newly arrived ready tasks and restores the job's full downgrade budget.
+    pub(super) fn insert_tasks(&mut self, task_indices: Vec<TaskIndex>) {
+        self.ready_task_queue.extend(task_indices);
+        self.downgrade_counter = DOWNGRADE_LIVES;
+    }
+
+    /// Reads the job's next ready task from the queue without taking it.
+    ///
+    /// # Returns
+    ///
+    /// The next ready task, or [`None`] if the queue is empty.
+    pub(super) fn peek_next_task(&self) -> Option<TaskIndex> {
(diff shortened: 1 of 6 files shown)


components/spider-scheduler/src/core_impl/resource_group_round_robin/job_registry.rs
```diff
@@ -0,0 +1,395 @@
...
+
+/// The core's registry of job entries, supporting two access methods:
+///
+/// * Job ID: Lookup through the job ID-to-key mapping.
+/// * Job key: Direct lookup using the key returned when the job is created.
+#[derive(Default)]
+pub(super) struct JobRegistry {
+    entries: SlotMap<JobKey, JobEntry>,
+    id_to_key: HashMap<JobId, JobKey>,
+}
```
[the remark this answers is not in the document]
LinZhihao-723: That is fine for now. The actual non-removal access should be all through the key.

components/spider-scheduler/src/core_impl/resource_group_round_robin/job_registry.rs
```diff
@@ -0,0 +1,395 @@
...
+    /// The next ready task, or [`None`] if the queue is empty.
+    pub(super) fn pop_next_task(&mut self) -> Option<TaskIndex> {
+        self.ready_task_queue.pop_front()
+    }
+
+    /// Empties the job's ready task queue.
+    ///
+    /// # Returns
+    ///
+    /// The ready tasks the job still held.
```
sitaowang1998: `still held` is confusing.

sitaowang1998 (commented):
LGTM.

components/spider-scheduler/src/core_impl/resource_group_round_robin/job_registry.rs
LinZhihao-723: Fixed.

sitaowang1998 (approved):

PR #449, opened 2026-08-21
feat(spider-scheduler): Add the dispatch queue for the resource-group-round-robin scheduler core.

LinZhihao-723:
# Description

This is the second piece of the `resource_group_round_robin` scheduler core, following the job registry in #444. It lands the **dispatch queue subsystem**: every channel an assignment or a hint travels through on its way from the core to an execution manager. The core and the dispatch service that use it follow in later PRs, so nothing is wired up and no behaviour changes.

## What the module is

Four types, in one module:

* **`DispatchQueueRegistry`** — the shared handle. One `Arc` deep, so a clone is one pointer to one allocation holding the per-group table, a `SessionTracker` clone, and **both ends** of the broadcast queue.
* **`RgDispatchQueueReader`** — the read side of one resource group's queue. A pinned execution manager blocks on `recv_pinned`.
* **`Hint`** — what a general execution manager receives from the broadcast queue. It names a resource group; it carries no assignment.
* **`RgDispatchQueueWriter`** — the write side, which the core's scheduling unit owns. Its whole surface is `try_send` and `queue_len`.

An assignment is stored exactly once, in the queue of the resource group that owns it. Hints carry no payload, which is what makes exactly-once dispatch structural rather than protocol-enforced.

## The design decisions worth reviewing

**A hint can be spent at most once, and the type says so.** `Hint`'s field and constructor are both module-private, so the only way to hold one is to have received it from the broadcast queue — no holder of a reader can wrap one. `consume_and_try_recv` takes `self`, and `Hint` is deliberately neither `Clone` nor `Copy`, so spending the same hint twice is a compile error rather than a convention. `RgDispatchQueueReader` has no spend at all: the pinned path, which holds a reader and nothing else, has no route to a group's hint counter. That is the accident this shape exists to prevent — an earlier revision put the spend on the reader, where any holder could decrement a count it had no claim on.

Dropping a hint unspent stays possible, and is relied upon: the dispatch service inspects `session_id()` and lets a stale-session hint go without touching any counter. There is deliberately no `Drop` impl.

**`try_send` publishes the assignment and decides its hint as one operation.** Three orderings used to be obligations on the call site: the assignment must reach the queue before the queue's occupancy is sampled; the occupancy `S` must be sampled before the hint count `H`; and a hint that has been taken out must be sent rather than dropped. All three are now internal to one method, so there is one body to review instead of every caller. `try_make_hint` is private and `try_send` is its only caller.

**The registry owns both ends of the broadcast queue**, which has two consequences worth checking:

* The queue cannot close while the registry lives, so an unbounded wait on an empty queue would never return. `next_hint` therefore takes a `wait_time`, exactly as `recv_pinned` does.
* It is the one structure a session bump has to empty *explicitly* rather than by dropping something — every other channel resets itself when its senders go. So `clear` drains it as well as clearing the table, and does so after the clear, which makes the postcondition unconditional: once the table is empty, every hint the queue could still hold names a group that is already gone.

**Every queue in the subsystem is unbounded.** The admission threshold is what limits a group's occupancy, so a channel bound would be a second, redundant limit whose only possible effect is to reject a send the design's coverage proof requires to succeed.

**Groups are stamped with the session at creation.** The registry holds a `SessionTracker` clone and reads it itself, so no caller passes a session id in. The two accessors create a group on demand through one private `get_or_create`, which is the single place the stamp is applied.

## Error handling

`try_send` reuses the existing `SchedulerError::DispatchQueueClosed` — no new variant. Its docstring is widened to cover both queues, which is the only change to `error.rs`.

Both closures report the same error because no caller can act on the difference: both are fatal to the core, neither leaves anything to recover, and the broadcast one cannot arise in a running scheduler at all. An earlier revision gave the broadcast case its own variant; it named a distinction that was simultaneously unreachable and unactionable.

## A note on the file name

`core_impl/resource_group_round_robin/dispatch_queue.rs` coexists with the crate-root `src/dispatch_queue.rs`, which holds the `DispatchQueueHandle` trait. That is intended rather than an oversight: the registry will implement that trait in a later PR, which is also why `try_send` returns `SchedulerError` — the vocabulary the trait already documents for this condition.

## Visibility and the dead-code expectation

Every item is `pub(super)`: visible throughout `resource_group_round_robin` so the forthcoming core can use it, and no wider. Nothing is re-exported from `core_impl`, so none of this reaches `spider_scheduler`'s public API.

Because the consumer has not landed, the module reads as dead to the compiler. Unlike the job registry, its own tests exercise every item, so a bare `#[expect(dead_code)]` would be *unfulfilled* under `cfg(test)` and fire the very lint it suppresses. The declaration is therefore conditional:

```rust
#[cfg_attr(
    not(test),
    expect(
        dead_code,
        reason = "the core and the dispatch service that consume the queues have not landed yet"
    )
)]
mod dispatch_queue;
```

`expect` rather than `allow` for the same reason as #444: once the core uses the queues, the expectation becomes unfulfilled and the compiler prints the reason as the instruction to delete the line.

Please read these as decisions rather than as misses:

* **No `impl DispatchQueueHandle`.** The registry is shaped to implement it — `next_hint`, `get_dispatch_queue_reader` and the session tracker are the pieces it needs — but the impl, and the wrapper type pairing the registry with the reschedule queue, land in a later PR.
* **No configuration, no wiring, no behaviour.** `SchedulerConfig` gains no variant, `scheduler.yaml` gains no key, and no code path outside the module's own tests reaches it.
* **`spider-core` is untouched.** `SessionTracker` is used as it already exists; no method was added to it.
* **One manifest line.** `dashmap` moves from `[dev-dependencies]` to `[dependencies]`, now that `DashMap` appears in non-test code.

# Checklist

  breaking change.

# Validation performed

* [x] Add unit tests to assert expected behaviors, including the coverage gate (`try_send` publishes a hint only while the group's queue is uncovered), the pinned and general read paths, that `clear` both drops every group and drains the broadcast queue, that a group re-created after a session advance carries the new session, and that both closure paths surface `DispatchQueueClosed`.

## Summary by CodeRabbit

* **New Features**
  * Added resource-group scheduling support for distributing task assignments through dedicated queues.
  * Added session-aware queue handling, timeout-based retrieval, delivery hints, and queue lifecycle management.
  * Added safeguards for stale scheduling hints and cancelled or unavailable assignments.

* **Bug Fixes**
  * Improved handling and reporting when assignment queues are closed or unavailable.
  * Added registry cleanup support to prevent outdated scheduling state from lingering.


components/spider-scheduler/src/error.rs
```diff
@@ -32,7 +32,6 @@ pub enum SchedulerError {
     #[error(transparent)]
     Storage(#[from] StorageClientError),
 
-    /// The dispatching queue is closed and can no longer accept assignments.
```
LinZhihao-723, line 35: This error is not just used in the reader side, but also the writer side.

components/spider-scheduler/src/core_impl/resource_group_round_robin/dispatch_queue.rs
```diff
@@ -0,0 +1,978 @@
...
+    ///
+    /// The caller must hold exactly one outstanding hint for this group. A decrement with no hint
+    /// outstanding wraps the count, permanently corrupting the group's hint accounting.
+    fn decrement_living_hint(&self) {
+        self.living_hint.fetch_sub(1, Ordering::AcqRel);
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
LinZhihao-723, line 407: Haven't self-reviewd this test mod yet. May publish some more cleaning commits.

components/spider-scheduler/src/core_impl/resource_group_round_robin/dispatch_queue.rs
```diff
@@ -0,0 +1,978 @@
...
+    ///
+    /// Returns an error if:
+    ///
+    /// * [`SchedulerError::DispatchQueueClosed`] if either queue this publication writes into is
+    ///   closed: the group's dispatch queue, in which case no hint is published, or the broadcast
+    ///   queue, in which case the assignment is queued but uncovered. The two are indistinguishable
+    ///   to the caller, and a running scheduler never observes the latter: the registry holds both
+    ///   ends of the broadcast queue, so it can only close once the registry itself is gone, i.e.
+    ///   once the scheduler is shutting down; the error is unreachable in a running scheduler but
+    ///   stays fatal to the core.
```
sitaowang1998: This section is hard to read.

LinZhihao-723: Simplified.

LinZhihao-723 (commented):
Test cases are written with assistant from claude agent. I think overall they make sense.

sitaowang1998 (approved):

PR #454, opened 2026-08-22
ci: Add Huntsman end-to-end testing using a Spider cluster deployed with Docker Compose.

sitaowang1998:
# Description

This PR:
- Adds a task that:
  1. Builds rust binaries & images.
  2. Pulls MariaDB image.
  3. Starts Docker Compose cluster with test libraries mounted into worker containers.
  4. Runs the e2e test driver.
  5. Cleans up the cluster.
- Modifies GitHub filter actions to support fine-grained Rust tests filter.
- Adds a new job that runs the e2e test task on x64 and arm.

# Checklist

  breaking change.

# Validation performed

- [x] The e2e test task pass on a Ubuntu 22.04 x64 machine.
- [x] GitHub workflows pass.

## New Features
- Added automated Rust end-to-end testing on amd64 and arm64 environments.
- End-to-end tests can run on schedules, manually, or when relevant changes are detected.
- Added local Docker Compose support for managing end-to-end test dependencies.

## Tests
- Improved change detection to distinguish Rust unit-test and end-to-end test changes.
- Rust unit tests now run only when relevant files are modified.
- Added reusable test package staging to streamline Huntsman test execution.
- Local test environments now accept configurable storage ports and database images.


taskfiles/test.yaml
```diff
@@ -40,6 +40,12 @@ tasks:
         vars:
           STORAGE_TASK: "spider-huntsman-unit-tests-executor"
 
+  rust-e2e-tests:
+    cmds:
+      - task: ":build:rust"
+      - task: ":docker:build"
```
LinZhihao-723: `:docker` is not a proper imported mod, right? I'm not sure if this would cause any checksum file issues. When I run this locally for multiple times, the second run will trigger the docker build twice even nothing has changed. I'm not sure if this is a problem we need to solve now, but it's worth to point it out.

taskfiles/test.yaml
```diff
@@ -40,6 +40,12 @@ tasks:
...
+      - task: "spider-huntsman-e2e-executor"
```
LinZhihao-723: Why do we call these three tasks in sequence, instead of making Rust-build and docker-build as dependency tasks?

taskfiles/test.yaml
```diff
@@ -40,6 +40,12 @@ tasks:
         vars:
           STORAGE_TASK: "spider-huntsman-unit-tests-executor"
 
+  rust-e2e-tests:
```
LinZhihao-723: Shall we rename to `huntsman-e2e-tests`? We use `rust-unit-tests` since it is supposed to cover all Rust unit tests in the workspace, including cases like huntsman test harness' unit tests. For this task, we should be explicit that it is for Huntsman e2e.

taskfiles/test.yaml
```diff
@@ -257,3 +263,47 @@ tasks:
             echo ""
           fi
         done
+
+  spider-huntsman-e2e-executor:
+    internal: true
+    vars:
+      G_RUST_RELEASE_DIR: "{{.G_RUST_BUILD_DIR}}/release"
+      G_TDL_PACKAGES_DIR: "{{.G_BUILD_DIR}}/tdl_packages"
```
LinZhihao-723: - Why do we need to redefine these two variables?
- Any `G_` prefix variables should not appear in this section iiuc.

taskfiles/test.yaml
```diff
@@ -257,3 +263,47 @@ tasks:
...
+        sh: "cat '{{.G_BUILD_DIR}}/spider-worker-image.id'"
+    dir: "{{.ROOT_DIR}}"
+    deps: ["toolchains:rust"]
+    cmds:
+      - |-
+        . "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
+        cargo build --release --package huntsman-nn-tasks
+        mkdir -p "{{.G_TDL_PACKAGES_DIR}}/nn"
+        cp "{{.G_RUST_RELEASE_DIR}}/libnn.so" \
+           "{{.G_TDL_PACKAGES_DIR}}/nn/libnn.so"
```
LinZhihao-723: Can we make a separate task for this for better readability?

taskfiles/test.yaml
```diff
@@ -257,3 +263,47 @@ tasks:
...
+      - |-
+        if [ -f "{{.ROOT_DIR}}/tools/deployment/spider-compose/.env" ]; then
+          . "{{.ROOT_DIR}}/tools/deployment/spider-compose/.env"
+        fi
+        docker pull "${SPIDER_DATABASE_IMAGE_REF:-mariadb:10.11.16}"
+      - task: ":docker:compose:local:clean"
+      - defer:
+          task: ":docker:compose:local:clean"
```
LinZhihao-723: This is confusing:
- Why do we need the first `clean`? Is it to clean any running/previously-started compose?
- Why the second `clean` is initiated here? Should be set this `defer` after `up`?

taskfiles/test.yaml
```diff
@@ -257,3 +263,47 @@ tasks:
...
+          SPIDER_PULL_POLICY: "never"
+          SPIDER_STORAGE_IMAGE_REF: "{{.SPIDER_STORAGE_IMAGE_REF}}"
+          SPIDER_SCHEDULER_IMAGE_REF: "{{.SPIDER_SCHEDULER_IMAGE_REF}}"
+          SPIDER_WORKER_IMAGE_REF: "{{.SPIDER_WORKER_IMAGE_REF}}"
+          SPIDER_WORKER_LIBRARY_DIR: "{{.G_TDL_PACKAGES_DIR}}"
+      - |-
+        . "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
+        if [ -f "{{.ROOT_DIR}}/tools/deployment/spider-compose/.env" ]; then
+          . "{{.ROOT_DIR}}/tools/deployment/spider-compose/.env"
+        fi
```
LinZhihao-723: Is this only for `SPIDER_STORAGE_PUBLISHED_PORT`?
This is too verbose to repeat twice in the task execution section. Might be worth to make it a formal parameter under `vars`.

taskfiles/test.yaml
```diff
@@ -257,3 +263,47 @@ tasks:
...
+        SPIDER_ENDPOINT="http://127.0.0.1:${SPIDER_STORAGE_PUBLISHED_PORT:-50051}" \
+        SPIDER_CONCURRENCY="8" \
+          cargo test --release --package e2e --test nn -- --nocapture
```
LinZhihao-723: Our conventioin is to use `nextest` right?

taskfiles/test.yaml
```diff
@@ -40,6 +40,12 @@ tasks:
         vars:
           STORAGE_TASK: "spider-huntsman-unit-tests-executor"
 
+  rust-e2e-tests:
+    cmds:
+      - task: ":build:rust"
```
LinZhihao-723: Why do we need this step?

taskfiles/test.yaml
```diff
@@ -257,3 +263,47 @@ tasks:
             echo ""
           fi
         done
+
+  spider-huntsman-e2e-executor:
```
LinZhihao-723: We shouldn't call this task `executor`. The MariaDB wrapper is called `executor` cuz it parameterizes the task to execute under a spawned MariaDB instance. This task does not parameterize anything but calls the hardcoded test case binary.
If we make ":docker:build" a dependency of this task, we should probably remove it and move its implementation to `huntsman-e2e-tests` directly.

LinZhihao-723 (changes requested):
Reviewed only `test.yaml` taskfile. Please read my comments carefully and reply. The current implementation looks somewhat ad hoc. If it has to be in the current form, at least we need to document why.

taskfiles/test.yaml
sitaowang1998: * Yes. The first `clean` is to clean any running/previously-started Docker Compose cluster.
* `defer` is usually set before the actual run. If it is set after `up`, and `up` fails midway, the `clean` would not run.

taskfiles/test.yaml
sitaowang1998: This step loads all environment variables a user defines that overrides the default values, not use port.

taskfiles/test.yaml
sitaowang1998: To build the test packages that will be mounted into worker containers.

taskfiles/test.yaml
LinZhihao-723: - I don't think the first clean is needed tbh. Let's remove it.
- I don't think it's a common case. If you look at the official `defer` example [here](https://taskfile.dev/docs/guide#doing-task-cleanup-with-defer), `defer` should be called after the previous task/command successfully provisions the resources. I think in this particular case, the only reason you want to defer is `up` is not an all-or-nothing task (or maybe it is). If so, I think we should document something like "`docker compose up` is not transactional and has no rollback."

taskfiles/test.yaml
LinZhihao-723: What are the other variables needed to call the unit test driver? These env vars are for the actual Spider cluster, no?

taskfiles/test.yaml
LinZhihao-723: Then why does this step happen as a part of the current `spider-huntsman-e2e-executor`? It's repeated.

taskfiles/test.yaml
sitaowang1998: There is nothing needed for the CI. However, when testing locally we might need to adjust the queue size, number of worker replicas, the port in case 50051 is used, etc.

LinZhihao-723: I don't think you get what I'm asking.

```yaml
      - |-
        . "{{.G_RUST_TOOLCHAIN_ENV_FILE}}"
        if [ -f "{{.ROOT_DIR}}/tools/deployment/spider-compose/.env" ]; then
          . "{{.ROOT_DIR}}/tools/deployment/spider-compose/.env"
        fi
        SPIDER_ENDPOINT="http://127.0.0.1:${SPIDER_STORAGE_PUBLISHED_PORT:-50051}" \
        SPIDER_CONCURRENCY="8" \
          cargo test --release --package e2e --test nn -- --nocapture
```
In this particular section, the env file is loaded. But it looks like this is only for `SPIDER_STORAGE_PUBLISHED_PORT`; the rest of the env vars, including the settings you mentioned, have no effect on this step.

Similarly, in the following step:

```yaml
      - |-
        if [ -f "{{.ROOT_DIR}}/tools/deployment/spider-compose/.env" ]; then
          . "{{.ROOT_DIR}}/tools/deployment/spider-compose/.env"
        fi
        docker pull "${SPIDER_DATABASE_IMAGE_REF:-mariadb:10.11.16}"
```

The env file is only used for `SPIDER_DATABASE_IMAGE_REF`.

My question is: since the rest of the env variables are not used in the above two steps, should we define `SPIDER_STORAGE_PUBLISHED_PORT` and `SPIDER_DATABASE_IMAGE_REF` as task variables in the `vars` section?

taskfiles/test.yaml
sitaowang1998: * I don't understand what do you mean by a proper imported mod.
* `docker:build` does not have any checksum file in the task itself. Instead, the underlying script calls `docker build`, which does use checksum and layer caching. I've tried on my side to run the test task twice. The image building runs the second time, but it hits the cache and does not actually run the builds, and takes nearly no time.
```
sitao@baker3:~/spider$ task test:huntsman-e2e-tests
task: [docker:build-storage] '/home/sitao/spider/tools/docker/build.sh' storage
[+] Building 0.9s (22/22) FINISHED                                                                                                                          docker:default
 => [internal] load build definition from Dockerfile                                                                                                                  0.0s
 => => transferring dockerfile: 1.36kB                                                                                                                                0.0s
 => resolve image config for docker-image://docker.io/docker/dockerfile:1                                                                                             0.2s
 => CACHED docker-image://docker.io/docker/dockerfile:1@sha256:ecfaec9ed6d810b56388c508f4121597bfbba70d41a6dfeee4d8cad5f295fc32                                       0.0s
 => [internal] load metadata for docker.io/library/ubuntu:jammy                                                                                                       0.2s
 => [internal] load .dockerignore                                                                                                                                     0.0s
 => => transferring context: 2B                                                                                                                                       0.0s
 => [internal] load build context                                                                                                                                     0.1s
 => => transferring context: 214.17kB                                                                                                                                 0.1s
 => [builder  1/13] FROM docker.io/library/ubuntu:jammy@sha256:2edbbc5dc405e9612ba3584ce95480277e3eb374407b5505fe26f17df77c7dbc                                       0.0s
 => CACHED [runtime-base 2/2] RUN useradd --create-home spider-user                                                                                                   0.0s
 => CACHED [builder  2/13] WORKDIR /spider                                                                                                                            0.0s
 => CACHED [builder  3/13] COPY --link Cargo.toml Cargo.lock taskfile.yaml ./                                                                                         0.0s
 => CACHED [builder  4/13] COPY --link .cargo ./.cargo                                                                                                                0.0s
 => CACHED [builder  5/13] COPY --link components ./components                                                                                                        0.0s
 => CACHED [builder  6/13] COPY --link examples/huntsman ./examples/huntsman                                                                                          0.0s
 => CACHED [builder  7/13] COPY --link taskfiles ./taskfiles                                                                                                          0.0s
 => CACHED [builder  8/13] COPY --link tests/huntsman ./tests/huntsman                                                                                                0.0s
 => CACHED [builder  9/13] COPY --link tools/scripts ./tools/scripts                                                                                                  0.0s
 => CACHED [builder 10/13] COPY --link tools/yscope-dev-utils ./tools/yscope-dev-utils                                                                                0.0s
 => CACHED [builder 11/13] RUN tools/scripts/lib_install/ubuntu/install-dev-huntsman.sh                                                                               0.0s
 => CACHED [builder 12/13] RUN sh -c "$(curl --fail --location --silent --show-error https://taskfile.dev/install.sh)" --     -d -b /usr/local/bin "v3.49.1"          0.0s
 => CACHED [builder 13/13] RUN task build:rust                                                                                                                        0.0s
 => CACHED [storage 1/1] COPY --from=builder --link /spider/build/rust-targets/release/spider-storage /usr/local/bin/                                                 0.0s
 => exporting to image                                                                                                                                                0.0s
 => => exporting layers                                                                                                                                               0.0s
 => => writing image sha256:208b58b3f73e60742bf414391b047f492ae260d9e4c36883773e0990ccd99b1c                                                                          0.0s
task: [docker:build-scheduler] '/home/sitao/spider/tools/docker/build.sh' scheduler
[+] Building 0.8s (22/22) FINISHED                                                                                                                          docker:default
 => [internal] load build definition from Dockerfile                                                                                                                  0.0s
 => => transferring dockerfile: 1.36kB                                                                                                                                0.0s
 => resolve image config for docker-image://docker.io/docker/dockerfile:1                                                                                             0.1s
 => CACHED docker-image://docker.io/docker/dockerfile:1@sha256:ecfaec9ed6d810b56388c508f4121597bfbba70d41a6dfeee4d8cad5f295fc32                                       0.0s
 => [internal] load metadata for docker.io/library/ubuntu:jammy                                                                                                       0.1s
 => [internal] load .dockerignore                                                                                                                                     0.0s
 => => transferring context: 2B                                                                                                                                       0.0s
 => [builder  1/13] FROM docker.io/library/ubuntu:jammy@sha256:2edbbc5dc405e9612ba3584ce95480277e3eb374407b5505fe26f17df77c7dbc                                       0.0s
 => [internal] load build context                                                                                                                                     0.1s
 => => transferring context: 214.17kB                                                                                                                                 0.1s
 => CACHED [runtime-base 2/2] RUN useradd --create-home spider-user                                                                                                   0.0s
 => CACHED [builder  2/13] WORKDIR /spider                                                                                                                            0.0s
 => CACHED [builder  3/13] COPY --link Cargo.toml Cargo.lock taskfile.yaml ./                                                                                         0.0s
 => CACHED [builder  4/13] COPY --link .cargo ./.cargo                                                                                                                0.0s
 => CACHED [builder  5/13] COPY --link components ./components                                                                                                        0.0s
 => CACHED [builder  6/13] COPY --link examples/huntsman ./examples/huntsman                                                                                          0.0s
 => CACHED [builder  7/13] COPY --link taskfiles ./taskfiles                                                                                                          0.0s
 => CACHED [builder  8/13] COPY --link tests/huntsman ./tests/huntsman                                                                                                0.0s
 => CACHED [builder  9/13] COPY --link tools/scripts ./tools/scripts                                                                                                  0.0s
 => CACHED [builder 10/13] COPY --link tools/yscope-dev-utils ./tools/yscope-dev-utils                                                                                0.0s
 => CACHED [builder 11/13] RUN tools/scripts/lib_install/ubuntu/install-dev-huntsman.sh                                                                               0.0s
 => CACHED [builder 12/13] RUN sh -c "$(curl --fail --location --silent --show-error https://taskfile.dev/install.sh)" --     -d -b /usr/local/bin "v3.49.1"          0.0s
 => CACHED [builder 13/13] RUN task build:rust                                                                                                                        0.0s
 => CACHED [scheduler 1/1] COPY --from=builder --link /spider/build/rust-targets/release/spider-scheduler /usr/local/bin/                                             0.0s
 => exporting to image                                                                                                                                                0.0s
 => => exporting layers                                                                                                                                               0.0s
 => => writing image sha256:52efcfce4b7d28a6196c5df0e0ee978be62755d2d2a82943d8f80ba61fd179c4                                                                          0.0s
task: [docker:build-worker] '/home/sitao/spider/tools/docker/build.sh' worker
[+] Building 0.8s (23/23) FINISHED                                                                                                                          docker:default
 => [internal] load build definition from Dockerfile                                                                                                                  0.0s
 => => transferring dockerfile: 1.36kB                                                                                                                                0.0s
 => resolve image config for docker-image://docker.io/docker/dockerfile:1                                                                                             0.1s
 => CACHED docker-image://docker.io/docker/dockerfile:1@sha256:ecfaec9ed6d810b56388c508f4121597bfbba70d41a6dfeee4d8cad5f295fc32                                       0.0s
 => [internal] load metadata for docker.io/library/ubuntu:jammy                                                                                                       0.1s
 => [internal] load .dockerignore                                                                                                                                     0.0s
 => => transferring context: 2B                                                                                                                                       0.0s
 => [builder  1/13] FROM docker.io/library/ubuntu:jammy@sha256:2edbbc5dc405e9612ba3584ce95480277e3eb374407b5505fe26f17df77c7dbc                                       0.0s
 => [internal] load build context                                                                                                                                     0.1s
 => => transferring context: 214.17kB                                                                                                                                 0.1s
 => CACHED [runtime-base 2/2] RUN useradd --create-home spider-user                                                                                                   0.0s
 => CACHED [builder  2/13] WORKDIR /spider                                                                                                                            0.0s
 => CACHED [builder  3/13] COPY --link Cargo.toml Cargo.lock taskfile.yaml ./                                                                                         0.0s
 => CACHED [builder  4/13] COPY --link .cargo ./.cargo                                                                                                                0.0s
 => CACHED [builder  5/13] COPY --link components ./components                                                                                                        0.0s
 => CACHED [builder  6/13] COPY --link examples/huntsman ./examples/huntsman                                                                                          0.0s
 => CACHED [builder  7/13] COPY --link taskfiles ./taskfiles                                                                                                          0.0s
 => CACHED [builder  8/13] COPY --link tests/huntsman ./tests/huntsman                                                                                                0.0s
 => CACHED [builder  9/13] COPY --link tools/scripts ./tools/scripts                                                                                                  0.0s
 => CACHED [builder 10/13] COPY --link tools/yscope-dev-utils ./tools/yscope-dev-utils                                                                                0.0s
 => CACHED [builder 11/13] RUN tools/scripts/lib_install/ubuntu/install-dev-huntsman.sh                                                                               0.0s
 => CACHED [builder 12/13] RUN sh -c "$(curl --fail --location --silent --show-error https://taskfile.dev/install.sh)" --     -d -b /usr/local/bin "v3.49.1"          0.0s
 => CACHED [builder 13/13] RUN task build:rust                                                                                                                        0.0s
 => CACHED [worker 1/2] COPY --from=builder --link /spider/build/rust-targets/release/spider-execution-manager /usr/local/bin/                                        0.0s
 => CACHED [worker 2/2] COPY --from=builder --link /spider/build/rust-targets/release/spider-task-executor /usr/local/bin/                                            0.0s
 => exporting to image                                                                                                                                                0.0s
 => => exporting layers                                                                                                                                               0.0s
 => => writing image sha256:263115d658a853c9078a304b700ec134803d2d55608d63994a7f49f4a34ff557                                                                          0.0s
```

LinZhihao-723: - Similar like `toolchians`:
  ```
  includes:
      toolchains: "toolchains.yaml"
  ```
- Fine. The current behavior looks good then.

I don't think we need to import `:docker` for now then.

sitaowang1998: We should also need to import `build` as well.

LinZhihao-723: I think we don't need to import it for now. We import `toolchain` 'cuz that's the only way to solve the checksum issue.

taskfiles/test.yaml
```diff
@@ -40,6 +42,46 @@ tasks:
...
+      - ":build:rust"
+      - ":docker:build"
+    cmds:
+      - task: "build-and-stage-huntsman-tdl-package"
+        vars:
+          CARGO_PACKAGE_NAME: "huntsman-nn-tasks"
+          CARGO_LIBRARY_NAME: "nn"
+          TDL_PACKAGE_NAME: "nn"
+      - "docker pull \"{{.SPIDER_DATABASE_IMAGE_REF}}\""
+      - task: ":docker:compose:local:clean"
```
LinZhihao-723: As suggested in https://github.com/y-scope/spider/pull/454#discussion_r3856422110, this is not needed, right?

```suggestion
```

taskfiles/test.yaml
```diff
@@ -40,6 +42,46 @@ tasks:
...
+  huntsman-e2e-tests:
+    vars:
+      SPIDER_STORAGE_PUBLISHED_PORT: >-
+        {{.SPIDER_STORAGE_PUBLISHED_PORT | default "50051"}}
+      SPIDER_DATABASE_IMAGE_REF: >-
+        {{.SPIDER_DATABASE_IMAGE_REF | default "mariadb:10.11.16"}}
+    dir: "{{.ROOT_DIR}}"
+    deps:
+      - ":build:rust"
+      - ":docker:build"
```
LinZhihao-723: As discussed https://github.com/y-scope/spider/pull/454#discussion_r3856085419, this step is not needed, right?


```suggestion
    deps: [":docker:build"]
```

taskfiles/test.yaml
```diff
@@ -40,6 +42,46 @@ tasks:
...
+      - task: "build-and-stage-huntsman-tdl-package"
+        vars:
+          CARGO_PACKAGE_NAME: "huntsman-nn-tasks"
+          CARGO_LIBRARY_NAME: "nn"
+          TDL_PACKAGE_NAME: "nn"
+      - "docker pull \"{{.SPIDER_DATABASE_IMAGE_REF}}\""
+      - task: ":docker:compose:local:clean"
+      # Register cleanup before startup so a partially created Compose deployment is removed if
+      # `docker:compose:local:up` fails.
+      - defer:
```
LinZhihao-723:
```suggestion
      - defer:
          # Register cleanup in advance so a partially created Compose deployment is removed if
          # `docker:compose:local:up` fails.
```

taskfiles/test.yaml
```diff
@@ -257,3 +296,23 @@ tasks:
             echo ""
           fi
         done
+
+  # Builds and stages a Huntsman TDL package library in the task executor's package layout.
+  #
+  # @param {string} CARGO_PACKAGE_NAME The Cargo package that builds the library.
+  # @param {string} CARGO_LIBRARY_NAME The Cargo library target name.
+  # @param {string} TDL_PACKAGE_NAME The TDL package directory and runtime library name.
```
LinZhihao-723:
```suggestion
  # @param {string} TDL_PACKAGE_NAME The TDL package name.
```

LinZhihao-723 (changes requested):
You should read review comments more carefully. If you don't apply the change, you should comment why.
This is a basic rule you should follow: you should **ALWAYS** do it without needing me to point it out in every PR.

taskfiles/docker.yaml
```diff
@@ -55,23 +55,25 @@ tasks:
...
   #
+  # * Users can override image refs for each service; when unset, each defaults to
+  #   `ghcr.io/y-scope/spider/<service>:${SPIDER_IMAGE_TAG:-main}`.
+  # * Values in the `env` section take precedence over those loaded from the dotenv file.
+  #
+  # @param {string} SPIDER_ENV_FILE An absolute path to a dotenv file to load into the deployment's
+  # environment. If not set, no dotenv file is loaded.
   # @param {string} SPIDER_PULL_POLICY The image pull policy. If not set, default to "always".
-  # @param {string} SPIDER_STORAGE_PUBLISHED_PORT The host port for Spider Storage.
-  # @param {string} SPIDER_DATABASE_IMAGE_REF The database image reference.
```
LinZhihao-723: Prefer to specify these overrides through the env file.

taskfiles/docker.yaml
```diff
@@ -55,23 +55,25 @@ tasks:
...
   # @param {string} SPIDER_STORAGE_IMAGE_REF The storage image reference.
   # @param {string} SPIDER_SCHEDULER_IMAGE_REF The scheduler image reference.
   # @param {string} SPIDER_WORKER_IMAGE_REF The worker image reference.
```
LinZhihao-723, line 69: Kept these parameters to provide explicit Spider component image control for development purposes.

taskfiles/docker.yaml
```diff
@@ -55,23 +55,25 @@ tasks:
...
   # @param {string} SPIDER_WORKER_LIBRARY_DIR The host directory containing the libraries.
```
LinZhihao-723, line 70: Kept this one since we need to process abs/relative paths.

taskfiles/test.yaml
```diff
@@ -57,15 +59,24 @@ tasks:
           CARGO_LIBRARY_NAME: "nn"
           TDL_PACKAGE_NAME: "nn"
       - "docker pull \"{{.SPIDER_DATABASE_IMAGE_REF}}\""
+      - |-
+        cat > "{{.SPIDER_ENV_FILE}}" << 'EOF'
+        SPIDER_STORAGE_PUBLISHED_PORT={{.SPIDER_STORAGE_PUBLISHED_PORT}}
+        SPIDER_DATABASE_IMAGE_REF={{.SPIDER_DATABASE_IMAGE_REF}}
+        SPIDER_WORKER_REPLICAS={{.SPIDER_WORKER_REPLICAS}}
+        SPIDER_SCHEDULER_ROUND_ROBIN_DISPATCH_QUEUE_CAPACITY={{mul .SPIDER_WORKER_REPLICAS 4}}
```
LinZhihao-723, line 70: Set dispatch queue size.

LinZhihao-723 (commented):
Reimplemented dotenv.

.github/actions/filter-relevant-changes/action.yaml
```diff
@@ -51,8 +64,20 @@ runs:
...
+            - "examples/huntsman/**"
+            - "tests/huntsman/**"
+          rust_unit_tests:
+            - *_rust_common
+            - "examples/huntsman/**"
+            - "tests/huntsman/em-runtime/**"
+            - "tests/huntsman/integration-test-tasks/**"
+            - "tests/huntsman/task-executor/**"
+            - "tests/huntsman/tdl-integration/**"
+            - "tests/huntsman/test-utils/**"
```
LinZhihao-723:
```suggestion
            - "tests/huntsman/**"
```
`e2e` crate has unit tests. They're testing the e2e harness.

.github/actions/filter-relevant-changes/action.yaml
```diff
@@ -51,8 +64,20 @@ runs:
...
           rust:
-            - *_common
-            - ".cargo/**"
-            - "Cargo.lock"
-            - "Cargo.toml"
-            - "components/**"
+            - *_rust_common
+            - "examples/huntsman/**"
+            - "tests/huntsman/**"
+          rust_unit_tests:
```
LinZhihao-723: Actually, using `**` makes it the same as rust_changed. We should drop it.

.github/actions/filter-relevant-changes/action.yaml
```diff
@@ -16,6 +16,12 @@ outputs:
   rust_changed:
     description: "Whether Rust related files have changed."
     value: "${{steps.paths-filter.outputs.rust}}"
+  rust_unit_tests_changed:
+    description: "Whether Rust unit-test related files have changed."
+    value: "${{steps.paths-filter.outputs.rust_unit_tests}}"
+  rust_e2e_tests_changed:
```
LinZhihao-723: Rename to `huntsman_e2e_tests_required`.
`changed` suffix here would be weird: the test itself doesn't change, but the test workflow will be needed.

sitaowang1998 (commented):
LGTM.

There is a misunderstanding between us. You use `.env` file to pass the environment variables to the docker task. I was thinking leaving most configurations inside the `.env` itself and points the docker task to it. Either way works.

BTW, can't approve my own PR. lol

LinZhihao-723 (changes requested):

.github/actions/filter-relevant-changes/action.yaml
```diff
@@ -51,8 +61,12 @@ runs:
...
           rust:
-            - *_common
-            - ".cargo/**"
-            - "Cargo.lock"
-            - "Cargo.toml"
-            - "components/**"
+            - *_rust_common
+            - "examples/huntsman/**"
+            - "tests/huntsman/**"
+          rust_e2e_tests:
```
LinZhihao-723:
```suggestion
          huntsman_e2e_tests:
```

.github/actions/filter-relevant-changes/action.yaml
```diff
@@ -16,6 +16,9 @@ outputs:
   rust_changed:
     description: "Whether Rust related files have changed."
     value: "${{steps.paths-filter.outputs.rust}}"
+  huntsman_e2e_tests_required:
+    description: "Whether Huntsman end-to-end-test needs to run."
+    value: "${{steps.paths-filter.outputs.rust_e2e_tests}}"
```
LinZhihao-723:
```suggestion
    value: "${{steps.paths-filter.outputs.huntsman_e2e_tests}}"
```

LinZhihao-723 (changes requested):

tools/scripts/docker/resolve-huntsman-e2e-endpoint.py
LinZhihao-723, line 1: Rewrit as a uv script.
Just realized our Python linting flow won't lint scripts under `tools/scripts`. We should fix that. I've linted this file manually.
@coderabbitai Create an issue for adding Python script linting into our Python linting task.

taskfiles/test.yaml
```diff
@@ -40,6 +42,67 @@ tasks:
...
+          SPIDER_WORKER_LIBRARY_DIR: "{{.G_TDL_PACKAGES_DIR}}"
+      - >-
+        tools/scripts/docker/resolve-huntsman-e2e-endpoint.py connect
+          --network "{{.SPIDER_COMPOSE_NETWORK}}"
+          --storage-port "{{.SPIDER_STORAGE_PORT}}"
+          --published-port "{{.SPIDER_STORAGE_PUBLISHED_PORT}}"
+        > "{{.SPIDER_ENDPOINT_FILE}}"
+      - defer: >-
+          tools/scripts/docker/resolve-huntsman-e2e-endpoint.py disconnect
+            --network "{{.SPIDER_COMPOSE_NETWORK}}"
```
LinZhihao-723, line 98: Do this as a single step so its failure is self-contained.

LinZhihao-723 (approved):
For the PR title, how about:
```
ci: Add Huntsman end-to-end testing using a Spider cluster deployed with Docker Compose.
```

PR #450, opened 2026-08-21
feat(spider-storage): Add support for MySQL database.

sitaowang1998:
# Description

## Background
The current `Spider` only support MariaDB as the database in storage, and it uses `RETURNING ...` syntax, which is a MariaDB-only extension.

During @20001020ycx testing, the incompatibility with MySQL database was exposed.

## Change
This PR replaces all 5 `RETURNING id` in the code with `sqlx::last_insert_id`. Now `Spider` supports MySQL as database.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

## Unit Tests
Run unit tests with both MariaDB and MySQL 8.0-debian. All tests pass.

## End-to-End Tests
Run the end-to-end tests from #385 using Docker Compose local support. The end-to-end tests run on two clusters, one using the bundled MariaDB, the other one with bundled MariaDB replaced by MySQL 8.0-debian. Both test executions pass.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MariaDB compatibility for creating sessions, jobs, resource groups, execution managers, and scheduler registrations.
  * Generated identifiers are now retrieved reliably after insert operations, with existing error handling and transaction behaviour preserved.

PR #456, opened 2026-08-24
feat(spider-proto): Add optional resource-group fields to the scheduler's next-task request and the storage's execution manager registration.

sitaowang1998:
# Description

This PR adds an optional execution manager in several gRPC protocols:
- Execution manager to scheduler's task request.
- Execution manager's register request includes an optional external resource group id, and the reply includes an optional resource group id. 

# Checklist

  breaking change.

# Validation performed

- [x] GitHub workflows pass.

## Summary by CodeRabbit

* **New Features**
  * Added optional resource-group information to task scheduling requests.
  * Added secure external resource-group credentials to execution-manager registration.
  * Registration responses can include an associated resource-group identifier.
  * Existing workflows continue to operate when no resource group is specified.

* **Bug Fixes**
  * Registration requests with external resource-group credentials are now handled correctly.
  * Unsupported resource-group scheduling requests return a clear response.

### DIFF
--- a/components/spider-proto/storage/storage.proto
+++ b/components/spider-proto/storage/storage.proto
@@ -152,9 +152,14 @@ message VerifyResourceGroupRequest {
   bytes password = 2;
 }
 
+message ExternalResourceGroupCredentials {
+  string external_resource_group_id = 1;
+  bytes password = 2;
+}
+
 message RegisterExecutionManagerRequest {
   string ip_address = 1;
-  optional string external_resource_group_id = 2;
+  ExternalResourceGroupCredentials resource_group_credentials = 2;
 }
 
 message ExecutionManagerIdRequest {
--- a/components/spider-proto/storage/storage.proto
+++ b/components/spider-proto/storage/storage.proto
@@ -154,6 +154,7 @@ message VerifyResourceGroupRequest {
 
 message RegisterExecutionManagerRequest {
   string ip_address = 1;
+  optional string external_resource_group_id = 2;
 }
 
 message ExecutionManagerIdRequest {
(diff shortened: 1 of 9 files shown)


components/spider-proto/storage/storage.proto
```diff
@@ -154,6 +154,7 @@ message VerifyResourceGroupRequest {
 
 message RegisterExecutionManagerRequest {
   string ip_address = 1;
+  optional string external_resource_group_id = 2;
 }
```
LinZhihao-723, line 163: Shouldn't we need the credential to be passed for validation?

LinZhihao-723 (changes requested):

LinZhihao-723 (changes requested):
You should spend more time polishing the PR title, potentially ask your coding agent to come up one from the PR conclusion that follows the repo's convention.
The original PR title `feat(huntsman): Add optional resouce group id in protocol.` has the following issues:
- First, you should use "resource group ID" instead of "resource group id". When an ID appears in a formal document, it should always be in the capital form.
- If you read this PR title without knowing any context, can you tell what is the scope of the protocol change? Is the resource group ID added to every protocol? No, but that is what the title sounds like. I don't mind short PR title/commit messages, for example, #450, but it must be clear enough on its own.
- When you updated the protocol to also include resource group password, the title becomes a stale description: it now not only ships the ID, but a set of credentials. It should be updated accordingly.

I've changed the PR title to fix the above problems.
Nit: Even we touched multiple crates, but the core changes are in `spider-proto` and propagated to other crates. I changed it to `spider-proto`, but u can switch back to `huntsman` if you want.

LinZhihao-723 (approved):

PR #462, opened 2026-08-27
feat(spider-scheduler): Add scheduling state of a single resource group for the resource-group-round-robin scheduler core.

LinZhihao-723:
# Description

This is the third piece of the `resource_group_round_robin` scheduler core, following the job registry in #444 and the dispatch queue in #449. It lands `RgSchedulingState`, the per-resource-group state the core draws assignments from. The core that drives it follows in a later PR, so nothing is wired up and no behaviour changes.

## What the module holds

`RgSchedulingState` owns one resource group's scheduling position: a queue of owed finalizations, an active job list rotated over by a round-robin arm, a pending job queue that feeds it, and two per-tick downgrade buffers. `try_make_assignment` is its whole decision surface — it publishes at most one assignment per call and reports why it could not through `MakeAssignmentError`.

Finalization tasks are dispatched ahead of regular ones. Regular tasks come from the inner round-robin over active jobs, which is the second level of the core's two-level rotation.

## The decisions worth reviewing

**A job spends at most one downgrade life per tick.** A job that yields nothing loses a life, and is retired once it has none left. The two buffers are what keep that accounting honest: a job removed from the active list and a job passed over in the pending queue are both held aside until `apply_downgrades` runs at the end of the tick, so neither can be examined — and charged — twice. Returning them to the pending queue immediately would let a single tick spend the whole budget. Demoted jobs go back to the queue's head with their budget restored; passed-over pending jobs go to its tail with their budget untouched, which is what keeps retirement reachable at all.

**A publication failure is unrecoverable by design.** The task is taken before it is published rather than after. The dispatch queues are unbounded, so `publish` can only fail on a closed queue, and the core answers that by ending the tick and stopping the runtime — there is no later tick to preserve the task for. An earlier revision read the task, published, then removed it; that bought a re-admission that can never happen, at the cost of a three-step dance and a redundant arena lookup per assignment.

**Admission is bounded by the tick's remaining free space, not by a channel bound.** `try_make_assignment` refuses once the group's queue occupancy reaches `free`. This is the dynamic threshold scheme with its control parameter fixed at 1; the queues themselves stay unbounded, as #449 established.

**`pop_regular_task`'s termination rests on `DOWNGRADE_LIVES == 1`.** With one life, a barren active job is always demoted on its first visit, so the loop can never revisit one and needs no visit counter. That precondition is a `const` assert in the function rather than a comment, and it names what a larger budget would require.

# Checklist

  breaking change.

# Validation performed

* [x] Add unit test cases to assert the basic behaviors.

## Summary by CodeRabbit

* **New Features**
  * Added resource-group round-robin scheduling to improve fairness and balance across jobs.
  * Prioritized finalization work while managing active and pending jobs more effectively.
  * Added safeguards for queue capacity, task availability, job downgrades, and queue closure conditions.

* **Bug Fixes**
  * Improved job promotion, retirement tracking, and task dispatch ordering for more predictable scheduling behaviour.

* **Tests**
  * Added comprehensive coverage for scheduling order, admission limits, task publication, queue handling, and job lifecycle transitions.

### DIFF
--- a/components/spider-scheduler/src/core_impl/resource_group_round_robin/scheduling_state.rs
+++ b/components/spider-scheduler/src/core_impl/resource_group_round_robin/scheduling_state.rs
@@ -168,7 +168,7 @@ impl RgSchedulingState {
     ///
     /// `free` is the tick's remaining free space in the whole dispatch buffer, read but not
     /// modified here -- the caller decrements it once the assignment is published. This is only
-    /// used to determine the admission threshold.
+    /// used by the admission threshold.
     ///
     /// # Returns
     ///
@@ -193,7 +193,8 @@ impl RgSchedulingState {
         if !self.has_schedulable_task() {
             return Err(MakeAssignmentError::NoTask);
         }
-        if self.dispatch_queue_size() >= free {
+        if self.dispatch_queue_size() >= free * SHARING_COEFFICIENT {
+            // Admission threshold reached.
             return Err(MakeAssignmentError::DispatchQueueFull);
         }
 
@@ -379,6 +380,9 @@ impl RgSchedulingState {
     }
 }
 
+/// The sharing coefficient used by Choudhury & Hahne's dynamic threshold algorithm.
+const SHARING_COEFFICIENT: usize = 1;
+
 #[cfg(test)]
 mod tests {
     use anyhow::bail;
@@ -411,44 +415,6 @@ mod tests {
 
     /// The free space passed to an assignment attempt that is not meant to reach the threshold.
     const FREE_SPACE: usize = 32;
-
-    /// # Returns
-    ///
-    /// A newly created, inactive scheduling state publishing into `rg_id`'s dispatch queue.
-    fn make_state(
-        dispatch_queue_registry: &DispatchQueueRegistry,
-        rg_id: ResourceGroupId,
-        active_job_list_capacity: usize,
-    ) -> RgSchedulingState {
-        RgSchedulingState::new(
-            rg_id,
-            dispatch_queue_registry.get_dispatch_queue_writer(rg_id),
-            active_job_list_capacity,
-        )
-    }
-
-    /// Registers a job of `num_tasks` buffered ready tasks, whose task indices are `0..num_tasks`.
-    ///
-    /// # Returns
-    ///
-    /// The key of the registered job, which still needs a scheduling position.
-    ///
-    /// # Errors
-    ///
-    /// Returns an error if:
-    ///
-    /// * [`anyhow::Error`] if the job is already registered.
-    fn make_job_entry(
-        registry: &mut JobRegistry,
-        job_id: JobId,
-        num_tasks: usize,
-    ) -> anyhow::Result<JobKey> {
-        let task_indices: Vec<TaskIndex> = (0..num_tasks).collect();
-        let UpsertOutcome::New(job_key) = registry.upsert(job_id, task_indices) else {
-            bail!("job {job_id} is already registered");
-        };
-        Ok(job_key)
-    }
 
     /// One scheduling state and the structures it publishes into.
     struct StateFixture {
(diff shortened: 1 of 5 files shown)


components/spider-scheduler/src/core_impl/resource_group_round_robin/scheduling_state.rs
```diff
@@ -0,0 +1,868 @@
...
+
+use super::dispatch_queue::RgDispatchQueueWriter;
+use super::job_registry::DOWNGRADE_LIVES;
+use super::job_registry::JobKey;
+use super::job_registry::JobRegistry;
+use crate::core::TaskAssignmentIdIssuer;
+
+/// Errors returned by assignment decision-making.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, thiserror::Error)]
+pub(super) enum MakeAssignmentError {
```
sitaowang1998, line 23: Can we have a better naming? Maybe `AssignmentError` or `AssignmentDecisionError`?

components/spider-scheduler/src/core_impl/resource_group_round_robin/scheduling_state.rs
```diff
@@ -0,0 +1,868 @@
...
+    /// Gives a newly registered job its scheduling position in this group.
+    pub(super) fn place_new_job(&mut self, job_key: JobKey) {
+        if self.active_jobs.len() < self.active_job_list_capacity {
+            self.active_jobs.push(job_key);
+        } else {
+            self.pending_jobs.push_back(job_key);
+        }
+    }
+
+    /// Tops the active job list up to capacity from the pending job queue.
```
sitaowang1998:
```suggestion
    /// Fills the active job list up to capacity from the pending job queue.
```

components/spider-scheduler/src/core_impl/resource_group_round_robin/scheduling_state.rs
```diff
@@ -0,0 +1,868 @@
...
+    ///  * The task index to dispatch.
+    /// * [`None`] if no active or pending job yields a task.
+    fn pop_regular_task(
+        &mut self,
+        job_registry: &mut JobRegistry,
+        jobs_to_retire: &mut Vec<JobKey>,
+    ) -> Option<(JobId, TaskIndex)> {
+        const _: () = assert!(
+            1 == DOWNGRADE_LIVES,
+            "An barren active job with no tasks is demoted on its first visit, so this loop never \
```
sitaowang1998:
```suggestion
            "A barren active job with no tasks is demoted on its first visit, so this loop never \
```

components/spider-scheduler/src/core_impl/resource_group_round_robin/scheduling_state.rs
```diff
@@ -0,0 +1,868 @@
...
+             revisits one; a larger downgrade budget needs a per-tick buffer for active jobs \
+             already visited."
+        );
+
+        loop {
```
sitaowang1998, line 279: 1. We need a comment on this loop. Explain briefly what this loop does.
2. It might be my problem, but I don't see any return of `None` if no active or pending job yields a task. In fact, I don't see any return in this case.

sitaowang1998 (changes requested):

components/spider-scheduler/src/core_impl/resource_group_round_robin/scheduling_state.rs
LinZhihao-723: 1. tbh I don't think the implementation of this function is hard to understand if we refer to the implementation doc. I rewrote the loop manually on top of Claude's implementation to make it straightforward:
    - Line 275 to 279: the active job list is already empty. We either promote a job from the pending queue or the loop stops.
    - Line 281 to 283: Adjust the arm.
    - Line 284 to 288: Get the job entry. If the job entry is already deleted by the registry, it should be discarded and replaced by a pending job. This is a well-documented behavior according to the design doc.
    - Line 289 to 293: A new task assignment is made.
    - Line 295 to 301: The job has no more tasks to schedule. Downgrade it. This is also a documented behavior according to the design doc.

    Each section is short and self-explanatory, assuming readers have context of how the scheduling policy works. Documenting the policy details is not the responsibility of an inline comment.
2. In line 276,  we apply the try operation to `pop_promotable_job`, which returns `None` if there's no job that can be promoted. `None` is propagated from this call.

components/spider-scheduler/src/core_impl/resource_group_round_robin/scheduling_state.rs
LinZhihao-723: My bad. I replaced "barren" with "exhausted," but I thought the agent switched it back. Will do another round of check to fix it.

components/spider-scheduler/src/core_impl/resource_group_round_robin/scheduling_state.rs
LinZhihao-723: The naming is to match `try_make_assignment` method, which is the only exposed API that returns this error.
- `AssignmentError` is more confusing as it sounds like an error of the assignment.
-  `AssignmentDecisionError` is a bit broad while we want to tighten the scope that this error is only used for `try_make_assignment`.

Let me improve the docstring for better clarification.

LinZhihao-723: Updated docstring.

components/spider-scheduler/src/core_impl/resource_group_round_robin/scheduling_state.rs
LinZhihao-723: As discussed offline, "top up" is a better term to match the behavior of this method. Rewrite it as "Tops up the active job list to capacity..."

components/spider-scheduler/src/core_impl/resource_group_round_robin/scheduling_state.rs
LinZhihao-723: Fixed.

sitaowang1998 (approved):

PR #464, opened 2026-08-31
feat(spider-storage)!: Add resource group verification in execution manager registration.

sitaowang1998:
# Description

This PR:
- Adds a nullable resource group id in execution manager database table.
- Adds resource group verification in database execution manager register function.
- Adds `UNAUTHENTICATED` gRPC status for wrong resource group id or password. The parsing of this status code in execution manager will be in a later PR.
- Adds unit tests.

> [!Note]
> This PR is a breaking change because the database schema changes.

# Checklist

  breaking change.

# Validation performed

- [x] GitHub workflows pass.

## New Features
- Execution managers can now be registered with optional resource-group credentials.
- Successful registrations report their associated resource group.
- Resource-group credentials are validated during registration.

## Bug Fixes
- Unknown resource groups and invalid passwords now return clear authentication errors.
- Authentication failures are handled consistently and stop affected runtime activity.

## Tests
- Added coverage for valid credentials, unknown groups, incorrect passwords, and resource-group associations.


components/spider-storage/src/state/service.rs
```diff
@@ -641,7 +641,7 @@ impl<
     ///
     /// # Returns
     ///
-    /// The ID of the registered execution manager on success.
+    /// The ID of the registered execution manager and its resource group ID on success.
```
LinZhihao-723:
```suggestion
    /// A tuple on success, containing:
    ///
    /// * The ID of the registered execution manager.
    /// * The ID of the resource group to which the execution manager is dedicated, if any.
```

components/spider-storage/src/db/protocol.rs
```diff
@@ -350,20 +350,26 @@ pub trait ExecutionManagerLivenessManagement: Clone + Send + Sync {
     /// # Parameters
     ///
     /// * `ip_address` - The execution manager IP address.
+    /// * `resource_group_credentials` - The external ID and password of the resource group to bind
+    ///   the execution manager to, or [`None`] for a general execution manager.
     ///
     /// # Returns
     ///
-    /// The ID of the registered execution manager on success.
+    /// The ID of the registered execution manager and its resource group ID on success.
```
```suggestion
    /// A tuple on success, containing:
    ///
    /// * The ID of the registered execution manager.
    /// * The ID of the resource group to which the execution manager is dedicated, if any.
```

components/spider-storage/src/db/protocol.rs
```diff
@@ -350,20 +350,26 @@ pub trait ExecutionManagerLivenessManagement: Clone + Send + Sync {
...
     ///
+    /// * [`DbError::ExternalResourceGroupNotFound`] if the external resource group ID does not
+    ///   exist.
+    /// * [`DbError::InvalidPassword`] if the resource group password is incorrect.
     /// * Forwards [`sqlx::error::Error`] on DB operation failure.
     async fn register_execution_manager(
         &self,
         ip_address: IpAddr,
-    ) -> Result<ExecutionManagerId, DbError>;
+        resource_group_credentials: Option<(&str, &[u8])>,
```
LinZhihao-723: How about adding a dedicated struct type, `ExternalResourceGroupCredentials`, to hold the ID and the password?
We should then replace `ResourceGroupManagement::add`'s input with this credentials type.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -707,12 +732,16 @@ const fn execution_managers_creation_query() -> &'static str {
 CREATE TABLE IF NOT EXISTS `{EXECUTION_MANAGERS_TABLE_NAME}` (
   `id` BIGINT UNSIGNED NOT NULL AUTO_INCREMENT,
   `ip_address` VARCHAR(45) NOT NULL,
+  `resource_group_id` BIGINT UNSIGNED,
```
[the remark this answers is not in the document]
LinZhihao-723: @sitaowang1998 We should probably make this a breaking change (we should also mark the protocol change PR as breaking but we forgot to...)

components/spider-storage/src/grpc.rs
```diff
@@ -295,6 +295,7 @@ impl<
     ///
     /// The [`Status`] to send to the client:
     ///
+    /// * `UNAUTHENTICATED` for an unknown resource group or an invalid password.
```
LinZhihao-723, line 299: For completeness, I think we should update https://github.com/y-scope/spider/blob/21b50db3d65f4c29277796293bd0642e30ec02fc/components/spider-execution-manager/src/client/grpc/liveness.rs#L99 as well to handle the unauth error.

LinZhihao-723 (changes requested):

components/spider-execution-manager/src/client/liveness.rs
```diff
@@ -24,6 +24,10 @@ pub enum LivenessResponseError {
     #[error("execution manager already marked dead")]
     MarkedDead,
 
+    /// The external resource group id rejected by storage.
+    #[error("external resource group id rejected: `{0}`")]
+    IllegalExternalResourceGroupId(String),
```
LinZhihao-723: This error description is not correct: the failure might not be caused by an illegal RG ID, but a wrong password. We should:
- Rename it to `ResourceGroupAuth`.
- Delete the `String` field: there's no need to carry the RG in the error.

components/spider-execution-manager/src/client/grpc/liveness.rs
```diff
@@ -104,6 +104,29 @@ fn status_to_error(status: &Status) -> LivenessResponseError {
...
+) -> LivenessResponseError {
+    match (status.code(), external_resource_group_id) {
+        (Code::Unauthenticated, Some(external_resource_group_id)) => {
+            LivenessResponseError::IllegalExternalResourceGroupId(
+                external_resource_group_id.to_owned(),
+            )
+        }
+        _ => status_to_error(status),
+    }
+}
```
LinZhihao-723: We don't need RG ID in the error. Let's drop this method and move it to `status_to_error`.

components/spider-execution-manager/src/liveness.rs
```diff
@@ -156,6 +156,15 @@ impl<LivenessClientType: LivenessClient + Clone> LivenessActor<LivenessClientTyp
...
             }
+            Err(LivenessResponseError::IllegalExternalResourceGroupId(
+                external_resource_group_id,
+            )) => {
+                tracing::error!(
+                    external_resource_group_id,
+                    "Storage rejected the external resource group ID. Cancelling the runtime."
+                );
+                self.cancellation_token.cancel();
+            }
```
LinZhihao-723: Shouldn't this be unreachable from a heartbeat request? We should explicitly log this is unexpected and shutdown the runtime.

components/spider-execution-manager/src/liveness.rs
```diff
@@ -349,6 +358,26 @@ mod tests {
...
+            Arc::clone(&client),
+            SessionTracker::new(0),
+            cancellation_token.clone(),
+        );
+
+        tokio::time::timeout(Duration::from_secs(1), cancellation_token.cancelled())
+            .await
+            .expect("token was not cancelled within 1s");
+        join_actor(join).await;
+    }
```
LinZhihao-723: Please check the test semantics carefully. This scenario doesn't make sense, as the actor should never receive an illegal RG ID error.

components/spider-storage/src/db/protocol.rs
```diff
@@ -24,6 +24,15 @@ pub struct RecoverableJobContext {
...
 }
 
+/// Credentials identifying and authenticating an external resource group.
+pub struct ExternalResourceGroupCredentials {
+    /// The external resource group ID.
+    pub external_resource_group_id: String,
+
+    /// The resource group password.
+    pub password: Vec<u8>,
+}
```
LinZhihao-723, line 34: This might be out of the scope for this PR, but I think we should do the following in a follow up PR:
- Move this type to core. This should be how users provide their RG credentials through the public client API.
- Use this type to rewrite the public API in `spider-client`.
- Define this type in the gRPC protocol so that it is shipped as a bundle, not two separate fields.

If you agree, let's ask code rabbit to create an issue.

LinZhihao-723 (changes requested):

components/spider-storage/src/db/protocol.rs
sitaowang1998: @coderabbit Please open an issue about this.

kirkrodrigues: @coderabbitai Open an issue about this.

components/spider-execution-manager/src/client/grpc/liveness.rs
```diff
@@ -95,11 +95,13 @@ impl LivenessClient for GrpcLivenessClient {
 ///
 /// * [`LivenessResponseError::MarkedDead`] for `FAILED_PRECONDITION`.
 /// * [`LivenessResponseError::IllegalId`] for `INVALID_ARGUMENT`.
+/// * [`LivenessResponseError::ResourceGroupAuth`] for `UNAUTHENTICATED`.
 /// * [`LivenessResponseError::Transport`] for any other code.
```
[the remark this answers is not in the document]
LinZhihao-723: Those were already documented.

LinZhihao-723 (approved):

PR #466, opened 2026-08-31
feat(spider-scheduler): Parameterize the shared inbound-queue reader over a per-lane result formatter.

LinZhihao-723:
# Description

The `resource_group_round_robin` core polls the same three inbound-queue lanes as the round-robin core, but it does not consume the drained entries in the same shape: it schedules a resource group's jobs one batch at a time, so it needs the regular-task lane grouped by job and both finalization lanes reduced to the jobs they name. That reshaping is pure — it touches no scheduler state — so there is no reason for it to run on the core's critical path.

This PR adds the extension point that lets a core say what shape it wants, and moves the reshaping into the background polling task. It is a refactor: no behaviour changes, and the round-robin core is functionally untouched. The first formatter that actually reshapes anything lands in a follow-up PR.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **New Features**
  * Added support for transforming inbound queue results independently across regular and finalization processing lanes.
  * Existing behaviour remains available by default, while enabling consumers to receive lane-specific result formats.

* **Refactor**
  * Improved inbound polling result handling for greater flexibility without changing scheduling behaviour.

### DIFF
--- a/components/spider-scheduler/src/core_impl/inbound_queue_reader.rs
+++ b/components/spider-scheduler/src/core_impl/inbound_queue_reader.rs
@@ -9,15 +9,69 @@ use crate::error::StorageClientError;
 use crate::storage_client::SchedulerStorageClient;
 use crate::types::InboundEntry;
 
+/// The reshaping a core applies to the entries drained from each inbound-queue lane.
+///
+/// The formatting runs inside the lane's background polling task, so it is kept off the core's
+/// critical path.
+pub(super) trait InboundPollResultFormatter: Send + 'static {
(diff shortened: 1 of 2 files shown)


sitaowang1998 (approved):
LGTM. Only one high level question: why only one formatter for both commit and cleanup lanes?

LinZhihao-723:
> LGTM. Only one high level question: why only one formatter for both commit and cleanup lanes?

From the core's perspective, their inbound entries share the same texture. I can't imagine a case where they need different formatting so I decided to use one.

PR #469, opened 2026-09-01
refactor(spider-core)!: Unify external resource group credentials (resolves #468).

### LINKED ISSUE #468: Unify external resource group credentials across core, client, and gRPC
## Summary

Define external resource group credentials as a shared public contract.

## Required changes

- Move `ExternalResourceGroupCredentials` from `components/spider-storage/src/db/protocol.rs` to the core crate.
- Update the public `spider-client` API to accept and use this shared credentials type.
- Define the credentials as a single message in the gRPC protocol, instead of separate external resource group ID and password fields.
- Update storage, gRPC, and client call sites to use the shared contract.

## Rationale

The public client API and the gRPC protocol should use one credential type. This reduces duplicated credential fields and keeps the public API aligned with the internal storage contract.

## Affected areas

- Core types
- `spider-client` public API
- gRPC protocol definitions and generated bindings
- Storage service credential handling

## Acceptance criteria

- The shared credentials type is defined in the core crate.
- The public `spider-client` API accepts the shared credentials type.
- The gRPC protocol transports credentials as one message.
- Storage and gRPC call sites compile against the shared contract.
- Tests cover credential serialization and public client registration with resource group credentials.

## Backlinks

- Source PR: https://github.com/y-scope/spider/pull/464
- Source comment: https://github.com/y-scope/spider/pull/464#discussion_r3905917147
- Requested by: @kirkrodrigues

sitaowang1998:
# Description

This PR resolves #468.

Defines external resource group credentials in core to be shared by components.

## Changes

- Move `ExternalResourceGroupCredentials` from `spider-storage` to `spider-core`.
- Update `SpiderClient::add_resource_group` to accept the shared credentials type.
- Transport resource group credentials as a nested protobuf message.
- Add conversions between the core and protobuf credential types.
- Update storage and end-to-end call sites.
- Add tests.

## Breaking changes
The gRPC protocol changes as `AddResourceGroupRequest` now uses `ExternalResourceGroupCredentials`.

# Checklist

  breaking change.

# Validation performed

- [ ] GitHub workflows pass.

## Summary by CodeRabbit

- **New Features**
  - External resource-group credentials are now handled as a single structured value across registration workflows.
  - Credentials can be loaded from environment variables, with passwords protected more securely.
  - Resource-group and execution-manager registration now use consistent credential handling.
  - Credential access is available through dedicated methods rather than direct field access.

- **Bug Fixes**
  - Missing resource-group credentials now produce a clear invalid-argument error.

- **Tests**
  - Added coverage confirming credentials remain intact through protocol encoding and decoding.


sitaowang1998:
A high level question: Do we need to make `password` a `secrecy::Secret`?

LinZhihao-723:
> A high level question: Do we need to make `password` a `secrecy::Secret`?

I think it's worth doing so, as long as it can be easily supported to construct/reconstruct across the gRPC layer. Can you check this first?

sitaowang1998:
> > A high level question: Do we need to make `password` a `secrecy::Secret`?
> 
> I think it's worth doing so, as long as it can be easily supported to construct/reconstruct across the gRPC layer. Can you check this first?

My idea is to just use `secrecy::SecretSlice<u8>` for the password field of the `ExternalResourceGroupCredential` struct, which can easily support conversion to and from a normal `Vec<u8>`, and feed into gRPC layer. However, in protobuf buffer the password will still be stored the password in plain text. Given that the buffer is short-lived, and that gRPC transmission is encrypted, I would say it is safe to use `secrecy` just for the credential struct.

LinZhihao-723:
> SecretSlice<u8>

Do you think it makes more sense to use `SecretString` instead? It seems to be common to assume the password/API-secret is a string. For example, aws credentials are secret strings.

components/spider-core/src/types/resource_group.rs
```diff
@@ -0,0 +1,87 @@
+//! Resource group types shared across Spider components.
+
+use secrecy::ExposeSecret;
+use secrecy::SecretSlice;
+
+/// Environment variable that supplies the external resource group ID.
+pub const EXTERNAL_RESOURCE_GROUP_ID_ENV: &str = "SPIDER_EXTERNAL_RESOURCE_GROUP_ID";
+
+/// Environment variable that supplies the external resource group password.
+pub const EXTERNAL_RESOURCE_GROUP_PASSWORD_ENV: &str = "SPIDER_EXTERNAL_RESOURCE_GROUP_PASSWORD";
```
LinZhihao-723: How about dropping `external`?
- `EXTERNAL_RESOURCE_GROUP_PASSWORD_ENV` -> `RESOURCE_GROUP_PASSWORD_ENV`
- `SPIDER_EXTERNAL_RESOURCE_GROUP_PASSWORD` -> `SPIDER_RESOURCE_GROUP_PASSWORD`

components/spider-storage/src/db/mod.rs
```diff
@@ -7,9 +7,9 @@ pub use mariadb::MariaDbStorageConnector;
 pub use protocol::DbStorage;
 pub use protocol::ExecutionManagerLivenessManagement;
 pub use protocol::ExternalJobOrchestration;
-pub use protocol::ExternalResourceGroupCredentials;
 pub use protocol::InternalJobOrchestration;
 pub use protocol::RecoverableJobContext;
 pub use protocol::ResourceGroupManagement;
 pub use protocol::SchedulerRegistrationManagement;
 pub use protocol::SessionManagement;
+pub use spider_core::types::resource_group::ExternalResourceGroupCredentials;
```
LinZhihao-723: We shouldn't re-export this mod in `spider-storage`.

```suggestion
```

LinZhihao-723 (changes requested):
Overall lgtm.

components/spider-core/src/types/resource_group.rs
sitaowang1998: I agree with changing the env name, but not the variable name. Inside Spider we need to distinguish user provided resource group id and password against internal resource group id.

LinZhihao-723: I'm not suggesting to rename the first en var. I don't think we need to differentiate "external" and "internal" passward since there's only one password, right?

LinZhihao-723 (approved):
- Directly modified the PR title: we should use `resolves` for non-bug-fix issues.
- We need to mention that the issue is resolved in the PR description to close it since GH changed its behavior...

PR #470, opened 2026-09-02
feat(spider-scheduler): Add the resource-group-round-robin scheduler core implementation.

LinZhihao-723:
# Description

This is the core the four preceding PRs were building towards: the job registry (#444), the dispatch queue (#449), the per-resource-group scheduling state (#462), and the inbound-poll result formatter (#466, #467). It lands the tick loop that drives them, which retires the last of the `dead_code` expectations those PRs left behind.

The core is a single-threaded, tick-based loop scheduling at two levels of round robin: the outer level interleaves resource groups, the inner level interleaves the active jobs within each group. It owns all of the state it decides with — job entries in a generational arena, per-group scheduling states in an append-only vector, and the dispatch queue registry — so nothing it holds across an await point is thread-bound and the loop's future is `Send`.

The core is not wired up: it does not implement `SchedulerCore`, the scheduler config has no variant selecting it, and `DispatchQueueHandle` is unchanged. Nothing reaches it and no behaviour changes. Wiring it up needs a dispatch-queue handle that can express the pinned-versus-general split, which is its own PR.

## The tick

Five steps, in order. Steps 2 and 3 are skipped while a storage poll is still in flight; steps 4 and 5 always run, so the dispatch queues keep being refilled from already-buffered tasks while the core waits on storage.

1. Collect the inbound poll's results and drain the reschedule queue. A storage session newer than the tracked one is applied here, and nothing later in the tick branches on it.
2. Fold the results into the finalized job table, the dedup set, and the job registry. Finalizations are processed before regular tasks, so a regular task arriving in the same batch as its job's finalization is discarded rather than scheduled.
3. Apply the per-group updates, creating a group's scheduling state on first use and activating every group the updates touch.
4. Refill the dispatch queues under the admission policy. This is the scheduling decision, and the section below covers it.
5. Retire the jobs that ran dry.

## The admission policy is where the fairness comes from

A group may be admitted more work only while its own queue occupancy is below the buffer's current free space. That is the dynamic queue-length threshold of Choudhury & Hahne, and two properties of the implementation are what make it mean anything: free space is recomputed by decrement on every decision rather than once per tick, and admission is interleaved across groups one assignment at a time.

Both matter more than they look. Filling one group to its threshold before considering the next produces a staircase rather than an equilibrium — with a 256-slot buffer and five backlogged groups, batch-filling yields `64, 64, 64, 64, 0` and no free space, while quantum-1 rotation yields five groups at 42 with 41 free. Every threshold in the batch case was read correctly at the moment it was read; the outcome is still wrong, because a threshold is only meaningful when every group is measured against the same buffer state. Two of the tests assert on the resulting occupancies rather than on code paths, which is what makes them worth reading.

## What this deliberately does not do

* No `impl SchedulerCore`, no `SchedulerConfig` variant, no change to `DispatchQueueHandle`, and no change to the round-robin core. The module stays unreachable from configuration for one more PR.
* The dispatch service that serves pinned and general execution managers, and the gRPC plumbing that lets an execution manager name its resource group, both land later. The dispatch queues' read side is still `dead_code`-gated for that reason.
* `SHARING_COEFFICIENT` and `DOWNGRADE_LIVES` stay compile-time constants.

# Checklist

  breaking change.

# Validation performed

* [x] Add unit tests to assert the scheduler core's behavior, driving `tick` directly and draining the group queues by hand to exercise the real dispatch structures:
    * **Rotation** — the arm persists across ticks; dropping an exhausted group does not skip the group `swap_remove` moves into its slot.
    * **Activation** — a group that is exhausted but still holds assignments stays active.
    * **Admission** — one tick leaves every backlogged group at the dynamic threshold; no group is batch-filled while another waits; a newly active group is admitted against a backlogged incumbent; a lone group takes no more than half the buffer.
    * **Polling** — dispatch and retirement continue while a storage poll is in flight; each lane's fetch count is sized from the per-lane counters.
    * **Sessions** — a bump clears the dedup set and the finalized job table, zeroes every lane counter, and re-admits the tasks storage replays.
    * **Finalization** — a cleanup is scheduled after the same job committed; a repeated finalization is scheduled once; an expired entry leaves the table while a fresh one stays; an expired finalization re-admits the job's later tasks.
    * **Failure** — a closed dispatch queue and a closed broadcast queue each fail the tick; the loop stops when cancelled.
    * **Rescheduling** — an assignment returned by a lost execution manager is re-admitted.

## Summary by CodeRabbit

- **New Features**
  - Added resource-group-aware round-robin scheduling for balanced task dispatch across active groups.
  - Added lane-aware buffering and polling for ready, commit-ready, and cleanup-ready work.
  - Added support for task rescheduling, cancellation, deduplication, expiration, and finalized-job cleanup.
  - Added session handling that safely resets scheduling state and permits replayed tasks to be admitted.
- **Bug Fixes**
  - Improved handling of closed dispatch queues and in-flight storage polls.
  - Adjusted polling limits dynamically based on lane capacity and queued work.
- **Tests**
  - Added coverage for finalization, duplicate task prevention, batching, stale-session handling, and group reactivation.

### DIFF
--- a/components/spider-scheduler/src/core_impl/resource_group_round_robin/implementation.rs
+++ b/components/spider-scheduler/src/core_impl/resource_group_round_robin/implementation.rs
@@ -3,9 +3,11 @@
 
 use std::collections::HashMap;
 use std::collections::HashSet;
+use std::collections::VecDeque;
 use std::num::NonZeroU64;
 use std::num::NonZeroUsize;
 use std::time::Duration;
+use std::time::Instant;
 
 use serde::Deserialize;
 use spider_core::session::SessionTracker;
@@ -58,16 +60,16 @@ pub(super) struct RgRoundRobinConfig {
     /// fill the inbound-queue reading request.
     pub(super) storage_poll_timeout_ms: u64,
 
-    /// The time (in milliseconds) that the scheduler will spend on each tick.
+    /// The time (in milliseconds) that the scheduler will spend on each tick. If the tick spends
+    /// less than the configured interval, the core will sleep for the remainder.
     pub(super) tick_interval_ms: NonZeroU64,
+
+    /// The time (in seconds) that a job may remain in the finalized job table before the scheduler
+    /// drops it from the table.
+    pub(super) finalized_job_expiration_timeout_sec: u64,
 }
 
 /// The resource-group-aware round-robin scheduler core created from a [`RgRoundRobinConfig`].
-///
-/// The core owns all of the state it decides with -- job entries in a generational arena, per
-/// resource group scheduling states in an append-only vector, and the dispatch queue registry the
-/// execution-manager-facing service reads from -- so nothing it holds across an await point is
-/// thread-bound.
 ///
 /// # Type Parameters
 ///
@@ -78,8 +80,12 @@ pub(super) struct RgRoundRobinConfig {
 /// All member variables are marked `pub(super)` to allow the test module to inspect the internal
 /// states.
 pub(super) struct RgRoundRobin<SchedulerStorageClientType: SchedulerStorageClient + 'static> {
-    pub(super) global_task_set: HashSet<(JobId, TaskId)>,
+    pub(super) global_task_set: GlobalTaskSet,
     pub(super) finalized_jobs: HashSet<JobId>,
+
+    /// The insertion time of every job in [`Self::finalized_jobs`], in insertion order.
+    pub(super) finalized_job_queue: VecDeque<(JobId, Instant)>,
+
     pub(super) job_registry: JobRegistry,
 
     /// The scheduling states of every resource group the core has seen this session.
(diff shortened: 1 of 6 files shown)


components/spider-scheduler/src/core_impl/resource_group_round_robin/implementation.rs
LinZhihao-723, line 1: Checking off the implementation.

components/spider-scheduler/src/core_impl/resource_group_round_robin/tests.rs
LinZhihao-723, line 1: This file is not ready for the reviewer yet. Will take some more time to polish it tmr.

sitaowang1998 (dismissed):
The implementation looks good.

sitaowang1998 (dismissed):

LinZhihao-723 (commented):
Self-reviewed the test case implementation. Now checking off this PR.

sitaowang1998 (approved):

PR #471, opened 2026-09-09
feat(spider-scheduler): Wire the execution manager's resource group from the next-task request down to the dispatch queue.

LinZhihao-723:
# Description

An execution manager may be pinned to a resource group, and `NextTaskRequest` has carried an optional `resource_group_id` since #456 — but the scheduler rejected it outright, so a pinned execution manager had no way to say which group's assignments it wanted. This PR carries that field from the wire to the dispatch queue.

Nothing serves a pinned request yet. The resource-group-aware core (#470) owns a dispatch queue registry that can, but it does not implement `SchedulerCore` and no configuration selects it, so today every scheduler still runs the single-queue core. What changes here is that the request now reaches the layer where that decision belongs, and the single-queue implementation answers for itself rather than the protocol layer refusing on its behalf.

## What moves

`NextTaskRequest::unpack` stops returning `Code::Unimplemented` for `resource_group_id` and produces a `NextTaskRequestPayload` instead of a tuple. The gRPC handler destructures it and passes the group through `SchedulerServiceState::next_task` into `DispatchQueueHandle::dequeue`, which gains the group as a parameter.

```
NextTaskRequest --unpack--> NextTaskRequestPayload --> next_task --> dequeue
```

The tuple became a struct because the unpacked form reached four fields, two of them `Option`s: at that width, positional returns stop being readable at the call site and start being a place to swap two arguments of the same type.

# Checklist

  breaking change.

# Validation performed

  * This ensures the e2e test can pass, meaning that the current system is unaffected.

## Summary by CodeRabbit

- **New Features**
  - Task requests can now include an optional resource group and previous assignment details.
  - Resource group identifiers are preserved and passed through scheduling requests.
  - Scheduler services support resource-group-aware task retrieval where available.

- **Bug Fixes**
  - Unsupported resource-group requests are now reported clearly instead of failing with an unimplemented error.
  - Queued assignments remain available after a request is rejected as unsupported.


components/spider-proto-rust/src/unpack/scheduler.rs
```diff
@@ -62,3 +69,58 @@ impl RequestUnpack for ShutdownRequest {
         ))
     }
 }
+
+#[cfg(test)]
+mod tests {
```
sitaowang1998: Do we really need unit test for this?

sitaowang1998 (dismissed):

components/spider-proto-rust/src/unpack/scheduler.rs
LinZhihao-723: Yeah I also doubt we need this but forgot to remove it. It's agent-generated. My bad.

sitaowang1998 (approved):

PR #472, opened 2026-09-10
feat(spider-scheduler): Implement `DispatchQueueHandle` for the resource-group-round-robin scheduler core's dispatch queue.

LinZhihao-723:
# Description

This PR depends on #471.

This PR implements `DispatchQueueHandle` for resource-group-round-robin scheduler core's dispatch queue on top of `DispatchQueueRegistry`. The following major changes are made:

- Add non-blocking methods for popping a hint from the global hint broadcast queue and popping assignments from per-rg dispatch queue.
- Implement the loop for pinned (for dedicated EMs) dispatching and non-pinned (for general EMs) assignment dispatching, with session validation logic.
- Implement `DispatchQueueHandle` interface for `DispatchQueueRegistry` so that it can be used to serve EMs' requests.
- Reorder symbols according to the visibility change. Some of the methods will no longer be needed outside of the mod so they're made private.

# Checklist

  breaking change.

# Validation performed

## Summary by CodeRabbit

* **Improvements**
  * Improved task dispatching for both resource-group-specific and general scheduling requests.
  * Stale task assignments are now ignored when scheduling sessions change, helping prevent outdated work from being dispatched.
  * Scheduling continues to work correctly after session updates and when queues are empty or closed.
  * Zero-wait requests now return promptly without unnecessary waiting.

* **Tests**
  * Added coverage for task dispatching, session changes, empty and closed queues, and no-wait behaviour.


components/spider-scheduler/src/core_impl/resource_group_round_robin/dispatch_queue.rs
```diff
@@ -361,6 +243,230 @@ impl DispatchQueueRegistry {
...
+    ///
+    /// # Returns
+    ///
+    /// The next assignment of `rg_id`, or [`None`] if none became available within `wait_time`.
+    async fn dequeue_pinned(
+        &self,
+        rg_id: ResourceGroupId,
+        wait_time: Duration,
+    ) -> Option<TaskAssignment> {
+        let deadline = tokio::time::Instant::now() + wait_time;
```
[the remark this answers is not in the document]
LinZhihao-723: I don't think the current implementation can peacefully handle this. A more systematic solution would be:
- Cap the wait time: the service level should validate the largest timeout is within a certain range.
- If an overflow is detected through `checked_add`, just set it to `now`.

Can you create an issue to keep track of this?

components/spider-proto-rust/src/unpack/scheduler.rs
```diff
@@ -69,3 +69,58 @@ impl RequestUnpack for ShutdownRequest {
...
+        };
+
+        let payload = request.unpack().expect("the request should unpack");
+
+        assert_eq!(payload.em_id, ExecutionManagerId::from(EM_ID));
+        assert_eq!(payload.prev_assignment, None);
+        assert_eq!(payload.rg_id, None);
+        assert_eq!(payload.wait_time, Duration::from_millis(WAIT_TIME_MS));
+    }
+}
```
LinZhihao-723:
```suggestion
```
As suggested in the previous PR, let's remove it.

components/spider-scheduler/src/core_impl/resource_group_round_robin/dispatch_queue.rs
```diff
@@ -361,6 +243,230 @@ impl DispatchQueueRegistry {
...
+        loop {
+            // Re-fetched per loop iteration because a retry follows a stale-session assignment,
```
sitaowang1998, line 259: Why pass tense and why hyphen?

components/spider-scheduler/src/core_impl/resource_group_round_robin/dispatch_queue.rs
```diff
@@ -239,10 +115,7 @@ impl DispatchQueueRegistry {
     /// # Returns
     ///
     /// The read side of `rg_id`'s dispatch queue, creating the group if it has none.
-    pub(super) fn get_dispatch_queue_reader(
-        &self,
-        rg_id: ResourceGroupId,
-    ) -> RgDispatchQueueReader {
+    fn get_dispatch_queue_reader(&self, rg_id: ResourceGroupId) -> RgDispatchQueueReader {
```
20001020ycx: Although you have claimed this in the PR description, but it looks like the order (pub or private) of `DispatchQueueRegistry`'s methods are mixed up, see below for what my AI agent found:
```
  impl DispatchQueueRegistry {
      pub(super) fn new                        L103   pub
      fn get_dispatch_queue_reader             L118   private   ← private before public below
      pub(super) fn get_dispatch_queue_writer  L125   pub
      fn try_next_hint                         L138   private   ← private before public below
      async fn next_hint                       L164   private   ← private before public below
      pub(super) fn len                        L174   pub
      pub(super) fn clear                      L193   pub
      pub(super) fn close_broadcast_queue      L203   pub (cfg test)
      pub(super) fn close_dispatch_queue       L213   pub (cfg test)
      pub(super) fn num_outstanding_hints      L221   pub (cfg test)
      fn get_or_create                         L228   private
      async fn dequeue_pinned                  L252   private
      async fn dequeue_general                 L284   private
  }
```

20001020ycx (commented):
style nits by AI agent, looking into the functionality now

components/spider-scheduler/src/core_impl/resource_group_round_robin/dispatch_queue.rs
```diff
@@ -361,6 +243,230 @@ impl DispatchQueueRegistry {
...
+            if assignment.session_id != current_session_id {
+                // If the assignment is from a stale session, drop it.
+                continue;
+            }
+            return Some(assignment);
+        }
+    }
+}
+
+/// Serves both the execution managers pinned to a resource group and those that take work from any
```
20001020ycx, line 312: Let me confirm my understanding: I guess once the EM is available, it polls the work from grpc. And looks like you have a deadline logic in your polling method, so I wonder say if there's no task in the dispatch queue and it hits deadline, what would happen? does EM idle a while then call dequeue or the polling itself is achieved through this deadline logic.

20001020ycx, line 312: > those that take work from any

You mean the general EM right? 

> so that the registry is itself the queue the execution-manager-facing service drains.

Explanation is a bit convoluted. You are saying this is how EM is draining the dispatch queue right?

20001020ycx (commented):
I think I am clear on what the dispatch on different rg do. Will read other parts later to go through the docstring.

components/spider-scheduler/src/core_impl/resource_group_round_robin/dispatch_queue.rs
```diff
@@ -361,6 +243,230 @@ impl DispatchQueueRegistry {
...
+    }
+}
+
+/// One resource group's claim on a general execution manager, drawn on that group's hint count.
+///
+/// A hint can only come out of the broadcast queue: both the wrapped reader and the constructor are
+/// private, so no caller outside this module can turn a reader it happens to hold into a hint.
+/// Spending one consumes it, and it is deliberately neither [`Clone`] nor [`Copy`], so "a hint is
+/// spent at most once" is a property the type system enforces rather than a rule a call site
+/// follows. Dropping a hint unspent withdraws nothing from its group's count, which is what lets a
```
20001020ycx: The like the comment for this class, but I am confused on the last point, what do you mean by dropping a hint? I don't see a drop method implemented?

20001020ycx (dismissed):
LGTM, left some comments where I found a bit confusing to me

components/spider-scheduler/src/core_impl/resource_group_round_robin/dispatch_queue.rs
20001020ycx: Per our offline discussion, the polling mechanism is achieved through this deadline logic. It is blocked while deadline is not reached. Once the deadline hits and the control returns back to EM, it issues another RPC call to drain the dispatch queue, with the same deadline polling mechanism. 

And this polling method is better than native rpc call every X seconds, we can make this general in communication for the Spider. To understand the overall communication overhead, we might wanna to collect metrics for each rpc calls.

components/spider-scheduler/src/core_impl/resource_group_round_robin/dispatch_queue.rs
LinZhihao-723: I will update the docstring to explain this: the hint must be dropped on a session bump.

components/spider-scheduler/src/core_impl/resource_group_round_robin/dispatch_queue.rs
LinZhihao-723: Rewrote for clarity.

20001020ycx (dismissed):

components/spider-scheduler/src/core_impl/resource_group_round_robin/dispatch_queue.rs
LinZhihao-723: - Hyphen is a more standard form than "refetch"/"refetched" in tech writing (confirmed with different LLM models). For example, the k8s document uses "re-fetch": https://github.com/kubernetes-sigs/kustomize/blob/master/examples/chart.md?utm_source=chatgpt.com#how-does-the-pull-work
- This is a reduced passive clause, not pass tense. Rewrite it in the full term as "The reader is re-fetched per loop ..."

20001020ycx (approved):

PR #477, opened 2026-09-10
feat(spider-scheduler)!: Support the resource-group-round-robin in the scheduler service.

LinZhihao-723:
# Description

This is a breaking change because it changes the Docker Compose config layout.

This PR depends on #472 and #481.

This PR wires the new scheduler (resource-group-round-robin) in the scheduler service, meaning that it can be selected by the scheduler runtime through the scheduler config. Changes including:

- Make `ResourceGroupRoundRobinConfig` and `ResourceGroupRoundRobinCore` exposed from the `resource_group_round_robin`'s mod as public symbols.
- Add `ResourceGroupRoundRobinConfig` as a variant in the policy config enum.
- Deployment support:
  - Helm chart: Nothing needs to be changed. The current config already allows switching policy configs.
  - Docker Compose: Change the layout of the scheduler configs. Now the scheduler configs are replicated based on the policy name. The top-level compose file will select the correct config based on the env var `SPIDER_SCHEDULER_POLICY`.

Future PRs are needed to:

- Enable resource-group dedicated EMs.
- Set the resource-group-round-robin scheduler core as the default policy.

# Checklist

  breaking change.

# Validation performed

* [x] Passed e2e test locally by setting the scheduler to the resource-group-round-robin. Ensure all test cases pass and compare the e2e time to make sure it doesn't have an observable slowdown compared to the default policy.

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in resource-group round-robin scheduler with configurable capacity, timing, queue, and cleanup settings.
  - Scheduler requests now support resource group identifiers when provided.
  - Deployment configurations can select a scheduler policy through an environment setting, with round-robin as the default.
  - Added ready-to-use resource-group round-robin Compose configuration and example defaults.

- **Tests**
  - Added coverage for resource-group request handling and scheduler runtime startup and shutdown.


tools/deployment/spider-compose/configs/scheduler/resource_group_round_robin.yaml
```diff
@@ -0,0 +1,37 @@
...
+            active_job_list_capacity:
+              ${SPIDER_SCHEDULER_RESOURCE_GROUP_ROUND_ROBIN_ACTIVE_JOB_LIST_CAPACITY:-4}
+            cleanup_ready_task_capacity:
+              ${SPIDER_SCHEDULER_RESOURCE_GROUP_ROUND_ROBIN_CLEANUP_READY_TASK_CAPACITY:-256}
+            commit_ready_task_capacity:
+              ${SPIDER_SCHEDULER_RESOURCE_GROUP_ROUND_ROBIN_COMMIT_READY_TASK_CAPACITY:-256}
+            dispatch_queue_capacity:
+              ${SPIDER_SCHEDULER_RESOURCE_GROUP_ROUND_ROBIN_DISPATCH_QUEUE_CAPACITY:-16}
+            finalizing_job_expiration_timeout_sec:
+              ${SPIDER_SCHEDULER_RESOURCE_GROUP_ROUND_ROBIN_FINALIZING_JOB_EXPIRATION_TIMEOUT_SEC:-21600}
```
20001020ycx: Suggested by AI:
The default is set to 21600 while in round_robin.yaml, it is 300. Is this intentional? If so, which one should we follow

tools/deployment/spider-compose/configs/scheduler/resource_group_round_robin.yaml
```diff
@@ -0,0 +1,37 @@
...
+          port: ${SPIDER_SCHEDULER_PORT:-50052}
+        em_registry:
+          dead_em_cutoff_sec: ${SPIDER_SCHEDULER_EM_REGISTRY_DEAD_EM_CUTOFF_SEC:-30}
+          liveness_tracking_interval_ms:
+            ${SPIDER_SCHEDULER_EM_REGISTRY_LIVENESS_TRACKING_INTERVAL_MS:-1000}
+        scheduler:
+          policy: "resource_group_round_robin"
+          config:
+            active_job_list_capacity:
+              ${SPIDER_SCHEDULER_RESOURCE_GROUP_ROUND_ROBIN_ACTIVE_JOB_LIST_CAPACITY:-4}
```
20001020ycx: I assume this is a similar one to round_robin's `active_job_queue_capacity`. Not only the name is different now, and the default is reduced to 4 rather than 16. Is this intentional? If so, which one should we follow.

20001020ycx (commented):
Thank you so much for the change in docker compose, this was indeed something we had missed for the initial implementation. And I found the approach simple and elegant, really appreciate the effort.

I have reviewed all files for docker compose and taskfiles and it LGTM overall in this round, left some questions and suggestions for inconsistencies

Moreover, I have a question while reading your PR description. You said that
> Future PRs are needed to: … Set the resource-group-round-robin scheduler core as the default policy.

This is something which you already did in this PR. Therefore, I wonder, what is the default policy that we shall abide?

components/spider-scheduler/src/core_impl/resource_group_round_robin/dispatch_queue.rs
```diff
@@ -10,140 +10,16 @@ use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
 use std::time::Duration;
 
+use async_trait::async_trait;
```
20001020ycx, line 13: This is the change in PR #472, can you ask AI to exclude the changes for easier review? I found AI to be effective in these kind of things, and it is very reliable.

components/spider-scheduler/src/core_impl/resource_group_round_robin/tests.rs
```diff
@@ -273,17 +276,18 @@ impl CoreFixture {
...
     /// The assignments taken, in dispatch order.
     async fn drain_reader(&self, rg_id: ResourceGroupId) -> Vec<TaskAssignment> {
-        let reader = self
-            .dispatch_queue_registry
-            .get_dispatch_queue_reader(rg_id);
         let mut assignments = Vec::new();
-        while let Some(assignment) = reader.recv_pinned(Duration::ZERO).await {
+        while let Ok(Some(assignment)) = self
+            .dispatch_queue_registry
+            .dequeue(Some(rg_id), Duration::ZERO)
```
20001020ycx, line 288: This is more of a question from my side: Why the duration is 0? isnt this make the polling implemented in 472 end instantly? If I understand correctly, this means drain immediately without any looping in dequeue_pinned. Would this result in more traffic than we anticipated?

components/spider-scheduler/src/core_impl/resource_group_round_robin/mod.rs
```diff
@@ -4,35 +4,137 @@
...
+use crate::core::SchedulerCore;
+use crate::core::TaskAssignmentIdIssuer;
+use crate::dispatch_queue::SharedDispatchQueueHandle;
+use crate::error::SchedulerError;
+use crate::storage_client::SchedulerStorageClient;
+use crate::types::TaskAssignment;
+
+/// The configuration of the resource-group-aware round-robin scheduler core.
+#[derive(Clone, Debug, Deserialize)]
+pub struct ResourceGroupRoundRobinConfig {
```
20001020ycx, line 37: I am curious in why we define this here? In round robin implementation, it seems to be that mod.rs is just using the config and core for round robin. But here you are defining and exposing them at mod.rs for rgrr. And at upper layer's mod.rs, you use
```
pub use resource_group_round_robin::ResourceGroupRoundRobinConfig;
pub use resource_group_round_robin::ResourceGroupRoundRobinCore;
```
over
```
pub use round_robin::*;
```

Shall we make them consistent or is there any issue tracking this?

20001020ycx (commented):
Reviewed all files except dispatch_queue.rs since it is mostly the changes from pr 472. Left a few questions for the other parts.

tools/deployment/spider-compose/configs/scheduler/resource_group_round_robin.yaml
LinZhihao-723: Sorry, I haven't checked the generated default values yet. I don't think they're set properly yet. I will fix them in a future commit.

tools/deployment/spider-compose/configs/scheduler/resource_group_round_robin.yaml
LinZhihao-723: - The renaming is intentional. I think it would be better to not call it a "queue" since it's not strictly FIFO.
- The default is a problem. I will fix it.

components/spider-scheduler/src/core_impl/resource_group_round_robin/dispatch_queue.rs
LinZhihao-723: Sorry I don't fully understand. This PR depends on #472 otherwise it won't work. What do you mean "exclude the changes"?

20001020ycx: I understand that PR depends on #472, however, current PR has 700+ lines of changes, this is because it has the identical 500+ lines of change in PR 472. I think it will be clearer for the reviewer if you can exclude the identical change in PR 472, and only shows the actual diff of 200+.

The reason why I think this is a better practice is because I am not sure if there's anything related to PR 477 that needs to be done in dispatch_queue.rs. As you can see in my PR review, I didnt review this file because I assumed it is identical with the change in PR 472. However, this assumption can be wrong because an author may change some minor functionality in dispatch_queue.rs required for 477.

tools/deployment/spider-compose/configs/scheduler/resource_group_round_robin.yaml
20001020ycx: gotcha, shall we also update the round robin's naming convention in this PR?

components/spider-scheduler/src/core_impl/resource_group_round_robin/dispatch_queue.rs
LinZhihao-723: - Without #472, this PR would be incomplete:
  - It can't be tested end-to-end.
  - The core trait requires a method to return a dispatch queue handle; without #472 this method can't be implemented.
- If you need a clean diff, I think a better way is to use the IDE's diff mode, which compares #477's branch against #472's branch.

components/spider-scheduler/src/core_impl/resource_group_round_robin/mod.rs
LinZhihao-723: The old scheduler's code is not well structured. The current scheduler's complexity is much higher, so we came up with a more detailed internal code layout. In particular, to answer your question: the `implementation.rs` is already complicated enough, so I think it's better to split the top-level oriented types from the implementation file.
On the other hand, I don't think it's necessary to enforce the same layout for all scheduler implementations, though: from the parent mod's view, there should be only two public types exported: `XXXConfig` and `XXXCore`. It doesn't matter where the types are defined. This gives enough flexibility for different schedulers to lay out their code based on their complexity.

components/spider-scheduler/src/core_impl/resource_group_round_robin/tests.rs
LinZhihao-723: This is just used by the tester to quickly drain everything from the dispatch queue.

20001020ycx: My bad

components/spider-scheduler/src/core_impl/resource_group_round_robin/dispatch_queue.rs
20001020ycx: gotcha, sure

tools/deployment/spider-compose/configs/scheduler/resource_group_round_robin.yaml
LinZhihao-723: I think it's ok for now. We will probably deprecate the old scheduler at some point.

20001020ycx (approved):

LinZhihao-723:
The wolf-tests CI failed but it's unrelated to the changes introduced in this PR.

PR #485, opened 2026-09-15
feat(spider-helm)!: Deploy execution-manager workers as resource-group pools.

20001020ycx:
# Description

This PR implements the Helm side of the resource-group-aware scheduling design. It replaces the single type worker `Deployment` (`spiderConfig.worker.replicas`) with a list of pools (`spiderConfig.worker.pools`), each rendered as its own `Deployment`.

> [!NOTE]
> This PR is a breaking change: `spiderConfig.worker.replicas` is replaced by `spiderConfig.worker.pools`.

Below is how a subchart user overrides `values.yaml` to configure two pools: a general pool (no resource group) and a pool dedicated to a resource group, whose credential is validated when the worker registers. The general pool is what the chart ships as the default `values.yaml`.

```yaml
spiderConfig:
  worker:
    pools:
      - name: "general"            # general: runs any group's tasks (chart default)
        replicas: 4
        resource_group: null
      - name: "search"             # dedicated: runs only this group's tasks
        replicas: 2
        resource_group:
          external_id: "clp-search"
          password:
            secretKeyRef: {name: "spider-rg-credentials", key: "clp-search"}
```

The `values.yaml` above is used throughout the validation section below.

# Checklist

  breaking change.

# Validation performed

## 1. Render: default is one general pool; a dedicated pool carries an independent, Secret-backed credential

From `tools/deployment/spider-helm`, the default renders a single general worker `Deployment` with no resource-group env vars:

```bash
helm template sp . | grep -E "name: sp-spider-worker"            # -> name: sp-spider-worker-general
helm template sp . | grep -c "SPIDER_EXTERNAL_RESOURCE_GROUP_ID"  # -> 0
```

A general + dedicated overlay renders one `Deployment` per pool; only the dedicated pool's container gets the RG env vars, and its password is a `secretKeyRef`:

```bash
cat > /tmp/pools.yaml <<'EOF'
spiderConfig:
  worker:
    pools:
      - {name: general, replicas: 2, resource_group: null}
      - name: search
        replicas: 3
        resource_group:
          external_id: "clp-search"
          password:
            secretKeyRef: {name: spider-rg-credentials, key: clp-search}
EOF
helm template sp . -f /tmp/pools.yaml | grep -E 'sp-spider-worker-|SPIDER_.*RESOURCE_GROUP|secretKeyRef'
```

Expected — two worker Deployments, and the dedicated (`search`) pool's password sourced from the Secret (never a literal):

```yaml
- name: "SPIDER_EXTERNAL_RESOURCE_GROUP_ID"
  value: "clp-search"
- name: "SPIDER_RESOURCE_GROUP_PASSWORD"
  valueFrom:
    secretKeyRef: { name: "spider-rg-credentials", key: "clp-search" }
```

## 2. E2E: a dedicated pool registers under its group with an independent, Secret-sourced password

0. Environments — a kind cluster (v0.24.0); `storage`/`scheduler:main` from ghcr; a worker image built from #484; and the RG-password Secret. The chart only *references* the Secret via `secretKeyRef` — a parent chart (e.g. CLP) renders it from a Secret template, while here we create it directly:

```bash
kubectl create secret generic spider-rg-credentials --from-literal=clp-search='s3cr3t-search-pw'
```

1. Install the Helm chart

```bash
helm install sp . -f values.yaml
```

2. Each worker in the pool has its own `Deployment` wrt to its resource group:

```bash
kubectl get deploy | grep sp-spider-worker
```

```text
sp-spider-worker-general   1/1   ...
sp-spider-worker-search    1/1   ...
```

3. The dedicated pool's password is a `secretKeyRef` (not a literal); the general pool has no RG env vars:

```bash
kubectl get pod -l app.kubernetes.io/component=worker-search -o yaml \
  | grep -A5 'name: SPIDER_RESOURCE_GROUP_PASSWORD'
# -> valueFrom: secretKeyRef: {name: spider-rg-credentials, key: clp-search}
```

4. Before the group exists, the dedicated worker fails fast (design §10) and restarts:

```bash
kubectl logs -l app.kubernetes.io/component=worker-search --tail=5 | grep -o '"error":"[^"]*"'
# -> "error":"failed to register with storage: resource group authentication failed"
```

5. Create the group with the **same** `(external_id, password)` as the Secret, then let the worker retry:

```bash
kubectl exec sp-spider-database-0 -- mariadb --table -u spider-user -pspider-password spider-db \
  -e 'SELECT id, resource_group_id FROM execution_managers ORDER BY id;
      SELECT id, external_id, CAST(password AS CHAR) AS password FROM resource_groups;'
```

```text
+----+-------------------+     +----+-------------+------------------+
| id | resource_group_id |     | id | external_id | password         |
+----+-------------------+     +----+-------------+------------------+
|  1 |              NULL |     |  1 | clp-search  | s3cr3t-search-pw |
|  2 |                 1 |     +----+-------------+------------------+
+----+-------------------+     (EM 1 = general pool; EM 2 = search pool)
```

The table above shows that a dedicated pool's worker is authenticated and successfully registered under its resource group.

## 3. E2E: a dedicated pool runs only its own group's tasks, and never another group's

0. Environments — a kind cluster and `storage`/`scheduler`/`worker:main` from ghcr. The overlay `values.yaml` deploys a `general` pool plus a pool dedicated to `rg-dedicated`, similar to above config shown in the PR description. For testing purpose, we do NOT create a `rg-other` worker pool to prove the correctness of Spider's functionality.

1. A new e2e case, `tests/huntsman/e2e/tests/isolation.rs`, submits 60 single-task jobs under `rg-dedicated` and 60 under `rg-other` (which has no dedicated pool) concurrently. Each job runs a new `report_worker_pool` task that returns the RG its worker is pinned to (or `general`), so every task's placement is read straight from the job outputs:

```bash
SPIDER_ENDPOINT=http://127.0.0.1:50051 SPIDER_CONCURRENCY=16 \
  cargo test --release --package e2e --test isolation -- --nocapture
```

2. The reported placement — the dedicated worker ran only `rg-dedicated` tasks, and never an `rg-other` task, while all 60 `rg-other` tasks ran on the general workers:

```text
(submitted_rg, executing_pool) -> count
(rg-dedicated, rg-dedicated)   -> 36    # dedicated worker ran its own group's tasks
(rg-dedicated, general)        -> 24    # general workers may also run rg-dedicated tasks
(rg-other,     general)        -> 60    # all rg-other tasks ran on general workers
(rg-other,     rg-dedicated)   ->  0    # dedicated worker ran NONE of another group's tasks
```

3. The storage DB corroborates the binding and the split:

```bash
kubectl exec iso-spider-database-0 -- mariadb --table -u spider-user -pspider-password spider-db \
  -e 'SELECT id, resource_group_id FROM execution_managers ORDER BY id;
      SELECT resource_group_id, COUNT(*) AS jobs FROM jobs GROUP BY resource_group_id;'
```

```text
+----+-------------------+     +-------------------+------+
| id | resource_group_id |     | resource_group_id | jobs |
+----+-------------------+     +-------------------+------+
|  1 |              NULL |     |                 1 |   60 |   (rg-dedicated)
|  2 |              NULL |     |                 2 |   60 |   (rg-other)
|  3 |                 1 |     +-------------------+------+
+----+-------------------+     (EM 3 = dedicated pool, bound to rg-dedicated; EMs 1-2 = general)
```

The dedicated worker (EM 3) executed only its own group's (`rg-dedicated`) tasks, and none of the 60 `rg-other` tasks that ran concurrently on the general workers — exactly the isolation a dedicated pool must provide.

## Summary by CodeRabbit

- **New Features**
  - Helm deployments now support multiple worker pools, with separate deployments and replica counts for each pool.
  - Worker pools can optionally use external resource groups, with credentials supplied through configured values or Kubernetes secrets.
  - The default configuration includes a general worker pool with four replicas.

- **Chores**
  - Updated the Helm chart version to `0.2.0-dev.1`.


tools/deployment/spider-helm/templates/worker-deployment.yaml
```diff
@@ -1,48 +1,65 @@
...
+              value: {{ .external_id | quote }}
+            - name: "SPIDER_RESOURCE_GROUP_PASSWORD"
+              {{- $password := .password | default dict }}
+              {{- if $password.secretKeyRef }}
+              valueFrom:
+                secretKeyRef:
+                  name: {{ $password.secretKeyRef.name | quote }}
+                  key: {{ $password.secretKeyRef.key | quote }}
+              {{- else }}
+              value: {{ required "a dedicated pool's `resource_group.password` needs `value` or `secretKeyRef`" $password.value | quote }}
```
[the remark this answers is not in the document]
20001020ycx: My agent said that this runtime change is reflected in $484, so this suggestion is not valid

tools/deployment/spider-helm/templates/worker-deployment.yaml
```diff
@@ -1,48 +1,65 @@
+{{- range .Values.spiderConfig.worker.pools }}
+---
 apiVersion: "apps/v1"
 kind: "Deployment"
 metadata:
-  name: {{ include "spider.componentFullname" (dict "root" . "component" "worker") }}
+  name: {{ include "spider.componentFullname" (dict "root" $ "component" (printf "worker-%s" .name)) }}
```
LinZhihao-723, line 6: `.name` is load-bearing here (it forms the `Deployment` name and all three label values), but nothing validates it, and the most natural way to scale a pool drops it:

```console
$ helm template sp . --set 'spiderConfig.worker.pools[0].replicas=9' | grep -E 'name: sp-spider-worker|component:'
  name: sp-spider-worker-%!s(<nil>)
    app.kubernetes.io/component: "worker-%!s(<nil>)"
```

`--set spiderConfig.worker.pools[0].replicas=N` replaces the whole list element, so `name` and `resource_group` vanish. That's the direct successor to the old `--set spiderConfig.worker.replicas=N`, and neither `helm template` nor `helm lint` fails on it (lint only warns), so the broken name reaches the API server.

Three more cases in the same value, all silent:

* Duplicate pool names render two `Deployment`s with the same `metadata.name` (I get `sp-spider-worker-dup` twice, with different `replicas`); whichever applies last wins.
* A pool name of 57+ characters pushes `app.kubernetes.io/component` past the 63-character label-value limit. `spider.componentFullname` truncates the resource name, but nothing truncates the label.
* A name such as `Search_1` renders a `metadata.name` that isn't a valid DNS-1123 subdomain.

Could we require and validate the name, e.g.

```
{{- $name := required "a worker pool needs `name`" .name }}
{{- if or (gt (len $name) 56) (not (regexMatch "^[a-z0-9]([-a-z0-9]*[a-z0-9])?$" $name)) }}
{{- fail (printf "worker pool name %q must be a lowercase RFC 1123 label of at most 56 characters" $name) }}
{{- end }}
```

plus a uniqueness check across `pools`? A `values.schema.json` would cover this and the two comments below in one place, if you'd prefer that route.

tools/deployment/spider-helm/templates/worker-deployment.yaml
```diff
@@ -1,48 +1,65 @@
...
   labels:
-    {{- include "spider.labels" . | nindent 4 }}
-    app.kubernetes.io/component: "worker"
+    {{- include "spider.labels" $ | nindent 4 }}
+    app.kubernetes.io/component: {{ printf "worker-%s" .name | quote }}
 spec:
-  replicas: {{ .Values.spiderConfig.worker.replicas }}
+  replicas: {{ .replicas }}
```
LinZhihao-723: A pool without `replicas` renders `replicas:` (null), which Kubernetes defaults to 1, so a pool meant for N workers silently runs one.

`required` only rejects nil and empty strings, so `replicas: 0` still renders `0` and scale-to-zero keeps working:

```suggestion
  replicas: {{ required "a worker pool needs `replicas`" .replicas }}
```

LinZhihao-723 (commented):
A few pre-merge items on the new `pools` values. All three are about a pool entry being accepted without validation; the rendering itself looks right: the default values, a general + dedicated overlay, the `secretKeyRef` path and the `required` guards on `password` all behave as the description says, and a full-chart diff against `main` shows no unintended changes outside the worker `Deployment`'s name and component labels.

tools/deployment/spider-helm/templates/worker-deployment.yaml
```diff
@@ -1,48 +1,65 @@
...
+          imagePullPolicy: {{ $.Values.image.worker.pullPolicy | quote }}
           command: ["spider-execution-manager", "--config", "/etc/spider/execution-manager.yaml"]
           env:
             - name: "RUST_LOG"
-              value: {{ .Values.spiderConfig.execution_manager.log_level | quote }}
-            {{- with .Values.spiderConfig.worker.extra_envs }}
+              value: {{ $.Values.spiderConfig.execution_manager.log_level | quote }}
+            {{- with .resource_group }}
+            - name: "SPIDER_EXTERNAL_RESOURCE_GROUP_ID"
+              value: {{ .external_id | quote }}
```
LinZhihao-723: `external_id` has no guard, unlike the three on `password` just below. A dedicated pool with a password but no `external_id` renders:

```yaml
            - name: "SPIDER_EXTERNAL_RESOURCE_GROUP_ID"
              value:
```

The worker then starts with an empty group ID and crash-loops on `UNAUTHENTICATED` from storage, instead of failing at render time the way the `password` cases do.

```suggestion
              value: {{ required "a dedicated pool's `resource_group` needs `external_id`" .external_id | quote }}
```

tools/deployment/spider-helm/templates/worker-deployment.yaml
20001020ycx: Agreed, although not all Spider config has guard, configuring this list of pools is error-prone, and let's make sure user configures correctly.

tools/deployment/spider-helm/templates/worker-deployment.yaml
20001020ycx: Ditto with above comment

tools/deployment/spider-helm/templates/worker-deployment.yaml
20001020ycx: Good catch, I spent some time investigating these cases and it turned out that point 2 and 3 will be eventually caught by k8s during helm install. Therefore, this wont be a silent error to user. So we dont need to implement additional guard to prompt user about such behaviour.

As for the point 1, I think we can explain its behavior in the doc: whichever applies last wins is a fine behavior. I dont think an explicit guard is required.

LinZhihao-723 (approved):
Before we merge, is it possible to have an e2e test by merging this PR, the main branch, and #477 to ensure if we configure dedicated workers:
- The worker will only execute tasks from the configured RG, if any.
- The worker will not execute tasks from the other RG while these tasks will be actively scheduled to run on general workers.

You can ask your coding agent to do it by modifying the e2e test cases to do above.

20001020ycx:
> Before we merge, is it possible to have an e2e test by merging this PR, the main branch, and #477 to ensure if we configure dedicated workers:
>
> * The worker will only execute tasks from the configured RG, if any.
> * The worker will not execute tasks from the other RG while these tasks will be actively scheduled to run on general workers.
>
> You can ask your coding agent to do it by modifying the e2e test cases to do above.

Absolutely — thanks for suggesting this during review. I've appended it as section 3 under *Validation performed*. A few details left out of the description to keep it readable:

- The test sets `scheduler.runtime.scheduler.policy: resource_group_round_robin`. Switching the chart default is left to a follow-up PR.
- It forwards the manager's RG id to the executor via `execution_manager.task_executor.inherited_env: ["SPIDER_EXTERNAL_RESOURCE_GROUP_ID"]` so a task can report which pool ran it. This is for this e2e test only; hence, not included in this PR.
- The `report_worker_pool` task is built locally from the existing `tests/huntsman/integration-test-tasks` package and mounted into the workers via `extra_volumes`, since the official worker image ships no task library.
- The test, `tests/huntsman/e2e/tests/isolation.rs`, reuses the `nn` test's `SpiderTestDriver` harness: it registers two resource groups (`rg-dedicated`, `rg-other`), submits `report_worker_pool` jobs under both, and asserts the dedicated worker ran only `rg-dedicated` tasks. Full source: [`test/2026-09-18-rg-isolation-e2e`](https://github.com/20001020ycx/spider/blob/test/2026-09-18-rg-isolation-e2e/tests/huntsman/e2e/tests/isolation.rs).

LinZhihao-723 (approved):
Thank you for updating the e2e test results. This PR should be ready to merge now!

Issue #489, opened 2026-09-17
Allow task graph inputs to reference a shared payload that is serialized and materialized once


LinZhihao-723:
### Request

Every task graph input has to carry its own payload, so a client that feeds the same value to many tasks pays for it once per input.

`SpiderClient::submit_job` takes a `Vec<TaskInput>`, one entry per task graph input, and each entry holds its own msgpack payload. Identical values are therefore serialized repeatedly on the client, framed into the same wire buffer, persisted into `jobs.compressed_serialized_job_inputs`, and materialized in the storage cache as one dataflow slot per input.

The huntsman e2e neural-network workload shows the shape of the problem: a first layer of 1000 neurons, each taking 2 inputs, needs 2000 graph inputs. If all of them are the same value, the client still sends 2000 copies and the storage cache still allocates 2000 slots holding equal bytes. zstd hides much of the wire cost, but not the client-side serialization, the deserialization on the storage side, or the per-slot memory the cache holds for the job's lifetime.

Task graph inputs should be able to reference a payload that is serialized once and shared by any number of inputs, with storage materializing it once as well.

### Possible implementation

Introduce a structured task-graph-level input in `spider-core` (`spider_core::types::io`), where each positional input is either its own value payload or a reference to a shared payload:

* `TaskGraphInput` holds the shared input payloads plus the positional inputs. `TaskGraphInputBuilder` serializes values into msgpack, hands out a `SharedInputId` per shared payload, and appends positional inputs that are either values or shared references. The type guarantees that every shared reference is in bounds, enforced by the builder and re-checked on deserialization, so downstream code never re-validates.
* Serialization: two wire frames (shared payloads, then tagged positional inputs) with zstd applied to the whole buffer, mirroring `TaskGraph::to_zstd_compressed_json`. The proto field stays `bytes`, so only the encoding inside it changes.
* `SpiderClient::submit_job` takes a `&TaskGraphInput`.
* `ValidatedJobSubmission` holds the deserialized `TaskGraphInput`, and the existing input-count check compares the number of positional inputs against the number of task graph inputs. The compressed bytes are still persisted verbatim.
* The storage cache's `TaskGraph::create` materializes each shared payload into a single dataflow slot whose `Arc` every referencing positional input shares. This is sound because task-graph-input slots never have a writer: writers exist only for task outputs.
[...]

PR #492, opened 2026-09-18
feat(spider-core): Add `TaskGraphInput` for task graph inputs to support shared input payloads.   

LinZhihao-723:
# Description

This PR partially implements the feature request #489 by defining a new type, `TaskGraphInput`. This PR does not resolve the feature request until we integrate this type into the client API and storage service in a future PR.

Specifically, this PR lands `TaskGraphInput`, the `spider-core` type that expresses that sharing, plus the `spider-utils` wire-framing primitives it needs. Nothing constructs or consumes it yet: the client, storage, and end-to-end integration follow in their own PR, so nothing is wired up and no behavior changes.

With `TaskGraphInput`, users will be able to define a shared input payload with a unique ID ticket for it. By referencing the ticket, this shared input can be reused across multiple input tasks. This type also supports wire-frame-based serialization/deserialization, with zstd compression support like the existing user-defined task inputs. The API of this type is defined to reject illegal shared payload references at construction time (either through `TaskGraphInputBuilder` or through the deserialization API), meaning that a legally constructed `TaskGraphInput` instance should always contain valid internal data references.

An end-to-end working prototype with changes introduced in this PR can be found in #490.

# Checklist

  breaking change.

# Validation performed

* [x] Add unit tests to assert expected behaviors across:
    * `TaskGraphInput` construction and serde.
    * The related wire-format utils.

## Summary by CodeRabbit

- **New Features**
  - Added support for building, serializing, compressing, and restoring task-graph inputs.
  - Added reusable shared input payloads to reduce duplicated data in task inputs.
  - Added validation and clear error reporting for malformed, oversized, or invalid input data.
  - Added support for wire frames with prefixes, multi-part payloads, and efficient payload access.

- **Bug Fixes**
  - Improved wire-frame handling to correctly preserve prefixes, trailing data, and count headers.
  - Added safeguards against truncated, corrupted, or oversized frames.


components/spider-core/src/types/io/mod.rs
```diff
@@ -1,9 +1,12 @@
+mod task_graph_input;
+
 use serde::Deserialize;
 use serde::Serialize;
 use spider_utils::wire::StreamDeserializer;
 use spider_utils::wire::WireError;
 use spider_utils::wire::WireFrameBuilder;
 use spider_utils::wire::{self};
+pub use task_graph_input::*;
```
sitaowang1998: I am not sure if a wildcast export is a good design. The `pub` might also be too broad.

components/spider-core/src/types/io/task_graph_input.rs
```diff
@@ -0,0 +1,844 @@
...
+    ) -> Result<Self, TaskGraphInputError> {
+        let (&tag_byte, body) = payload
+            .split_first()
+            .ok_or(TaskGraphInputError::EmptyEntry)?;
+        let tag =
+            EntryTag::from_repr(tag_byte).ok_or(TaskGraphInputError::UnknownEntryTag(tag_byte))?;
+        match tag {
+            EntryTag::ValuePayload => Ok(Self::ValuePayload(body.to_vec())),
+            EntryTag::SharedPayload => {
+                let raw_id = u32::from_le_bytes(
```
sitaowang1998, line 115: We should add to the docstring that we are using little endian.

sitaowang1998 (commented):
Implementation lgtm

components/spider-core/src/types/io/task_graph_input.rs
LinZhihao-723: - The current wire format uses little-endian for integer encoding, and it's pretty commonly used in modern serialization formats in general. It shouldn't be a surprise for either human readers or agent readers.
- Considering that users are not supposed to deserialize this format themselves, I don't think it's necessary to expose this level of detail in the docstring for these APIs.

components/spider-core/src/types/io/mod.rs
LinZhihao-723: Yeah I missed that. I'll fix it in an upcoming commit.

LinZhihao-723: Pushed. All pub symbols are needed.

sitaowang1998 (approved):

PR #493, opened 2026-09-22
feat(spider-client): Add resource-group creation or authentication API (resolves #486).

### LINKED ISSUE #486: Add an `add_or_verify` resource-group API that returns the ID whether or not the group already exists
### Request

A client that only knows a resource group's external ID and password has no single call that yields the group's Spider-assigned `ResourceGroupId`:

* `AddResourceGroup` returns the ID, but fails with `ALREADY_EXISTS` (mapped to `ClientError::ResourceGroupAlreadyExists` since #480) once the group exists.
* `VerifyResourceGroup` authenticates an existing group, but takes the `ResourceGroupId` the caller is trying to obtain and returns `common.Void`.

So every client that may or may not be the first to register a group has to call `add_resource_group`, catch `ResourceGroupAlreadyExists`, and then find the ID some other way. There is no way to recover it through the API.

This is not hypothetical. The huntsman e2e driver caches the resolved ID per process and registers the group on first use. As soon as a second test binary process used the same external ID, the run failed with `resource group already exists`, and the tests had to be given distinct external IDs as a workaround.

Add a resource-group API that creates the group when it is absent and authenticates it when it is present, returning the `ResourceGroupId` in both cases:

* Absent: insert the credentials and return the newly assigned ID.
* Present with a matching password: return the existing ID.
* Present with a mismatched password: fail as `UNAUTHENTICATED`, exactly as `VerifyResourceGroup` does today.

### Possible implementation

Name the operation `add_or_verify`, matching the existing `add`/`verify` vocabulary:

* `ResourceGroupManagement::add_or_verify(&self, credentials: ExternalResourceGroupCredentials) -> Result<ResourceGroupId, DbError>` in `spider-storage`'s DB protocol, implemented by `MariaDbStorageConnector`.
* `rpc AddOrVerifyResourceGroup(AddResourceGroupRequest) returns (ResourceGroupIdResponse);` on `ResourceGroupManagementService`, plus the storage-service handler.
* `SpiderClient::add_or_verify_resource_group(credentials) -> Result<ResourceGroupId, ClientError>`, delegating to `ResourceGroupManagementClient`.

The MariaDB implementation must stay correct when several clients race on the same external ID, which the `resource_groups` unique index on `external_id` already detects. Insert first, and on `MYSQL_ER_DUP_ENTRY` (the case `add` already maps to `DbError::ResourceGroupAlreadyExists`) select the existing row and verify the password, returning its ID. A single statement such as `INSERT ... ON DUPLICATE KEY UPDATE` followed by the password check works too, as long as a concurrent insert cannot surface `ResourceGroupAlreadyExists` to the caller and a mismatched password still fails as unauthorized.

Client error mapping is unchanged: a bad password maps to `ClientError::Unauthenticated`, and `ClientError::ResourceGroupAlreadyExists` should never be produced by this call.

Outcome: one call turns credentials into a `ResourceGroupId`, whether or not the group already exists.

### Out of scope

* Retiring `AddResourceGroup`/`VerifyResourceGroup`, which stay as they are.
* Updating the huntsman e2e driver to use the new API and dropping its per-test external resource group IDs. That cleanup belongs in a follow-up issue once this API exists.

sitaowang1998:
# Description

This PR resolves #486 by adding the proposed `add_or_verity_resource group` across gRPC, database and client.

## Changes
- Adds `add_or_verify` to the database interface and MariaDB implementation.
- Adds the gRPC function and regenerate Rust bindings.
- Adds the gRPC storage service.
- Exposes `SpiderClient::add_or_verify_resource_group`.
- Simplifies the e2e driver to share a constant credentials instead of per-test credential.

# Checklist

  breaking change.

# Validation performed

- [x] Adds storage service and database tests for creation, reuse, incorrect passwords, and concurrent call.
- [ ] GitHub workflows pass.

## Summary by CodeRabbit

- **New Features**
  - Added resource group registration and verification through the client and service APIs.
  - Existing resource groups now return their original ID when the supplied password is correct.
  - New resource groups are created when no matching external ID exists.

- **Bug Fixes**
  - Incorrect passwords are rejected without changing existing credentials.
  - Concurrent registration attempts now resolve consistently, allowing only the valid password to succeed.

- **Tests**
  - Added coverage for repeated registration, invalid passwords, and concurrent requests.
  - End-to-end workflows now use a shared resource group.


components/spider-storage/src/db/mariadb.rs
```diff
@@ -460,6 +460,43 @@ impl ResourceGroupManagement for MariaDbStorageConnector {
...
+            "INSERT INTO `{table}` (`external_id`, `password`) VALUES (?, ?) ON DUPLICATE KEY \
+             UPDATE `id` = `id`;",
+            table = RESOURCE_GROUPS_TABLE_NAME,
+        );
+        const SELECT_QUERY: &str = formatcp!(
+            "SELECT `id`, `password` FROM `{table}` WHERE `external_id` = ? FOR UPDATE;",
+            table = RESOURCE_GROUPS_TABLE_NAME,
+        );
+
+        use subtle::ConstantTimeEq;
```
LinZhihao-723: Hmmm, since this is now shared by both this method and `verify`, shall we promote it to a mod-level import?

components/spider-storage/src/state/test_utils.rs
```diff
@@ -189,6 +190,37 @@ impl ResourceGroupManagement for MockDbConnector {
...
+        match self
+            .resource_group_ids
+            .entry(credentials.get_external_resource_group_id().to_owned())
+        {
+            Entry::Occupied(entry) => {
+                let id = *entry.get();
+                let stored = self
+                    .resource_groups
+                    .get(&id)
+                    .ok_or(DbError::ResourceGroupNotFound(id))?;
```
LinZhihao-723: Hmmm, I undestand this is a mock function, but this error code is not a part of the trait contract, right?

components/spider-storage/tests/mariadb_test.rs
```diff
@@ -572,6 +572,94 @@ async fn test_add_duplicate_resource_group() {
...
+    let credentials = ExternalResourceGroupCredentials::new(
+        format!("test-resource-group-{}", rand::random::<u64>()),
+        b"password".to_vec(),
+    );
+
+    let (first, second) = tokio::join!(
+        storage.add_or_verify(credentials.clone()),
+        storage.add_or_verify(credentials),
+    );
+    assert_eq!(first?, second?);
```
LinZhihao-723: Should probably increase the concurrency level to make it realistic to trigger a concurrent add.

components/spider-storage/src/db/mariadb.rs
```diff
@@ -460,6 +460,43 @@ impl ResourceGroupManagement for MariaDbStorageConnector {
         Ok(ResourceGroupId::from(resource_group_id))
     }
 
+    async fn add_or_verify(
```
LinZhihao-723: More of a design question: the current implementation will burn an `AUTO_INCREMENT` ID on every single `insert` operation. Shall we do the transaction like:
- Select the row with the given external ID for update.
- If the row already exists, try to match the credential and return the ID.
- Otherwise, insert the row.

LinZhihao-723: Or is this going to create the key conflict issue on commit?

LinZhihao-723 (changes requested):
Overall lgtm. I need to spend some extra time to polish the docstring. I will probably directly push to this PR later.

components/spider-storage/src/db/mariadb.rs
sitaowang1998: The select-first approach could introduce a race when the resource group does not exist, depending on the locking level.

Under REPEATABLE READ, two transactions can both find no row and acquire compatible gap locks. When both subsequently insert, each can be blocked by the other’s gap lock, producing a deadlock.

Under READ COMMITTED, those search gap locks are disabled. Both transactions can still observe the missing row; one inserts successfully, while the other can receive a duplicate-key error when its insert proceeds.

Thus, we can set the transaction isolation level to `READ COMMITED` and use select-first method.

PR #495, opened 2026-09-23
fix(spider-execution-manager): Remove `SPIDER_RESOURCE_GROUP_PASSWORD` from the task executor environment (fixes #488).

### LINKED ISSUE #488: Task executors inherit the execution manager's whole environment, including `SPIDER_RESOURCE_GROUP_PASSWORD`
### Bug

`ProcessPool::spawn_executor` (`components/spider-execution-manager/src/process_pool.rs`) builds the `spider-task-executor` command without calling `Command::env_clear`, so every executor inherits the execution manager's entire environment. Two consequences:

* Once #484 lands, operators set `SPIDER_RESOURCE_GROUP_PASSWORD` on the execution manager to pin it to a resource group. That secret is then readable through `std::env::var` by every task executor process, i.e., by arbitrary user task code, for as long as the process runs.
* `TaskExecutorConfig::inherited_env` doesn't actually restrict anything: the variables it names are already inherited. Its only observable effect is the warning logged when a configured key is unset. This has been the case since #389.

The exposure is bounded when the execution manager is correctly pinned, since only that group's tasks run on it and the password is that group's own. It is not bounded when the execution manager falls back to unpinned while the password is still set in its environment (#487): tasks from every resource group can then read one group's password.

Expected: a task executor sees only the variables the execution manager intends to pass to it, and never the resource-group password.

Suggested fix: the minimal change is `command.env_remove(RESOURCE_GROUP_PASSWORD_ENV)` in `spawn_executor`. Making `inherited_env` mean what its docstring says requires `env_clear` plus an explicit allowlist (`PATH`, `HOME`, `LD_LIBRARY_PATH`, and whatever else the executor and TDL packages need), which is a larger change and deserves its own discussion.

### Spider version

`796c8dc` (`main`) for the inherited environment; the password exposure arrives with #484 (`abd2598`).

### Environment

N/A (found by code inspection; independent of the environment).

### Reproduction steps

1. With #484 applied, set `SPIDER_EXTERNAL_RESOURCE_GROUP_ID` and `SPIDER_RESOURCE_GROUP_PASSWORD` in the execution manager's environment, and leave `task_executor.inherited_env` empty in its config.
2. Run a job whose TDL task reads and reports `std::env::var("SPIDER_RESOURCE_GROUP_PASSWORD")`.
3. The task observes the password, even though the execution manager was not configured to forward any environment variable to its executors.

subodh-thallada:
# Description

This PR fixes #488.

`ProcessPool::spawn_executor` doesn't clear the environment of the task executor it spawns, so the executor inherits everything from the execution manager, including `SPIDER_RESOURCE_GROUP_PASSWORD` when the execution manager is pinned to a resource group. That means any task can read the password with `std::env::var`.

This PR takes the minimal fix suggested in the issue and calls `env_remove(RESOURCE_GROUP_PASSWORD_ENV)` on the executor command. It runs after the `inherited_env` loop, so the password is dropped even if someone lists it in `inherited_env`.

Making `inherited_env` an actual allowlist (`env_clear` + explicit keys) is left for a separate discussion, as mentioned in the issue.

# Checklist

# Validation performed

* Reproduced locally with a throwaway TDL task that returns `std::env::var("SPIDER_RESOURCE_GROUP_PASSWORD")`, run through `ProcessPool` with the password set and `inherited_env` empty. Before the change the task returned the password; after the change it returns `None`.
* The existing `process_pool` integration tests and `spider-execution-manager` unit tests pass.
* `cargo +nightly clippy` and `cargo +nightly fmt` are clean.

## Summary by CodeRabbit

* **Security**
  * Executor processes no longer receive the resource-group password environment variable, including when inherited environment variables are configured.


LinZhihao-723 (commented):
Hi @subodh-thallada, thanks again for your contribution!
I think this PR addresses the immediate issue with `SPIDER_RESOURCE_GROUP_PASSWORD`, but it doesn’t fundamentally solve the broader problem that all environment variables are currently inherited by the task executor.
The more complete fix would be to clear the environment before forking the task executor process, and then explicitly pass only a whitelist of allowed environment variables through the `inherited_env` option as exposed [here](https://github.com/y-scope/spider/blob/4692b8892faf738faadbbbc5d705cf2e9f103e20/tools/deployment/spider-helm/values.yaml#L50).
This is actually a known issue. The main reason we haven’t addressed it yet is that the CLP-side integration also needs to be updated to work with this whitelist-based approach.
I’ll hand this PR off to @20001020ycx to evaluate the impact on the CLP-side integration. Once that is clarified, I think we should update this PR so that the task executor starts with a clean environment and only receives the explicitly whitelisted variables.

PR #498, opened 2026-09-23
feat(spider-tdl)!: Expose the executing execution manager's resource group in `TaskContext`.

20001020ycx:
# Description

See PR title.

The goal of the newly added test is to verify the resource-group functionality end to end; specifically, it validates that:

- The worker will only execute tasks from its configured resource group, if any.
- The worker will not execute tasks from another resource group while those tasks are actively scheduled to run on the general workers.

To make the test easier, this PR also refactors `TaskContext`, with the proposed change below:

```rust
pub struct TaskContext {
    pub job_id: JobId,
    pub task_id: TaskId,
    pub task_instance_id: TaskInstanceId,
    pub task_resource_group_id: ResourceGroupId,                       // the task's group
    pub execution_manager_resource_group_id: Option<ResourceGroupId>, // the executing EM's pin; None = general
    // ...
}
```

Here we define them more carefully:

- `resource_group_id` is renamed to `task_resource_group_id` — the resource group the task belongs to (from the scheduler's task assignment). It is the same regardless of which worker runs the task.
- A new `execution_manager_resource_group_id: Option<ResourceGroupId>` carries the resource group the executing execution manager is dedicated to, or `None` for a general execution manager. The execution manager threads its registered pin into the `TaskContext` it already builds, so no environment forwarding is needed.

With this, the test (`tests/huntsman/e2e/tests/resource_group.rs`) submits interleaved jobs under a dedicated and a general resource group, each running a `report_worker_pool` task that reads its placement from the `TaskContext` and reports `general`, `dedicated`, or `poached` (ran on another group's dedicated worker); it then asserts the dedicated worker runs its own group's tasks and never a `poached` one. The test is not wired into CI yet — that needs a general + dedicated dual-pool deployment (#491).

> [!NOTE]
> Breaking change: the public `TaskContext.resource_group_id` field is renamed to `task_resource_group_id`. Task functions that read `ctx.resource_group_id` must update to `ctx.task_resource_group_id`.

# Checklist

- [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
- [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a breaking change.
- [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation Performed

- All CI checks pass.


LinZhihao-723 (changes requested):
Before we proceed further, can we break this PR into two:
- The first PR only updates `TaskContext`.
- The second PR will update the actual test cases.

20001020ycx:
> Before we proceed further, can we break this PR into two:
> 
> * The first PR only updates `TaskContext`.
> * The second PR will update the actual test cases.

Okay, on it!

